PR #5, opened 2016-02-25
Set env var for berkshelf version and usage in centos6 packer config

kdaily:
The berkshelf version is used but not set - so set it as env var.


dougalb:
Thanks!

dougalb:
I reverted this as the variables are pulled from packer_variables.json and not the environment variables.

kdaily:
Sure - apologies! Learned something new about packer today.

PR #11, opened 2016-05-27
first try re: #101

guicho271828:
https://github.com/awslabs/cfncluster/issues/101


dougalb:
We'll address this as part of the https://github.com/awslabs/cfncluster-cookbook/blob/master/recipes/_master_torque_config.rb, where we will ensure the **MasterServer** is added as a execution node and also `pbs_mon` is started.

PR #12, opened 2016-05-29
Upgrade torque to 6.0.1

guicho271828:
This enables several enhancements and bug fixes listed in http://www.adaptivecomputing.com/resources/docs/torque/6-0-1/torqueReleaseNotes6.0.1.pdf . Torque 6x is a recent verison, and is considered relatively unstable compared to the previous versions. So, I encourage frequent updates.

Among which, following are the issues resolved in 6.0.1 which seems critical:

> Down/offline nodes caused TORQUE to not online elastic nodes. pbs_
server is now able to bring up new nodes even when there are nodes in
the system that are down or offline. (TRQ-3066)

> Memory calculation issues reported when cgroups enabled and -l
vmem|pmem|mem are used. (TRQ-3499)

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -28,8 +28,8 @@ default['cfncluster']['sge']['url'] = 'http://arc.liv.ac.uk/downloads/SGE/releas
 default['cfncluster']['openlava']['version'] = '3.1.3'
 default['cfncluster']['openlava']['url'] = 'https://github.com/openlava/openlava/archive/3.1.3.tar.gz'
 # Torque software
-default['cfncluster']['torque']['version'] = '6.0.0'
-default['cfncluster']['torque']['url'] = 'https://github.com/adaptivecomputing/torque/archive/6.0.0.tar.gz'
+default['cfncluster']['torque']['version'] = '6.0.1'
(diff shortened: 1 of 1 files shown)


guicho271828:
by the way, I consider I can modify this value from extra_json value as follows:

```json
extra_json = { "cfncluster" : { "cfn_scheduler_slots" : "cores", "torque" : { "version" : "6.0.1", "url": "https://github.com/adaptivecomputing/torque/archive/6.0.1.tar.gz"}}}
```

dougalb:
How much testing has been done for these changes? Have all OS'es been tested with kitchen?

guicho271828:
RE: OSes, not really, I only tested it on ubuntu. Re: "how much", yesterday I lauched 12 c4.x8large spot or ondemand instances (216 cpu counts) and successfully conducted middle-sized jobs (payed ~150 USD in two days) with this extra_json configuration.
Torque is claimed to be compatible within major versions, so I believe it works on all other OSes.

dougalb:
We'll pull Torque 6.0.2 in for the next release. I'm going to close this as we will ensure it is tested on all supported OSes before releasing. Also I cannot find any documentation that _TRQ-3066_ helps with elastic scaling of instances in CfnCluster. Did you do any testing to see if changed the current behaviour when there are 0 instances in the **ComputeFleet**

PR #16, opened 2016-07-19
Fix SFTP setting in OpenSSH config on Debian/Ubuntu

pansapiens:
I found that SFTP wouldn't work on my Ubuntu master node - this was because the SFTP subsystem setting in /etc/ssh/sshd_config was RedHat specific.

On Debian/Ubuntu systems, the appropriate line is:

Subsystem sftp internal-sftp

This patch (should) fix this in the Chef recipe.


dougalb:
Thanks, we're just working on getting some big updates out. Is it possible for you to make this PR against the develop branch?

pansapiens:
@dougalb Whoops - that was my intention. New PR here: https://github.com/awslabs/cfncluster-cookbook/pull/17

PR #17, opened 2016-07-19
Fix SFTP setting in OpenSSH config on Debian/Ubuntu

pansapiens:
I found that SFTP wouldn't work on my Ubuntu master node - this was because the SFTP subsystem setting in /etc/ssh/sshd_config was RedHat specific.

On Debian/Ubuntu systems, the appropriate line is:

Subsystem sftp internal-sftp

This patch (should) fix this in the Chef recipe.

(new PR against develop, original here: https://github.com/awslabs/cfncluster-cookbook/pull/16)


dougalb:
Great. I'll check and probably pull tomorrow.

dougalb:
Added, thank you. Will be included in cfncluster-1.3.0.

PR #24, opened 2017-06-01
Fix for Torque builds failing on CentOS-7

dougalb:


bwbarrett:
rebase so you get rid of the merge commit and works for me.

dougalb:
Updated.

PR #26, opened 2017-06-07
Update ruby version for travisCI tests

dougalb:
Updates Ruby version for the Travis CI tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
 rvm:
-  - 2.1.5 
+  - 2.2.2
 install: bundle install --without aws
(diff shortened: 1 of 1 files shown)


dougalb:
Updating to be more specific, rvm 2.2.2

dougalb:
More sigh... This needs some work! It should really only run chefspec and not try and use vagrant etc.

rajachan:
PR #40 upgrades RVM and fixes other dependency issues that unblock the testing now. I also agree on the Vagrant comment, but I've added a native install of Vagrant for now to keep the current setup happy.

PR #32, opened 2017-08-31
Sftp: Correcting sftp subsystem path

mohanasudhan:
sftp-server is located at /usr/libexec/openssh/sftp-server
hence correcting the path in sshd
Fix: https://github.com/awslabs/cfncluster/issues/169

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -52,7 +52,7 @@ default['openssh']['server']['password_authentication'] = 'no'
 default['openssh']['server']['gssapi_authentication'] = 'yes'
 default['openssh']['server']['gssapi_clean_up_credentials'] = 'yes'
 default['openssh']['server']['x11_forwarding'] = 'yes'
-default['openssh']['server']['subsystem'] = 'sftp /usr/libexec/sftp-server'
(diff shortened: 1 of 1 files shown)


attributes/default.rb
```diff
@@ -52,7 +52,7 @@
 default['openssh']['server']['gssapi_authentication'] = 'yes'
 default['openssh']['server']['gssapi_clean_up_credentials'] = 'yes'
 default['openssh']['server']['x11_forwarding'] = 'yes'
-default['openssh']['server']['subsystem'] = 'sftp /usr/libexec/sftp-server'
+default['openssh']['server']['subsystem'] = 'sftp /usr/libexec/openssh/sftp-server'
```
bwbarrett: Can you confirm that path is correct for all the supported AMIs?

mohanasudhan: Ya just confirmed. Seems path is different on ubuntu (/usr/lib/openssh/sftp-server | /usr/lib/sftp-server) and for centos / alinux (/usr/libexec/openssh/sftp-server) 

Will update the PR

mohanasudhan: Seems debian has override already https://github.com/awslabs/cfncluster-cookbook/commit/1a8856ce388df19e4e2847f76511762b343bfcdf

This PR looks good to me now.

PR #37, opened 2017-09-18
Updating slurm service files to using systemd where appropriate.

jocherry:
Slurm has differing service files on master and compute nodes, leading to long timeouts when the wrong service file was called by our sysinit scripts.
Adding the newer systemd scripts on systems that support it, to correct this bug.

Signed-off-by: Jordan Cherry <cherryj@amazon.com>


jocherry:
https://github.com/awslabs/cfncluster/issues/209

recipes/_master_slurm_config.rb
```diff
@@ -48,6 +48,15 @@
   mode '0755'
 end
 
+cookbook_file '/etc/systemd/system/slurmctld.service' do
+  source 'slurmctld.service'
+  owner 'root'
+  group 'root'
+  mode '0755'
+  action :create
+  only_if { ::File.exist?('/bin/systemctl') }
```
mohanasudhan: CentOS6 and alinux didn't adopt systemd. Is this fix only for ubuntu1604 and centos7?

dougalb: Why not use if node['init_package'] = systemd

PR #43, opened 2017-10-31
Fix Rubocop warnings.

rajachan:
- Rubocop warns against using EO* delimiters for heredocs.
  Using more meaningful ones.
- Use percentage notation for symbol arrays
- Force `zero_only` octal style (to allow `mode 0744`, etc.)
- Enumerate Gemfile packages in alphabetical order as expected by rubocop
- Adding Build status to README

---
Down from 58 Rubocop offenses to one minor style issue, but there are no more warnings/errors that cause the build to fail.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,15 +3,15 @@ source 'https://rubygems.org'
 gem 'berkshelf'
 
 group :style do
+  gem 'foodcritic', '~> 6.0.1'
   gem 'rake', '~> 11.1.1'
   gem 'rubocop', '~> 0.38'
-  gem 'foodcritic', '~> 6.0.1'
 end
 
 group :test do
-  gem 'test-kitchen', '~> 1.13.2'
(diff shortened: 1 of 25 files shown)


Gemfile
```diff
@@ -3,15 +3,15 @@ source 'https://rubygems.org'
 gem 'berkshelf'
 
 group :style do
+  gem 'foodcritic', '~> 6.0.1'
```
bwbarrett: just a nit, but why the reorder?

rajachan: Rubcop wants it in alphabetical ordering.

bwbarrett: ah; makes sense, although a comment in the commit would have saved me some typing :p.

rajachan: That's fair. WIll add more info to the commit message for history.

The EOF/H change was to make rubocop happy again. It filters out EO* delimiters and warns about using a more meaningful one.

PR #46, opened 2017-11-03
Fixing kitchen test builds by reverting to curly braces

jocherry:
Tested running kitchen test locally
signed off by: cherryj@amazon.com


mohanasudhan:
We need to fix CI soon but in the interest of time in fixing kitchen test. I am approving this PR

PR #48, opened 2017-11-06
Upgrading packages in base_install chef receipe

mohanasudhan:
Upgrading packages during base_install receipe
execution. As we add upgrade here, we can now
remove packer command on upgrade

Confirmed platform_family is correct for amazon linux too
Ref: https://github.com/awslabs/cfncluster-cookbook/blob/cf6c7c287b0effcd4d76f8a884d334086cb4bf89/attributes/default.rb#L59

And tested 
kitchen test centos7-minimal - working
kitchen test amazon-linux is running successful ( need couple of minutes to confirm)

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -24,8 +24,14 @@ when 'rhel'
       command "yum-config-manager --enable #{node['cfncluster']['rhel']['extra_repo']}"
     end
   end
+  execute 'yum-update' do
+    command "yum -y update"
+  end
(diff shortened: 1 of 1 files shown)


mohanasudhan:
Ya receipe yum should update as per the link you referred should work but when we were testing centos ami build it was not working. we should definitely dive in to see what's happening, I am going to create an issue. 

Pushing this change to move a step forward with kitchen test

PR #51, opened 2017-11-07
Unpin kernel update in rhel platform family

mohanasudhan:
We are pinning to specific kernel version in nvidia_install
recepie, but centos6 ami we have is very old and the specific
version of kernel is not available in repository during
downgrade. Hence removing specific kernel version install for rhel
platform family

### DIFF
--- a/recipes/_nvidia_install.rb
+++ b/recipes/_nvidia_install.rb
@@ -18,10 +18,7 @@ if node['cfncluster']['nvidia']['enabled'] == 'yes'
 
   case node['platform_family']
   when 'rhel'
-    yum_package node['cfncluster']['kernel_devel_pkg']['name'] do
-      version node['cfncluster']['kernel_devel_pkg']['version']
-      allow_downgrade true
-    end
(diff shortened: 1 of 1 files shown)


recipes/_nvidia_install.rb
```diff
@@ -18,10 +18,7 @@
 
   case node['platform_family']
   when 'rhel'
-    yum_package node['cfncluster']['kernel_devel_pkg']['name'] do
-      version node['cfncluster']['kernel_devel_pkg']['version']
```
bwbarrett: is there a 'version' set somewhere that we should remove to not confuse ourselves later?

mohanasudhan: https://github.com/awslabs/cfncluster-cookbook/search?utf8=%E2%9C%93&q=%22yum_package%22&type=
https://github.com/awslabs/cfncluster-cookbook/search?utf8=%E2%9C%93&q=kernel_devel_pkg&type=

I don't see any other place where version is pin'ed with yum install.

PR #52, opened 2017-11-07
Fix for slurm install.

jocherry:
This was previously breaking during kitchen test, due to slurm service not being found.
We now choose the correct service to enable / start

Signed-off-by: Jordan Cherry <cherryj@amazon.com>

### DIFF
--- a/recipes/_compute_slurm_config.rb
+++ b/recipes/_compute_slurm_config.rb
@@ -28,17 +28,15 @@ cookbook_file '/etc/systemd/system/slurmd.service' do
   group 'root'
   mode '0755'
   action :create
-  only_if { ::File.exist?('/bin/systemctl') }
+  only_if { node['init_package'] == 'systemd' }
 end
 
-if ::File.exist?('/bin/systemctl') 
+if node['init_package'] == 'systemd' 
   service "slurmd" do
     supports restart: false
     action %i[enable]
   end
-end
-
-if  ! ::File.exist?('/bin/systemctl') 
+else
   service "slurm" do
     supports restart: false
     action %i[enable]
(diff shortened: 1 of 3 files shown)


recipes/_compute_slurm_config.rb
```diff
@@ -31,7 +31,16 @@
   only_if { ::File.exist?('/bin/systemctl') }
 end
 
-service "slurm" do
-  supports restart: false
-  action %i[enable]
+if ::File.exist?('/bin/systemctl')
```
bwbarrett: Isn't there a better test for init style?

dougalb: https://github.com/awslabs/cfncluster-cookbook/blob/develop/recipes/_ganglia_install.rb#L59

https://docs.chef.io/ohai.html using the init_package.rb plugin.

jocherry: yeah I saw that. but https://github.com/awslabs/cfncluster-cookbook/blob/develop/recipes/_ganglia_install.rb#L65 https://github.com/awslabs/cfncluster-cookbook/blob/develop/recipes/_ganglia_install.rb#L71 and 
https://github.com/awslabs/cfncluster-cookbook/blob/develop/recipes/_ganglia_install.rb#L78
made me unsure if this was reliable.
I googled a bit about this issue and became convinced checking for systemctl was the most reliable way to determine this, but I am not clear on ohai, will have a look.

bwbarrett: I think you're seeing two different things.  The ohai check is the right way to check for init system type from Chef.  The things on line 65 and 71 you reference are to not overwrite an existing startup file.  Remember that these recipes work in many contexts, and the starting state isn't always known, so some defensive programming may be the right thing.

PR #56, opened 2017-11-09
Add script for updating the CentOS base AMIs

bwbarrett:
Because there are no CentOS-provided community AMIs for
CentOS, we use a set of images imported from CentOS isos.
Add a script to upgrade / clean those AMIs to the latest
minor release of the series.

### DIFF
--- a/packer_update_centos_base.json
+++ b/packer_update_centos_base.json
@@ -1,0 +1,51 @@
+{
+  "variables" : {
+    "build_date" : "{{env `BUILD_DATE`}}"
+  },
+  "builders" : [
+    {
+      "type": "amazon-ebs",
+      "name" : "CentOS 6",
+      "region": "us-east-1",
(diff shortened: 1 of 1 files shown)


packer_update_centos_base.json
```diff
@@ -0,0 +1,51 @@
...
+    }
+  ],
+  "provisioners" : [
+    {
+      "type" : "shell",
+      "expect_disconnect" : true,
+      "inline" : [
+        "sudo yum -y upgrade",
+        "echo Update Complete.  Rebooting.",
+        "sudo reboot ; sleep 30"
```
bwbarrett: In case someone wonders, the sleep is because centos6 reboot returns before ssh shut down, so the next set of commands started running, and then the build failed when the instance actually rebooted.  The sleep keeps this set of shell commands from returning so quickly, so the instance reboot starts before the next set of shell provisioners starts.

PR #62, opened 2017-11-10
Fix kernel-lt and ENA support in CentOS 6 base AMI

bwbarrett:
The previous packer scripts to update the CentOS base
AMIs broke CentOS 6 in two critical ways.  First, the
kernel was changed from 3.10.x (the kernel-lt packages)
to 2.6.32 (the kernel packages).  Second, I forgot to
build ENA support into the updated kernel.  This patch
addresses both issues.

First, the default kernel in /etc/sysconfig/kernel is
changed from kernel to kernel-lt, which means that any
future yum upgrades will leave the kernel-lt version as
the default in grub (someone must have been updating
grub by hand previously).  So the 3.10/2.6.32 problem
should be fixed going forward.

Second, we now build the ENA kernel module after the
reboot step, so that we make sure we build the kernel
module for the new version of the kernel.  It might
be better to build the ENA kernel module for all
installed kernel versions (ie, include the 2.6.32
module as well), but I have mixed feelings given the
limited testing and decided to skip that step for now.

### DIFF
--- a/centos-upgrade-first-stage.sh
+++ b/centos-upgrade-first-stage.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+#
+# Script to run the first stage (everything through rebooting to flip into
+# the new kernel) for upgrading the CentOS AMIs from the imported version
+# to the latest in that major release series.
+#
+# By the time this script was written, CentOS 7.4 had shipped, which
+# includes ENA driver by default.  CentOS 6, being ancient, won't
+# include such upgrades, so we do that after upgrade (actually,
+# centos-upgrade-second-stage does that, so that we don't have to
+# guess at what the new default kernel version is).
+#
+
+if test "`sed -n '/CentOS release 6\..*/p' /etc/centos-release`" != ""; then
+    is_centos6=1
+else
+    is_centos6=0
+fi
+
+if test $is_centos6 -eq 1 -a "`/sbin/lsmod | grep ena`" != "" ; then
+    echo "This script is lazy and can not be run on instances which"
+    echo "use ENA for networking.  Either update the script or use"
+    echo "a non-ENA instance type."
+    echo "***** Aborting now *****"
+    exit 1
+fi
+
+if test $is_centos6 -eq 1; then
+    # CfnCluster CentOS AMIs have always shipped with the elrepo -lt
+    # kernel as the default, but it appears to have been set as the
+    # default by hand in grub.conf.  Make the -lt kernel the default
+    # in systconfig, so yum upgrade won't revert the default kernel
+    # back to the 2.6.32 default kernel.
+    echo "Setting kernel-lt as default kernel series"
+    sudo /bin/sed -r -i -e 's/^DEFAULTKERNEL=kernel$/DEFAULTKERNEL=kernel-lt/' /etc/sysconfig/kernel || exit $?
+fi
+
+# Upgrade everything!
+echo "Running upgrades!"
+sudo yum -y upgrade
+
+echo "Update Complete.  Rebooting."
+# sleep for 30 seconds to make sure packer doesn't try to run the next
+# step before the reboot happens
+sudo reboot ; sleep 30
--- a/centos-upgrade-second-stage.sh
+++ b/centos-upgrade-second-stage.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+#
+# Script to run the second stage (everything after rebooting into the
+# newest kernel) for upgrading the CenOS AMIs from the imported
+# version to the latest in that major release series.
+#
+
+if test "`sed -n '/CentOS release 6\..*/p' /etc/centos-release`" != ""; then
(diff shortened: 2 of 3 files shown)


centos-upgrade-first-stage.sh
```diff
@@ -0,0 +1,26 @@
...
+# to the latest in that major release series.
+#
+
+# Upgrade everything!
+sudo yum -y upgrade
+
+if test `cat /etc/centos-release | sed -n 's/CentOS release \([0-9]\).*/\1/p'` == 6; then
+    # Remove the 2.6.32 kernel series.
+    sudo yum -y remove kernel kernel-devel
+    export NEXT=`ls -1t /lib/modules/ | head -n 1`
```
mohanasudhan: I would prefer `ls -1v` and take last one. I am not sure about sort by time `t`

centos-upgrade-first-stage.sh
```diff
@@ -0,0 +1,26 @@
...
+    echo "Guessing new kernel is $NEXT"
+    cd amzn-drivers-master/kernel/linux/ena
```
mohanasudhan: I am missing on how we get this aws driver modules into centos machine, are we getting it from https://github.com/amzn/amzn-drivers/tree/master/kernel/linux/ena ? if so we are missing code on download.

packer_update_centos_base.json
```diff
@@ -34,17 +34,13 @@
     {
       "type" : "shell",
       "expect_disconnect" : true,
-      "inline" : [
-        "sudo yum -y upgrade",
-        "echo Update Complete.  Rebooting.",
-        "sudo reboot ; sleep 30"
-      ]
+      "script" : "centos-upgrade-first-stage.sh"
```
mohanasudhan: we need to restructure our folders for build scripts, just adding remark. will do it after release

centos-upgrade-first-stage.sh
bwbarrett: -t is actually the best option; you don't want the highest kernel version, you want the last one that was installed (because that's what Grub is going to boot).  But even -t is probably a bit too fragile.  I'll think of a better option here.  Now that I just made it a shell script (instead of inline packer commands), I think grepping through grub.cfg may (unfortunately) be the best option.  The other option is to only run this on non-ENA platforms, which would make this code much simpler.

centos-upgrade-first-stage.sh
bwbarrett: Yeah, this was me taking a shortcut; they're already on the image from the last update (shame on us for that).  I'll pull the latest version from Git.

packer_update_centos_base.json
bwbarrett: +1

centos-upgrade-first-stage.sh
mohanasudhan: ok

bwbarrett:
Nah, this first attempt is way too brittle; new patch coming Real Soon Now that sucks less.

bwbarrett:
Cut https://github.com/awslabs/cfncluster-cookbook/issues/63 to remind us about moving all the scripts somewhere more useful.

centos-upgrade-second-stage.sh
```diff
@@ -0,0 +1,42 @@
...
+    git clone https://github.com/amzn/amzn-drivers.git
+    cd amzn-drivers
+    git checkout ena_linux_1.3.0
+
+    echo "Building ENA source"
+    cd kernel/linux/ena
+    make
+
+    echo "Installing ENA source"
+    sudo cp ena.ko /lib/modules/`uname -r`/weak-updates/.
```
rajachan: Looking at the ena driver Makefile, it doesn't seem to support a make install. A better way to install it here is:

`make -C /lib/modules/$(BUILD_KERNEL)/build M=$(CURDIR) modules_install`

rajachan: Replace `$BUILD_KERNEL` with `uname -r` like you did earlier. I just took the line from the Makefile and added _install.

rajachan:
One way to deal with building kernel modules in a version-agnostic way is dkms. Maybe we can look at that after the release.

centos-upgrade-second-stage.sh
bwbarrett: Yeah, This is what we've done in the past, would rather not change that today....

mohanasudhan:
Looks good to me.

centos-upgrade-second-stage.sh
rajachan: I don't think it will be intrusive to break anything. The modules_install will make sure the exported symbols get updated in the symbol table correctly for dependencies to use.

bwbarrett: I welcome a patch after we release CfnCluster 1.4.0.  We need to get this thing shipped today, and changes that are good but not necessary can wait.

PR #65, opened 2017-11-11
amis: Use CentOS community AMIs as base

bwbarrett:
Use the CentOS-provided community AMIs as the base for our
CentOS AMIs, rather than our home-imported base AMI.  We
still update the AMIs to the latest release as a seperate
step from building the final images out of expediency for the
next release.

These packages configure the AMIs to match what the imported
base AMIs do.  They configure epel on both AMIs and elrepo
on CentOS 6 (although only the kernel package from elrepo).

### DIFF
--- a/packer_update_centos_base.json
+++ b/packer_update_centos_base.json
@@ -7,7 +7,7 @@
       "type": "amazon-ebs",
       "name" : "CentOS 6",
       "region": "us-east-1",
-      "source_ami": "ami-b7c484a0",
+      "source_ami": "ami-500d8546",
       "instance_type": "t2.micro",
       "ssh_username": "centos",
       "ssh_pty" : true,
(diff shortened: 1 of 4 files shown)


packer_update_centos_base.json
```diff
@@ -32,6 +32,11 @@
   ],
   "provisioners" : [
     {
+      "type": "file",
+      "source": "centos6.elrepo.repo",
+      "destination": "/tmp/centos6.elrepo.repo"
```
mohanasudhan: minor - we don't need to copy this file for centos 7

bwbarrett: Correct, but there's no good way in Packer to not copy it.  So we do a tiny bit of extra work so that we can have the same config file.

PR #66, opened 2017-11-12
amis: Fix CentOS base AMI dependencies

bwbarrett:
The Chef recipes expect pip and awscli to be installed, so do so.
They also expect the base AMIs to have SELinux disabled, so disable
SELinux, since CentOS community AMIs apparently enable it by
default.

### DIFF
--- a/centos-upgrade-first-stage.sh
+++ b/centos-upgrade-first-stage.sh
@@ -64,7 +64,17 @@ echo "Installing basic packages"
 sudo yum -y groupinstall development
 sudo yum -y install git yum-utils
 
+if test $is_centos6 -eq 1; then
+    sudo yum -y install python-pip
+    sudo pip install awscli --upgrade
+else
+    sudo yum -y install python-pip python-wheel awscli
+fi
+
 set +e
+
+echo "Disabling SELinux"
+sudo /bin/sed -r -i -e 's/SELINUX=enforcing/SELINUX=permissive/' /etc/selinux/config
 
 echo "Update Complete.  Rebooting."
 # sleep for 30 seconds to make sure packer doesn't try to run the next


jocherry:
Don't all AMIs want pip and AWS cli? Or do they do it seperately?

bwbarrett:
awscli is installed by default on all the front page AMIs.  I'm actually surprised it isn't in the CentOS community AMIs; that's kind of lame.

bwbarrett:
But yeah, long term, it would probably be better to not have that assumption in the Chef recipes.  At least they check for awscli.  sigh.  I get the assumption with SELinux; disabling SELinux requires a reboot, which is all kinds of problems to something like Chef.

centos-upgrade-first-stage.sh
```diff
@@ -64,8 +64,18 @@ echo "Installing basic packages"
 sudo yum -y groupinstall development
 sudo yum -y install git yum-utils
 
+if test $is_centos6 -eq 1; then
+    sudo yum -y install python-pip
+    sudo pip install awscli --upgrade
+else
+    sudo yum -y install python-pip python-wheel awscli
```
mohanasudhan: yum update of awscli for centos7 will bring bit older version 1.11.133 as per https://pkgs.org/download/awscli, but pip install get us latest cli (1.11.185). Is there any reason why you are not using pip install of awscli for centos7?

dougalb:
If you use a packaged AWS CLI from a repo, it will automatically update on
the launch of an instance, so preferred. Using pip pulls the latest from
pypi but it’s version is fixed.

centos-upgrade-first-stage.sh
bwbarrett: What @dougalb said below.  In general, we should always use the vendor-provided package over building our own.  There are a couple reasons.  First, updates are much cleaner.  Second, if a customer later wants to install a package that depends on the awscli package, all kinds of hell would break loose if we installed outside the vendor packaging system if there are any file conflicts.  Third, it can cause significant customer confusion over who owns the software.

We have to install awscli via pip for CentOS 6, because there is no package for it.  But CentOS 7 has a package, so that's the right answer.

dougalb:
The awscli package is part of EPEL, so depending on when you enable EPEL
you can install from there.

https://admin.fedoraproject.org/pkgdb/package/rpms/awscli/

PR #71, opened 2017-11-13
Updating centos7 ami (which has latest script update)

mohanasudhan:
Updating centos7 ami which has latest nvidia driver
fix (https://github.com/awslabs/cfncluster-cookbook/pull/68/files) made by Brian 


mohanasudhan:
Ya update with PR id.

PR #73, opened 2017-11-16
Use the realpath of the NVMe devices instead of the symlinks.

rajachan:
The rest of the setup-ephemeral-disks.sh scripts functions with the assumption
that $MAPPING refers to a block device in /dev, but when using the udev
symlink for instance store devices, it breaks (for eg., it ends up looking for
/dev/nvme-Amazon_EC2_NVMe_Instance_Storage_AWSxxx, which would not exist).

Use the right prefix when partitioning NVMe devices.

Traditional block device partitions take a label of the form `/dev/sd*<n>` for a
partition 'n', while NVMe devices take the label `/dev/nvme*p<n>`.

Signed-off-by: Raghu Raja <craghun@amazon.com

### DIFF
--- a/files/default/setup-ephemeral-drives.sh
+++ b/files/default/setup-ephemeral-drives.sh
@@ -28,8 +28,10 @@ function setup_ephemeral_drives () {
   mkdir -p ${cfn_ephemeral_dir} || RC=1
   chmod 1777 ${cfn_ephemeral_dir} || RC=1
   if ls /dev/nvme* >& /dev/null; then
+    IS_NVME=1
     MAPPING=$(realpath --relative-to=/dev/ -P  /dev/disk/by-id/nvme*Instance_Storage*)
   else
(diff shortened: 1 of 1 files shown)


rajachan:
The instance-local NVMe volumes are still not getting aggregated into an LVM, but that's due to a different issue caused by how we partition the volumes.

files/default/setup-ephemeral-drives.sh
```diff
@@ -46,11 +48,16 @@ function setup_ephemeral_drives () {
     for d in $DEVS; do
       d=/dev/${d}
       dd if=/dev/zero of=${d} bs=32k count=1 || RC=1
-      parted -s ${d} mklabel msdos || RC=1
+      parted -s ${d} mklabel gpt || RC=1
```
bwbarrett: long live msdos partition tables...

PR #74, opened 2017-11-17
Reboot after attaching EBS volumes.

rajachan:
Adding a reboot that subscribes to the execute action which attaches EBS volumes to the instance. This is necessary so the outdated NVMe driver in CentOS 6 can discover the newly added volume.

### DIFF
--- a/recipes/_master_base_config.rb
+++ b/recipes/_master_base_config.rb
@@ -33,12 +33,18 @@ execute "attach_volume" do
   creates dev_path
 end
 
+reboot 'ec2_nvme_reboot' do
+  action :reboot_now
+  reason 'Reboot so the NVMe driver can discover the newly attached EBS volume.'
+  subscribes :reboot_now, "execute[attach_volume]", :immediately
+end
+
 # wait for the drive to attach, before making a filesystem
 ruby_block "sleeping_for_volume" do
   block do
     wait_for_block_dev(dev_path)
   end
-  subscribes :run, "execute[attach_volume]", :immediately
+  subscribes :run, "reboot[ec2_nvme_reboot]", :immediately
 end
 
 # Setup disk, will be formatted xfs if empty


recipes/_master_base_config.rb
```diff
@@ -33,12 +33,18 @@
...
+  subscribes :reboot_now, "execute[attach_volume]", :immediately
+end
+
 # wait for the drive to attach, before making a filesystem
 ruby_block "sleeping_for_volume" do
   block do
     wait_for_block_dev(dev_path)
   end
-  subscribes :run, "execute[attach_volume]", :immediately
+  subscribes :run, "reboot[ec2_nvme_reboot]", :immediately
```
bwbarrett: why would we do this all this time?  Seems better to only do it with NVMe volumes.  Even then, not sure that I love the reboot here.  Shouldn't the volume be attached when the instance launches with the cloudformation template?

dougalb:
Maybe no support for NVME devices or instances in CentOS 6? At it's core it is now over 7 years old - https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#RHEL_6

bwbarrett:
@dougalb, that was where I was going...  I wasn't ok leaving CentOS 7 behind, but it really feels like leaving CentOS 6 behind is the right choice, given all the weird corner case issues.

rajachan:
@bwbarrett  and I discussed this offline and decided to add a conditional reboot for CentOS 6 for now. I will also check to see if triggering a rescan of the PCI bus helps the driver discover the volumes; that might be a cleaner solution. This seems to be the final hurdle to get a CentOS 6 MasterServer running on C5s, so doing something conditionally for this might be worthwhile.

rajachan:
Well, a rescan of the the PCI bus with `echo 1 > /sys/bus/pci/rescan`works. That's pretty much what we need then, instead of the reboot.

PR #75, opened 2017-11-18
Adding support to translate NVMe devices to EBS volume-id

mohanasudhan:
During chef bootstrapping, we add block device to
volume mapping in by-ebs-volume-id (which is used in NFS chef).
NVMe device translation to EBS volume-id is broken, hence
C5 instance were not failing to launch via cfncluster.
Following fix will help in move forward in releasing with C5.

Tested using custom_cookbook option and able to launch c5 on alinux, ubuntu14/6, centos7

### DIFF
--- a/recipes/_ec2_udev_rules.rb
+++ b/recipes/_ec2_udev_rules.rb
@@ -20,6 +20,13 @@ cookbook_file 'ec2-volid.rules' do
   mode '0644'
 end
 
+cookbook_file 'cfncluster-ebsnvme-id' do
+  path '/usr/local/sbin/cfncluster-ebsnvme-id'
+  user 'root'
+  group 'root'
+  mode '0744'
+end
+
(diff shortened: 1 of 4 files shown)


rajachan:
Looks good, as long as @bwbarrett signs off on shipping with the ebsnvme-id script.

recipes/_ec2_udev_rules.rb
```diff
@@ -20,6 +20,13 @@
   mode '0644'
 end
 
+cookbook_file 'ebsnvme-id' do
+  path '/sbin/ebsnvme-id'
```
bwbarrett: Let's install this in /usr/local/sbin, and maybe call it cfncluster-ebsnvme-id just in case the other gets installed by some other package and conflicts.

mohanasudhan: Ya addressed both, updated path and script name.

PR #77, opened 2017-11-21
Remove Python version dependency in cfncluster-ebsnvme-id.

rajachan:
CentOS 6 ships with Python 2.6, while the ebsnvme-id script expects Python 2.7.
The script works fine with v2.6 as well, so making it use whichever interpreter
is available by default on the system.


rajachan:
Tested on CentOS 6 running Python 2.6, as well as CentOS 7 running Python 2.7.5.

rajachan:
Will merge this PR after a quick test with Ubuntu AMIs too.

PR #78, opened 2017-11-21
Rescan PCI bus to allow outdated NVMe drivers to discover volumes.

rajachan:
With CentOS 6 AMIs that have an outdated NVMe driver, we need to force a rescan
of the PCI bus after attaching an EBS volume so the driver can attach to it. The
rescan should not be harmful for other platforms, so triggering it
unconditionally. Implementing it here as the rescan has to be done after the
boto call has been made to attach a volume (from attachVolume.py) while polling
for the device so the rescan can cause our udev rule to kick in for
/dev/disk/by-ebs-volumeid.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -7,9 +7,19 @@ require 'timeout'
 #
 def wait_for_block_dev(path)
   Timeout.timeout(60) do
-    sleep(1) until ::File.blockdev?(path)
+    until ::File.blockdev?(path)
+      sleep(5)
+      rescan_pci()
+    end
     Chef::Log.debug("device ${path} not ready - sleeping 1s")
   end
+end
+
+#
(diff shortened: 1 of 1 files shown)


rajachan:
Works on CentOS 6, testing underway for other platforms. Also note that I tried using the sysfs receipe from Supermarket, but that does not do the trick on all AMIs.

libraries/helpers.rb
```diff
@@ -7,11 +7,21 @@
 #
 def wait_for_block_dev(path)
   Timeout.timeout(60) do
-    sleep(1) until ::File.blockdev?(path)
+    until ::File.blockdev?(path)
+      sleep(5)
```
rajachan: Bumped up the sleep between retries to minimize the number of times we rescan the bus. Scanning each second seems a bit wasteful.

PR #81, opened 2017-11-27
Fix interactive dialog during apt-get upgrade

mohanasudhan:
During apt-get upgrade, ubuntu 16.04 getting struck with an interactive dialog stating
a new version of grub/menu.list is available and prompt is to select which version of menu.list to proceed. This is a known issue refer: 
https://devops.stackexchange.com/questions/1139/how-to-avoid-interactive-dialogs-when-running-apt-get-upgrade-y-in-ubuntu-16
https://serverfault.com/questions/645566/a-new-version-of-boot-grub-menu-lst-is-available-when-upgrading-ubuntu-on-an

Tested both ubuntu14 & 16 cookbook

### DIFF
--- a/recipes/_update_packages.rb
+++ b/recipes/_update_packages.rb
@@ -20,6 +20,6 @@ when 'rhel'
   end
 when 'debian'
   execute 'apt-upgrade' do
-    command "apt-get update && apt-get -y upgrade && apt-get autoremove"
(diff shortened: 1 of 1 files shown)


mohanasudhan:
I tried updating ubuntu 16.04 ami to latest and kitchen tests work. Instead of this PR, I am confident on just updating ami. Will send PR soon for ami update.

PR #87, opened 2018-01-04
Ubuntu Packer and _update_packages fixes

rajachan:
This PR has two related yet independent commits, so read both commit messages for justification.

I have tested the non-interactive upgrade fix manually with an Ubuntu 1604 AMI, but have not tested it via kitchen yet. Will hold off on merging this PR until I do that.

### DIFF
--- a/packer_ubuntu1404.json
+++ b/packer_ubuntu1404.json
@@ -69,14 +69,6 @@
     {
       "type" : "shell",
       "inline" : [
-        "sudo apt-cache search build-essential",
-        "sudo apt-get clean",
-        "sudo apt-get update"
-      ]
-    },
-    {
-      "type" : "shell",
-      "inline" : [
         "sudo apt-get -y install build-essential curl wget"
       ]
     },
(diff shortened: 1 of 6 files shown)


dougalb:
This seems like a lot of forcing? I would be a little concerned that it might force package upgrades that you do not want outside of addressing the interactive popup issue.

rajachan:
My understanding is that it would only apply to cases where debconf has to choose between a conflicting older config and one coming from a repository; as in the case of grub's menu.lst here. Even with other packages that fit this criteria, the safer option is to keep an older local config instead of pulling from the repo. The team had discussed this offline before I sent out a PR and reached consensus that this is not ideal, but until we have an Ubuntu AMI that comes with the upgraded grub we can make do with it.

packer_ubuntu1404.json
```diff
@@ -66,14 +66,6 @@
     }
   ],
   "provisioners" : [
-    {
-      "type" : "shell",
-      "inline" : [
-        "sudo apt-cache search build-essential",
-        "sudo apt-get clean",
-        "sudo apt-get update"
```
mohanasudhan: Can you update other packer file to use _update_packages recipe ?

PR #88, opened 2018-01-05
Region: Adding describe-regions api to pull regions list

mohanasudhan:
we want to release cfncluster to eu-west-3 region as well,
instead of adding specific region name, I am using
describe-regions api to pull in regions list.

### DIFF
--- a/build_ami.sh
+++ b/build_ami.sh
@@ -8,7 +8,7 @@ public=$3
 build_date=$4
 
 available_os="centos6 centos7 alinux ubuntu1404 ubuntu1604"
-available_regions="eu-west-1,eu-west-2,ap-southeast-1,ap-southeast-2,eu-central-1,ap-northeast-1,ap-northeast-2,us-west-2,sa-east-1,us-west-1,us-east-2,ap-south-1,ca-central-1"
+available_regions=$(aws ec2 describe-regions --output text | awk '{print $3}' | xargs | sed -e 's/ /,/g')
 
 if [ "x$os" == "x" ]; then
   echo "Must provide OS to build."


build_ami.sh
```diff
@@ -8,7 +8,7 @@ public=$3
 build_date=$4
 
 available_os="centos6 centos7 alinux ubuntu1404 ubuntu1604"
-available_regions="eu-west-1,eu-west-2,ap-southeast-1,ap-southeast-2,eu-central-1,ap-northeast-1,ap-northeast-2,us-west-2,sa-east-1,us-west-1,us-east-2,ap-south-1,ca-central-1"
+available_regions=$(aws --region $region ec2 describe-regions --output text | awk '{print $3}' | grep -v $region | xargs | sed -e 's/ /,/g')
```
rajachan: We don't really need `$region` here, do we? Also, if someone passes `all` as the region, the command will act up here. This could just be `aws ec2 describe-regions --output text | awk '{print $3}' | xargs | sed -e 's/ /,/g'`

mohanasudhan: Ya we do, available_regions is list of regions we copy ami to, after building ami in $region. 
Refer 
https://github.com/awslabs/cfncluster-cookbook/blob/180f5441830f6623907ebc94824e834fc52ab67f/packer_centos7.json#L21

https://www.packer.io/docs/builders/amazon-ebs.html#ami_regions

mohanasudhan: Packer detects duplicate and avoids copying

Tested locally

1515189434,,ui,message,    amazon-ebs: Copying to: eu-west-1
1515189434,,ui,message,    amazon-ebs: Avoiding copying AMI to duplicate region us-east-1

will update the code.

jocherry:
We don't want all regions thought do we?  For instance the gov cloud regions?

mohanasudhan:
We don't have govcloud support today.

PR #94, opened 2018-01-11
Update CentOS 6 base AMI.

rajachan:
Using the community CentOS AMI and updating to kernel-lt using the packer
pipeline we have is causing issues with the root volume resizing. Based on the
consensus the dev. team arrived at, this commit updates the base AMI by manually
deriving it from ami-9ff841e5 - the last-known AMI to have had kernel-lt working
well with growroot. The manual update included --- moving to the latest 4.x
kernel-lt, installing ENA driver, adding necessary grub commandline arguments to
grub.conf, and cleaning up temporary state and files.

Although not used for this AMI update, packer_update_centos_base.json has been
fixed to handle root volume resizing for future releases based on community AMIs.

### DIFF
--- a/centos-upgrade-second-stage.sh
+++ b/centos-upgrade-second-stage.sh
@@ -40,6 +40,11 @@ fi
 echo "Cleaning out old kernels"
 sudo package-cleanup -y --oldkernels --count=1
 
+if test $is_centos6 -eq 1; then
+    sudo yum install -y cloud-init cloud-utils dracut-modules-growroot cloud-utils-growpart
+    rpm -qa kernel-lt | sed 's/^kernel-lt-//'  | xargs -I {} sudo dracut -f /boot/initramfs-{}.img {}
+fi
+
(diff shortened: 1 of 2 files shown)


centos-upgrade-second-stage.sh
```diff
@@ -40,6 +40,11 @@ fi
 echo "Cleaning out old kernels"
 sudo package-cleanup -y --oldkernels --count=1
 
+if test $is_centos6 -eq 1; then
```
bwbarrett: Did you actually use these changes?  If not, why are they in this commit?

rajachan: I made a note in the second part of the commit message. I did not use it, but tested them outside of the rest of the packer steps and it works with the community AMI; just not when done along with the other updates.  Added it to the packer file so we don't lose it for the next release.

bwbarrett: ok; fair enough :)

PR #97, opened 2018-01-11
Bumping up cfncluster_version for the v1.4.2 release.

rajachan:

### DIFF
--- a/packer_variables.json
+++ b/packer_variables.json
@@ -1,5 +1,5 @@
 {
-  "cfncluster_version": "1.4.1",
+  "cfncluster_version": "1.4.2",
   "cfncluster_cookbook_version": "1.4.1",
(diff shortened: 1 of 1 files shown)


packer_variables.json
```diff
@@ -1,5 +1,5 @@
 {
-  "cfncluster_version": "1.4.1",
+  "cfncluster_version": "1.4.2",
   "cfncluster_cookbook_version": "1.4.1",
```
bwbarrett: did we not have to bump the cookbook version as well?

rajachan: Cookbook is still at 1.4.1 ( CfnCluster v1.4.1 release used cfncluster-cookbook v1.4.0 as-is).

bwbarrett: ah, ok.

PR #103, opened 2018-05-03
Poll on EBS Volume Attachment

sean-smith:
Previously we set a timeout of 60 seconds for attaching the
ebs volume to the instance. This changes that to poll every
5 seconds on the volume status. Volume status can be `attaching`,
`attached`, `detaching`, `detached` or `busy`.

See: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_VolumeAttachment.html

### DIFF
--- a/files/default/attachVolume.py
+++ b/files/default/attachVolume.py
@@ -4,6 +4,7 @@ import sys
 import parted
 import urllib2
 import boto.ec2
+import time
 
 # Get volumeId
 try:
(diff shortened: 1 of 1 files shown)


files/default/attachVolume.py
```diff
@@ -34,3 +35,14 @@
 # Attach the volume
 dev = availableDevices[0].replace('xvd', 'sd')
 conn.attach_volume(volumeId,instanceId,dev)
+
+# attaching | attached | detaching | detached | busy
+vol = conn.get_all_volumes([volumeId])[0]
+while vol.attachment_state() != "attached":
+    if vol.attachment_state() in ["busy" or "detached"]:
```
bwbarrett: Is there documentation that attachment_state is guaranteed to be attaching before the first call after attach_volume()?  Or did we just add another race to bite us later?

bwbarrett: I think you're missing my point.  I'm asking if it's possible for the `volume attachment_state()` to return `detached` after `get_all_volumes()` returned successfully.  If it is, your code will throw an error.  If the API guarantees you'll at least be as far as `attaching`, then your code looks good to me.  I just want to know the answer to that question...

sean-smith: I checked with ebs and the volume can never be in `detached` state after a call to attached. I've added a timeout of 180 seconds to avoid an infinite loop.

PR #104, opened 2018-05-18
fix kitchen tests

lukeseawalker:
- reboot after packages update:
  this solves the failing kitchen tests that weren't able to install the
  nvidia driver. The nvidia driver couldn't be installed because when
  doing a package update, there was a mismatch between the kernel library
  version and the version of the running kernel. Doing a reboot, it aligns
  those version, permitting to build and install the nvidia driver
  (with the dkms option)

- handle reboot on ubuntu:
  on ubuntu, the /tmp folder is erased by default.
  This patch prevent the folder to be erased during the reboot after
  packages upgrades. The configuration is then removed after the process
  restart after the reboot

- fix python depencency with poise-python:
  Poise-python 1.7.0 is up with full support for both Pip 10 and Chef 14.
  https://github.com/poise/poise-python/issues/107

### DIFF
--- a/recipes/_update_packages.rb
+++ b/recipes/_update_packages.rb
@@ -27,23 +27,10 @@ when 'debian'
   end
 end
 
-if tagged?('rebooted')
-  if node['platform'] == 'ubuntu'
-    if node['platform_version'] == "16.04"
-      file '/etc/tmpfiles.d/tmp.conf' do
-        action :delete
-        only_if { File.exist? '/etc/tmpfiles.d/tmp.conf' }
-      end
-    end
-    if node['platform_version'] == "14.04"
-      execute 'sed' do
-        command "sed -i s/TMPTIME=1/#TMPTIME=0/g /etc/default/rcS"
-      end
-    end
-  end
-end
-
+# Reboot after packages update
 if !tagged?('rebooted') && node['cfncluster']['update_packages']['reboot'] == 'true'
+  # On Ubuntu, the /tmp folder is erased by default after a reboot and its lifecycle is managed
+  # by different tools, depending on the Ubuntu version
   if node['platform'] == 'ubuntu'
     if node['platform_version'] == "16.04"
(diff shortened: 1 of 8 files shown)


recipes/_update_packages.rb
```diff
@@ -26,3 +26,45 @@
...
 end
+
+# Reboot after packages update
+if !tagged?('rebooted') && node['cfncluster']['update_packages']['reboot'] == 'true'
+  # On Ubuntu, the /tmp folder is erased by default after a reboot and its lifecycle is managed
+  # by different tools, depending on the Ubuntu version
+  if node['platform'] == 'ubuntu'
+    if node['platform_version'] == "16.04"
+      file '/etc/tmpfiles.d/tmp.conf' do
+        content 'd /tmp 1777 root root 1d'
```
sean-smith: What does this do?

recipes/_update_packages.rb
```diff
@@ -26,3 +26,45 @@
...
+      end
+    end
+    if node['platform_version'] == "14.04"
+      execute 'sed' do
+        command "sed -i s/#TMPTIME=0/TMPTIME=1/g /etc/default/rcS"
```
sean-smith: What does this do?

recipes/_update_packages.rb
lukeseawalker: It prevents the /tmp folder to be erased after reboot, which is the default on Ubuntu. /tmp folder is where all the kitchen test environment is copied.
https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html

recipes/_update_packages.rb
lukeseawalker: It prevents the /tmp folder to be erased after reboot, which is the default on Ubuntu. /tmp folder is where all the kitchen test environment is copied.
http://manpages.ubuntu.com/manpages/trusty/man5/rcS.5.html

PR #109, opened 2018-05-25
Move packer scripts/configs into a new amis directory

enricou:
Fix issue: #63 

- Move all the packer related scripts into an *amis* subdirectory.
- Fix `build_ami.sh` script to be executed from any folder.
- Use `template_dir` instead of `pwd` in packer templates to remove working-directory dependency.

Signed-off-by: Enrico Usai <usai@amazon.com>


amis/build_ami.sh
```diff
@@ -32,8 +33,8 @@ fi
 
 RC=0
 
-rm -rf ../vendor/cookbooks || RC=1
-berks vendor ../vendor/cookbooks || RC=1
+rm -rf ${cwd}/../../vendor/cookbooks || RC=1
```
mauri-melato: Double quotes around the paths

amis/build_ami.sh
```diff
@@ -9,6 +9,7 @@ build_date=$4
 
 available_os="centos6 centos7 alinux ubuntu1404 ubuntu1604"
 available_regions="eu-west-1,eu-west-2,ap-southeast-1,ap-southeast-2,eu-central-1,ap-northeast-1,ap-northeast-2,us-west-2,sa-east-1,us-west-1,us-east-2,ap-south-1,ca-central-1"
+cwd="$(dirname $0)"
```
mauri-melato: What about defining an var  `BUILD_PATH="${cwd}/../.."` to be used in the code below and inside the json templates. 

Within a json template we can map the variable
```
{
 "variables" : {
     [...]
     "build_path" : "{{env `BUILD_PATH`}}",
 }
[...]
```
and then use `{{build_path}}/vendor/cookbooks`  instead of `{{template_dir}}/../../vendor/cookbooks`

amis/build_ami.sh
```diff
@@ -43,28 +44,28 @@ fi
 case $os in
 all)
   for x in $available_os; do
-    packer build -machine-readable -var-file=packer_variables.json packer_$x.json
+    packer build -machine-readable -var-file=${cwd}/packer_variables.json ${cwd}/packer_$x.json
     RC=$?
   done
   ;;
 centos6)
```
mauri-melato: since all the OSes are executing the same command we could compact the code
`centos6 | centos7 | alinux | ...`

enricou:
I added a new patch to explain how execute the `build_ami.sh` script.

I tested the script from a personal EC2 instance in _us-east-1_ region, for _alinux_ os, and from an EC2 instance in the CfnCluster private account, for _centos6_ and _centos7_ os, because the CentOS AMIs IDs used in the CentOS packer templates are private.

amis/build_ami.sh
```diff
@@ -1,5 +1,22 @@
...
+# This script creates a set of CfnCluster AMIs.
+#
+# The following variables must be exported in the environment:
+#AWS_FLAVOR_ID=<instance-type>
+#AWS_VPC_ID=<us-east-1-vpc-id>
+#AWS_SUBNET_ID=<us-east-1-subnet-id>
+#NVIDIA_ENABLED=<no|yes>
+#
+# NOTE: The VPC and the Subnet must be in the us-east-1 region, because the packer templates refer to
+# AMI IDs from this region. Moreover, the CentOS AMIs are private.
```
mauri-melato: ...CentOs AMIs are private to the CfnCluster AWS account.

PR #114, opened 2018-06-04
Add a loop to get block device attribute from EC2 instance

lukeseawalker:
It could happens that also if a volume appears to be attached to the
intance, the block device mapping attribute could get some time to be
updated.

A syslog has been added too

### DIFF
--- a/files/default/ec2_dev_2_volid.py
+++ b/files/default/ec2_dev_2_volid.py
@@ -4,24 +4,26 @@ import boto.ec2
 import urllib2
 import sys
 import os
+import syslog
+import time
 
 # Get dev
 try:
-  dev = str(sys.argv[1])
+    dev = str(sys.argv[1])
 except IndexError:
-  print "Provide block device i.e. xvdf"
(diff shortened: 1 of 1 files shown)


files/default/ec2_dev_2_volid.py
```diff
@@ -4,24 +4,26 @@
...
+import time
 
 # Get dev
 try:
-  dev = str(sys.argv[1])
+    dev = str(sys.argv[1])
 except IndexError:
-  print "Provide block device i.e. xvdf"
-  sys.exit(1)
+    syslog.syslog(syslog.LOG_ERR, "%s - Provide block device i.e. xvdf" % os.path.basename(__file__))
```
mauri-melato: (minor) since syslog.sysog call is used multiple times and follows a format could you provide a function for it? the caller should just pass the message to append

lukeseawalker: syslog.syslog was already logging the script name, so no format is needed. Only the message is passed to the function

PR #115, opened 2018-06-04
Refactor recipes to fix Ubuntu interface renaming in kitchen steps

enricou:
The https://github.com/awslabs/cfncluster-cookbook/pull/70 added shell commands to avoid the systemd v197 interface naming
and use the ethX syntax, for the Ubuntu 16 packer template.
With this patch we are fixing the kitchen part too.

The new _default_pre recipe is shared from packer and kitchen and
includes both the _update_packages and the new _setup_network recipes.
This step will be execute before the reboot steps.

Tests:
* packer creation
* kitchen test

Ref:
https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

Signed-off-by: Enrico Usai <usai@amazon.com>

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -13,13 +13,13 @@ platforms:
 suites:
   - name: default
     run_list:
-      - recipe[cfncluster::_update_packages]
+      - recipe[cfncluster::_default_pre]
       - recipe[cfncluster::default]
     attributes:
 
   - name: default_gpu
     run_list:
-      - recipe[cfncluster::_update_packages]
+      - recipe[cfncluster::_default_pre]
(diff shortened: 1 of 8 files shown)


bwbarrett:
I don't think this was the right solution to the problem.  It means we need to reboot in order to get the device names to work the way we want.  In a custom AMI, that's a pretty miserable customer experience.  It seems like a much better solution is to remove our dependency on naming ethernet devices ethX and adopt the new hotness of Linux device naming.  @sean-smith and I worked through how to get all the data from IMDS so that we won't care; that seems much better.

PR #126, opened 2018-06-18
Install or update AWS CLI into AMIs for all distro

mauri-melato:
When building AMIs Ubuntu 14 and 16 Linux distributions get installed
with the latest version of the AWS CLI. Amazon, CentOS 6, 7 AMIs
are built from pre-cooked AMIs with already installed a different and
older version of the AWS-CLI.
This change installs and updates the AWS CLI to tested version 1.15.40,
for all distros when building the AMIs for CfnCluster.
It also executes the AWS-CLI update during the provisioning of the
master and compute nodes.

Signed-off-by: Maurizio Melato <mmelato@amazon.com>

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -59,7 +59,10 @@ cookbook_file 'CfnCluster-License-README.txt' do
 end
 
 # Install AWSCLI
-python_package 'awscli'
+python_package 'awscli' do
+  action :upgrade
+  version '1.15.40'
+end
 
 # TODO: update nfs receipes to stop, disable nfs services
 include_recipe "nfs"


bwbarrett:
What's the reason for this?  I'm not sure I'm ok with us taking ownership of aws cli package unless absolutely necessary.  It's yet another package that we'd own patching for.

mauri-melato:
This is a (temporary?) solution to update and align AWS-CLI distribution to a tested version into our AMIs that otherwise would come all with a different, older and unreliable version.
In particular, the un-crontrolled management recently brought to an issue with the cfncluster centos7 image where we got this error:
```bash
[centos7]# aws
Traceback (most recent call last):
  File "/bin/aws", line 19, in <module>
    import awscli.clidriver
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 25, in <module>
    from botocore.history import get_global_history_recorder
ImportError: No module named history
```
that's due to having first installed the `awscli` package [from epel](https://github.com/awslabs/cfncluster-cookbook/blob/5aa8a30662cf10ae83f617518cba077b17c44e0e/amis/centos-upgrade-first-stage.sh#L67-L72):
```bash
# yum installed |grep ...
awscli.noarch                           1.11.133-1.el7                 @epel
python2-boto.noarch                     2.45.0-3.el7                   @epel
python2-botocore.noarch                 1.6.0-1.el7                    @epel
```
and then updated it from the [base repo](https://github.com/awslabs/cfncluster-cookbook/blob/5aa8a30662cf10ae83f617518cba077b17c44e0e/recipes/_update_packages.rb#L17-L20):
```bash
# yum list installed |grep ...
awscli.noarch                           1.14.28-5.el7                  @base
python2-boto.noarch                     2.45.0-3.el7                   @epel
python2-botocore.noarch                 1.6.0-1.el7                    @epel
```
At this point we silently incurred into the issue reported in [CentOS issue tracker](https://bugs.centos.org/view.php?id=14756), [community](https://www.centos.org/forums/viewtopic.php?t=67020) and [aws-cli](https://github.com/aws/aws-cli/issues/2999) with a broken dependency,  where `python2-botocore-1.60-1.el7` installed from epel is not compatible with `awscli` from base.

mauri-melato:
Btw if there is not a strong reason to use centos package manager we should also install awscli using pip in the first place, as it's done for centos6.

bwbarrett:
I think this is a bad path and would strongly suggest you use the AWSCLI from the base os.  I recognize that's hard, but you're setting up customers for pain if you install awscli from outside the base repos.  I get this is more work, but you don't want to own maintenance of more software than you absolutely have to.

PR #134, opened 2018-06-26
Get MAC address from Instance Metadata avoiding to search for device name

enrico-usai:
Ubuntu 16 uses the systemd v197 interface naming so the first network interface name is not _eth0_.
There was a trick in the `_setup_network.rb` to avoid the v197 naming and use _ethX_ syntax. 

With this commit we are also removing this part. See https://github.com/awslabs/cfncluster-cookbook/pull/115

By using the **mac** metadata we are sure to get the instance's media access control (MAC) address.
In cases where multiple network interfaces are present, this refers to the eth0 device (the device for which the device number is 0).

This patch solves https://github.com/awslabs/cfncluster-cookbook/issues/116

### Tests 
I tested packer and kitchen with different OS and instance types

#### Command used to obtain the interface name
```
centos6
[centos@ip-172-30-1-46 ~]$ ip -o link | grep -i 0a:67:01:cf:69:00 | cut -d' ' -f2 | sed 's/://'
eth0

centos7
[centos@ip-172-30-1-248 ~]$  ip -o link | grep -i 0a:1f:bf:6e:38:c0 | cut -d' ' -f2 | sed 's/://'
eth0

alinux
[ec2-user@ip-10-0-20-98 ~]$ ip -o link | grep -i 02:ba:ff:ea:ba:52 | cut -d' ' -f2 | sed 's/://'
eth0

ubuntu14
[ubuntu@ip-10-0-18-187:~]$ ip -o link | grep -i 02:c2:fe:51:67:58 | cut -d' ' -f2 | sed 's/://'
eth0

ubuntu16
ubuntu@ip-10-0-20-170:~$ ip -o link | grep 02:91:40:a5:3e:8c | cut -d' ' -f2 | sed 's/://'
ens3
```

#### Kitchen
```
test-kitchen-ec2-sge-config-MasterServer-amazon-linux-latest
c3.8xlarge
Found Interface name: eth0 associated to MAC 12:10:17:d6:8a:5e

test-kitchen-ec2-sge-config-MasterServer-centos-6-minimal
t2.large
Found Interface name: eth0 associated to MAC 12:33:42:fa:c6:e8

test-kitchen-ec2-sge-config-MasterServer-centos-7-minimal
t2.large
Found Interface name: eth0 associated to MAC 12:2c:54:f1:87:04

test-kitchen-ec2-sge-config-MasterServer-ubuntu-14-04-lts
t2.large
Found Interface name: eth0 associated to MAC 12:bc:4a:2b:1f:de

test-kitchen-ec2-sge-config-MasterServer-ubuntu-16-04-lts
c4.xlarge
Found Interface name: ens3 associated to MAC 12:37:19:e8:c9:8e
```


enrico-usai:
This doesn't work for centos7, because `ifconfig` output format is different from other OSes.

bwbarrett:
It looks like you used the GitHub "update branch" option at some point, so you have a merge commit in your pull request.  Please clean that up before you merge into mainline.

PR #145, opened 2018-07-12
Add idle_time parameter to nodewatcher config file

fnubalaj:
Add idle_time parameter from cfncluster compute server parameters

Signed-off-by: Balaji Sridharan <fnubalaj@amazon.com>


templates/default/nodewatcher.cfg.erb
```diff
@@ -1,4 +1,5 @@
 [nodewatcher]
 region = <%= node['cfncluster']['cfn_region'] %>
 scheduler = <%= node['cfncluster']['cfn_scheduler'] %>
-proxy = <%= node['cfncluster']['cfn_proxy'] %>
\ No newline at end of file
+proxy = <%= node['cfncluster']['cfn_proxy'] %>
+idle_time = <%= node['cfncluster']['cfn_scale_down_idle_time'] %>
```
mauri-melato: should we call this in a more complete way, `scaledown_idletime` or anyway something more than `idle_time` ?

fnubalaj: Yes, scaledown_idletime sounds good.

templates/default/nodewatcher.cfg.erb
```diff
@@ -1,4 +1,5 @@
 [nodewatcher]
 region = <%= node['cfncluster']['cfn_region'] %>
 scheduler = <%= node['cfncluster']['cfn_scheduler'] %>
-proxy = <%= node['cfncluster']['cfn_proxy'] %>
\ No newline at end of file
+proxy = <%= node['cfncluster']['cfn_proxy'] %>
+scale_down_idle_time = <%= node['cfncluster']['cfn_scale_down_idle_time'] %>
```
mauri-melato: If it's not annoying and it's coherent with the rest of the parameters names I'd go for using the name `cfn_scaledown_idletime`

fnubalaj: I wanted to avoid that since scaledown is not one word, but I can accommodate that change. It would have to be changed in every repository.

PR #152, opened 2018-07-20
Remove Publish Pending and Run Scaling

sean-smith:
Remove old publish pending scripts and configure
scaling script.


recipes/_master_base_config.rb
```diff
@@ -155,6 +155,14 @@
   not_if { ::File.exist?("/home/#{node['cfncluster']['cfn_cluster_user']}/.ssh/known_hosts") }
 end
 
+# Install scaling.cfg
```
mauri-melato: A couple of doubts:
* about naming shouldn't we call it in a more specific way, like _aws_scaling_ ?
* is there a reason to not grouping the scaling [and sqswatcher?] configuration under the `/etc/cfncluster` dir ?

templates/default/cfncluster_supervisord.conf.erb
```diff
@@ -8,10 +8,15 @@ command = /usr/bin/sqswatcher
 redirect_stderr = true
 stdout_logfile = /var/log/sqswatcher
 
+[program:scaling]
+command = /usr/local/bin/scaling
```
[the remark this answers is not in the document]
sean-smith: changed.

recipes/_master_base_config.rb
sean-smith: 1. Let me propose a new name: `jobwatcher`, since all it's really doing is watching the queue, I think this is most appropriate.
2. I've added it to `/etc/` since both sqswatcher and nodewatcher are there.

PR #154, opened 2018-08-09
Use Source AMI Filter to get latest base AMI

sean-smith:
Uses `source_ami_filter` to find latest ami in region.

### DIFF
--- a/amis/build_ami.sh
+++ b/amis/build_ami.sh
@@ -13,19 +13,20 @@
 #
 # Usage: build_ami.sh <os> <region> [private|public] [<build-date>]
 #   os: the os to build (supported values: all|centos6|centos7|alinux|ubuntu1404|ubuntu1604)
-#   region: region to build (supported values: all|us-east-1|...)
+#   partition: partition to build in (supported values: commercial|govcloud)
+#   region: region to copy ami too (supported values: all|us-east-1|us-gov-west-1|...)
 #   private|public: specifies AMIs visibility (optional, default is private)
 #   build-date: timestamp to append to the AMIs names (optional)
 
 set -e
 
 os=$1
-region=$2
-public=$3
-build_date=$4
+partition=$2
+region=$3
+public=$4
+build_date=$5
 
 available_os="centos6 centos7 alinux ubuntu1404 ubuntu1604"
-available_regions="$(aws ec2 --region us-east-1 describe-regions --query Regions[].RegionName --output text | tr '\t' ',')"
 cwd="$(dirname $0)"
 export VENDOR_PATH="${cwd}/../../vendor/cookbooks"
 
@@ -36,8 +37,8 @@ if [ "x${os}" == "x" ]; then
 fi
 
 if [ "x${region}" == "x" ]; then
-  echo "Must provide AWS region to build for."
-  echo "Options: us-east-1 all"
+  echo "Must provide AWS region to copy ami into"
+  echo "Options: all us-east-1 us-gov-west-1 ..."
   exit 1
 fi
 
@@ -45,8 +46,18 @@ if [ "${public}" == "public" ]; then
   export AMI_PERMS="all"
 fi
 
-# NOTE: the AMI is always built in us-east-1 and then copied to the specified regions
+if [ "${partition}" == "commercial" ]; then
+  export AWS_REGION="us-east-1"
+elif [ "${partition}" == "govcloud" ]; then
+  export AWS_REGION="us-gov-west-1"
+else
+  echo "Must provide AWS partition to build for."
+  echo "Options: commercial govcloud"
+  exit 1
+fi
+
 if [ "${region}" == "all" ]; then
+  available_regions="$(aws ec2 --region ${AWS_REGION} describe-regions --query Regions[].RegionName --output text | tr '\t' ',')"
   export BUILD_FOR=${available_regions}
 else
   export BUILD_FOR=${region}
--- a/amis/packer_alinux.json
+++ b/amis/packer_alinux.json
@@ -13,13 +13,22 @@
     "instance_type" : "{{env `AWS_FLAVOR_ID`}}",
     "subnet_id" : "{{env `AWS_SUBNET_ID`}}",
     "vpc_id" : "{{env `AWS_VPC_ID`}}",
-    "ami_name_prefix" : "{{env `AMI_NAME_PREFIX`}}"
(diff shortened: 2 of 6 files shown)


amis/build_ami.sh
```diff
@@ -25,7 +25,7 @@ public=$3
 build_date=$4
 
 available_os="centos6 centos7 alinux ubuntu1404 ubuntu1604"
-available_regions="$(aws ec2 --region us-east-1 describe-regions --query Regions[].RegionName --output text | tr '\t' ',')"
+available_regions="$(aws ec2 --region ${region} describe-regions --query Regions[].RegionName --output text | tr '\t' ',')"
```
lukeseawalker: ${region} == all is going to break line 50

amis/packer_alinux.json
```diff
@@ -13,13 +13,15 @@
...
+    "region" : "{{env `AWS_REGION`}}",
+    "source_ami" : "{{env `ALINUX_IMAGE_ID`}}"
   },
   "builders" : [
     {
       "type" : "amazon-ebs",
-      "region" : "us-east-1",
-      "source_ami" : "ami-55ef662f",
+      "region" : "{{user `region`}}",
+      "source_ami" : "{{user `source_ami`}}",
```
lukeseawalker: Just asking, was not possible to leave a default like
"source_ami" : "{{user `source_ami`}}" | "ami-55ef662f",

sean-smith: @lukeseawalker aforementioned PR was never merged. Looks like it's still not supported

amis/build_ami.sh
sean-smith: I'll default it to `us-east-1` if it's set to `all`

amis/packer_alinux.json
lukeseawalker: ok, pity. I think it's better to move the defaults into the calling script, build_ami.sh, adding a new input parameter "build_region" and renaming the old "region" parameter to "target_region"

```
os=$1
region=$2
public=$3
build_date=$4
```
become
```
os=$1
build_region=$2
target_region=$3
public=$4
build_date=$5
```

In this way we can have build_region = us-east-1 or PDT as input parameter. Then we could check if ALINUX_IMAGE_ID or the others are not set and set the default value based on build_region. Something like
```
if ALINUX_IMAGE_ID not set {
   if build_region == us-east-1 then ALINUX_IMAGE_ID = ami-55ef662f
   if build_region  == PDT then ALINUX_IMAGE_ID = ami-xxx
}

if ALINUX_IMAGE_ID empty, then error
Repeat for the others *_IMAGE_ID
```

In this way we can have the default value for the image id (based on the build_region) but also the possibility to override that value, if needed.
What do you think @sean-smith?

sean-smith: I'd prefer to have a seperate file, like `default/attributes.rb` that just contains the ami id's. That'll allow us to programmatically update the base ami easily.

lukeseawalker: ok for the separate file, but we are not in the "domain" where we can use .rb files, as we are in the "packer" domain. There is a "packer_variables.json" file where we can store variable values, but as long as the packer syntax could not give as the possibility to use a value and if this value is not set to fallback to a default value, I'd prefer to move this logic to the calling script. Then we can use a properties file which contain the default values for the amis and source this file from the script. Something like
```
packer_amis.properties

US_EAST_1_DEFAULT_ALINUX_IMAGE_ID=ami-55ef662f
GOVCLOUD_DEFAULT_ALINUX_IMAGE_ID=ami-xxx
```
then from the build_ami.sh
```
source packer_amis.properties

if ALINUX_IMAGE_ID not set {
   if build_region == us-east-1 then ALINUX_IMAGE_ID = ${US_EAST_1_DEFAULT_ALINUX_IMAGE_ID}
   if build_region  == PDT then ALINUX_IMAGE_ID = ${GOVCLOUD_DEFAULT_ALINUX_IMAGE_ID}
}

if ALINUX_IMAGE_ID empty, then error
Repeat for the others *_IMAGE_ID
```

sean-smith: I'd much prefer to use `source_ami_filter` and not explicitly use the ami name. https://www.packer.io/docs/builders/amazon-ebs.html#source_ami_filter

amis/build_ami.sh
```diff
@@ -45,8 +44,9 @@ if [ "${public}" == "public" ]; then
   export AMI_PERMS="all"
 fi
 
-# NOTE: the AMI is always built in us-east-1 and then copied to the specified regions
+# NOTE: if region is all then the ami is built in us-east-1 and then copied to the specified regions
```
lukeseawalker: This sentence is wrong at the moment.
We are not setting AWS_REGION anywhere. If we want this value to be automatic assigned, we need to make a decision like:
- if ${region} = all or any other region except govcloud then AWS_REGION=us-east-1, meaning that we always want to build on us-east-1 and then copy the ami to other/s region (this is like we are doing right now)
- if ${region} = govcloud then AWS_REGION=govcloud, meaning that ami build for govcloud need to be built in govcloud

amis/packer_alinux.json
```diff
@@ -13,13 +13,22 @@
...
-      "source_ami" : "ami-55ef662f",
+      "region" : "{{user `region`}}",
+      "source_ami_filter": {
+        "filters": {
+          "virtualization-type": "hvm",
+          "name": "amzn-ami-hvm-*.*.*.*-x86_64-gp2",
+          "root-device-type": "ebs",
+          "owner-alias": "amazon"
+        },
+        "most_recent": true
```
lukeseawalker: we all need to agree that we are now using the latest ami

amis/packer_ubuntu1404.json
```diff
@@ -13,15 +13,24 @@
...
       "type" : "amazon-ebs",
-      "region" : "us-east-1",
+      "region" : "{{user `region`}}",
+      "source_ami_filter": {
+        "filters": {
+          "virtualization-type": "hvm",
+          "name": "ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*",
+          "root-device-type": "ebs"
+        },
+        "owners": ["099720109477", "513442679011"],
```
lukeseawalker: is 513442679011 the owner id for govcloud?

sean-smith: yes

PR #155, opened 2018-08-15
Turn off Ganglia Support

sean-smith:
Ganglia can be turned off by adding to the cluster config:

```
extra_json = { "cfncluster" : { "ganglia_enabled" : "no" } }
```


dougalb:
Why disable Ganglia by default? It is a valuable piece of data for a running cluster.

dougalb:
If that’s the concern, one idea is you could make it have a Security Group
that matches SSHFrom?

lukeseawalker:
I'd prefer if the patch give you the choice to start or not ganglia, leaving the installation steps unchanged.

PR #157, opened 2018-08-17
Set accounting_summary to TRUE for multi node job journaling

Elveskevtar:
For jobs that run on multiple nodes, each node will have an exit_status and each node will write its output to the journaling file on the master node. This makes it so that only the combined process writes an entry to the journaling file (if even one node fails, the job fails and returns the failed exit code).

Signed-off-by: Elveskevtar <kevtar@gmail.com>

### DIFF
--- a/recipes/_master_sge_config.rb
+++ b/recipes/_master_sge_config.rb
@@ -55,6 +55,17 @@ bash "add_host_as_master" do
   ADDHOST
 end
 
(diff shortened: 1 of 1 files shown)


sean-smith:
please test on all OSes.

Elveskevtar:
Tested on alinux, centos6, centos7, ubuntu1404, and ubuntu1604. All good!

PR #180, opened 2018-09-20
Fix proxy support

enrico-usai:
+ Add proxy configuration parameter to boto clients for all the runtime python scripts
+ Add possibility to download a custom node package by a proxy

### DIFF
--- a/files/default/attachVolume.py
+++ b/files/default/attachVolume.py
@@ -6,51 +6,59 @@ import urllib2
 import boto3
 import time
 
-# Get volumeId
-try:
-  volumeId = str(sys.argv[1])
-except IndexError:
-  print "Provide an EBS volume ID to attach i.e. vol-cc789ea5"
-  sys.exit(1)
 
-# Get instance ID
-instanceId = urllib2.urlopen("http://169.254.169.254/latest/meta-data/instance-id").read()
-
-# Get region
-region = urllib2.urlopen("http://169.254.169.254/latest/meta-data/placement/availability-zone").read()
-region = region[:-1]
-
-# Generate a list of system paths minus the root path
-paths = [ device.path for device in parted.getAllDevices() ]
-
-# List of possible block devices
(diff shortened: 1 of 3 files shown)


files/default/attachVolume.py
```diff
@@ -34,8 +36,17 @@ def main():
...
     availableDevices = [a for a in blockDevices if a not in paths]
 
+    # Parse configuration file to read proxy settings
+    config = ConfigParser.RawConfigParser()
+    config.read('/etc/boto.cfg')
+    proxy_config = Config()
+    if config.has_option('Boto', 'proxy') and config.has_option('Boto', 'proxy_port'):
+        proxy = config.get('Boto', 'proxy')
+        proxy_port = config.get('Boto', 'proxy_port')
+        proxy_config = Config(proxies={'https': proxy + ':' + proxy_port})
```
sean-smith: use `str.format()`

files/default/attachVolume.py
```diff
@@ -34,8 +36,17 @@ def main():
     # List of available block devices after removing currently used block devices
     availableDevices = [a for a in blockDevices if a not in paths]
 
+    # Parse configuration file to read proxy settings
+    config = ConfigParser.RawConfigParser()
+    config.read('/etc/boto.cfg')
```
sean-smith: Also fyi the `/etc/boto.cfg` is an old boto2 config file that we didn't remove when upgrading. Since we're not using it for anything else, I think it's fine to still rely on it.

recipes/base_install.rb
```diff
@@ -105,13 +105,19 @@
...
-      "cd /tmp &&" \
-      "curl -v -L -o /tmp/cfncluster-node.tgz #{node['cfncluster']['custom_node_package']} &&" \
-      "tar -xzf /tmp/cfncluster-node.tgz &&" \
-      "cd /tmp/cfncluster-node-* &&" \
-      "sudo /usr/bin/python setup.py install"
+  bash "install cfncluster-node" do
+    cwd '/tmp'
+    code <<-EOH
+      if [ "#{node['cfncluster']['cfn_proxy']}" != "NONE" ]; then
+        export http_proxy=#{node['cfncluster']['cfn_proxy']}
```
sean-smith: Why is this needed? You can just source `/tmp/proxy.sh`. See https://github.com/awslabs/cfncluster/blob/44ab57d8abd1444d45f5ede0371f6aa96d0c987f/cloudformation/cfncluster.cfn.json#L3187

files/default/attachVolume.py
enrico-usai: Yes, I see. 
As you probably know, in boto2 the proxy configuration was automatically loaded by the file, in boto3 we have to pass it explicitely by a `Config` parameter. 

We could remove it because it isn't required for boto3 but we still need a place from which to load the proxy settings and I agree with you to use this file.

PR #181, opened 2018-09-20
Fix error not being returned if node package installation fail

lukeseawalker:

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -66,6 +66,6 @@ end
 def pip_install_package(package, version)
   command = Mixlib::ShellOut.new("pip install #{package}==#{version}").run_command
   unless command.exitstatus.zero?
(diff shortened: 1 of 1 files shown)


recipes/base_install.rb
```diff
@@ -119,7 +119,10 @@
     # For CentOS 6 use shell_out function in order to have a correct PATH needed to compile cfncluster-node dependencies
     ruby_block "pip_install_cfncluster_node" do
       block do
-        pip_install_package('cfncluster-node', node['cfncluster']['cfncluster-node-version'])
+        command = pip_install_package('cfncluster-node', node['cfncluster']['cfncluster-node-version'])
```
mauri-melato: I would move the error checking & failure code into the `pip_install_package()` so every time we call it there is the check without having to remember it.

PR #189, opened 2018-10-04
Adding support for multiple EBS volumes

rexcsn:
- Takes in comma separated list of shared directory name and volume-id as string through dna.json, parse into list in program.

- Run for each loop for each volume-id, and carry out all attach and mounting procedure in _master_base_config. Note: it is crucial the execute or ruby block within the for each loop to have a unique name, so that the workflow is independent for each volume-id.

- Run for each loop for each shared directory, and carry out all NFS mounting procedure in _compute_base_config.

- Note: it is crucial to use mountpoints throughout the program, for example, if the user want to mount to a share directory called "apps", we need to mount to "/apps" throughout the program. Mounting to "apps", without the slash, will cause lots of error and prevent AutoScalingGroup from creating.

- This change is to prepare cfncluster for mounting multiple EBS volumes. This cookbook modification should remain valid for arbitary number of EBS volumes, as long as the input format is consistent.

Signed-off-by: Rex Chen <shuningc@amazon.com>

### DIFF
--- a/recipes/_compute_base_config.rb
+++ b/recipes/_compute_base_config.rb
@@ -13,26 +13,11 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Created shared mount point
-directory node['cfncluster']['cfn_shared_dir'] do
-  mode '1777'
-  owner 'root'
-  group 'root'
-  recursive true
-  action :create
-end
 
 node.default['cfncluster']['cfn_master'] = node['cfncluster']['cfn_master'].split('.')[0]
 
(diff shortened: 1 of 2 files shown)


recipes/_compute_base_config.rb
```diff
@@ -65,6 +50,33 @@
   shell '/bin/bash'
 end
 
+# Parse shared directory info and turn into an array
+_shared_dir_array = node['cfncluster']['cfn_shared_dir'].split(',')
+_shared_dir_array.each_with_index do |dir, index|
```
demartinofra: I'm very new to Ruby but in general I would avoid modifying a collection while iterating over it. 
What about using map or map! (for in place changes)?

PR #199, opened 2018-10-16
Install jq as prerequisite inside User Data

lukeseawalker:
jq is used to merge dna.json with extra.json before launching
chef-client. So it needs to be installed before the chef run.
This change is needed for custom AMI which do not have the jq package
pre-installed


sean-smith:
Are you installing it elsewhere?

PR #220, opened 2018-11-14
Set PATH as in the UserData script of the CloudFormation template

lukeseawalker:
In this way, during the test we are sure we do have the same PATH set
during the executing of the UserData script


recipes/tests.rb
```diff
@@ -90,6 +90,8 @@
 bash 'execute jq' do
   cwd Chef::Config[:file_cache_path]
   code <<-JQMERGE
+    # Set PATH in the same way it is set into UserData script
```
mauri-melato: Set PATH as in the UserData script of the CloudFormation template

PR #221, opened 2018-11-15
Changes to support EFS

rexcsn:
- Added efs_mount.rb recipe to create EFS shared_dir and mount EFS via nfs4 if the user is using EFS (shared_dir for EFS specified)
- Modified _master_base_config.rb to include efs_mount
- Modified _compute_base_config.rb to include efs_mount

Signed-off-by: Rex Chen <shuningc@amazon.com>

### DIFF
--- a/recipes/_compute_base_config.rb
+++ b/recipes/_compute_base_config.rb
@@ -26,6 +26,9 @@ node.default['cfncluster']['cfn_master'] = node['cfncluster']['cfn_master'].spli
 
 nfs_master = node['cfncluster']['cfn_master']
 
+# Mount EFS directory with efs_mount recipe
+include_recipe 'aws-parallelcluster::efs_mount'
+
 # Mount /home over NFS
 mount '/home' do
   device "#{nfs_master}:/home"
--- a/recipes/_master_base_config.rb
+++ b/recipes/_master_base_config.rb
@@ -28,6 +28,9 @@ end
 # Get VPC CIDR
 node.default['cfncluster']['ec2-metadata']['vpc-ipv4-cidr-block'] = get_vpc_ipv4_cidr_block(node['macaddress'])
 
+# Mount EFS directory with efs_mount recipe
+include_recipe 'aws-parallelcluster::efs_mount'
+
 # Parse shared directory info and turn into an array
 shared_dir_array = node['cfncluster']['cfn_shared_dir'].split(',')
 shared_dir_array.each_with_index do |dir, index|
--- a/recipes/efs_mount.rb
+++ b/recipes/efs_mount.rb
@@ -1,0 +1,45 @@
+#
+# Cookbook Name:: aws-parallelcluster
+# Recipe:: efs_mount
+#
+# Copyright 2013-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Get shared_dir path and mount EFS filesystem
+efs_shared_dir = "/" + node['cfncluster']['cfn_efs_shared_dir'].split(',')[0]
+
+# Check to see if EFS is created
+if efs_shared_dir != "/NONE"
+  # Create the EFS shared directory
+  directory efs_shared_dir do
+    owner 'root'
+    group 'root'
+    mode '1777'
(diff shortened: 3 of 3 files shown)


recipes/_compute_base_config.rb
```diff
@@ -26,6 +26,31 @@
 
 nfs_master = node['cfncluster']['cfn_master']
 
+# EFS
+# Get shared_dir path and mount EFS filesystem
+efs_shared_dir = "/"+node['cfncluster']['cfn_efs_shared_dir'].split(',')[0]
```
enrico-usai: Missing spaces around the `+`. See [SpaceAroundOperators](https://www.rubydoc.info/gems/rubocop/0.10.0/Rubocop/Cop/Style/SpaceAroundOperators).

recipes/_compute_base_config.rb
```diff
@@ -26,6 +26,31 @@
...
+
+# Check to see if EFS is created
+if efs_shared_dir != "/NONE"
+    # Create the EFS shared directory
```
enrico-usai: Please use 2 (not 4) spaces for indentation. See [IndentationWidth](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Layout/IndentationWidth).

recipes/_master_base_config.rb
```diff
@@ -28,6 +28,38 @@
 # Get VPC CIDR
 node.default['cfncluster']['ec2-metadata']['vpc-ipv4-cidr-block'] = get_vpc_ipv4_cidr_block(node['macaddress'])
 
+# Get shared_dir path and mount EFS filesystem
+efs_shared_dir = "/"+node['cfncluster']['cfn_efs_shared_dir'].split(',')[0]
```
enrico-usai: Same as before

recipes/_master_base_config.rb
```diff
@@ -28,6 +28,38 @@
...
+
+# Check to see if EFS is created
+if efs_shared_dir != "/NONE"
+    # Create the EFS shared directory
```
enrico-usai: same as before for indentation

recipes/_master_base_config.rb
```diff
@@ -28,6 +28,38 @@
...
+    device "#{node['cfncluster']['cfn_efs']}.efs.#{node['cfncluster']['cfn_region']}.amazonaws.com:/"
+    fstype 'nfs4'
+    options 'nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=30,retrans=2,noresvport,_netdev'
+    dump 0
+    pass 0
+    action %i[mount enable]
+  end
+
+  # Make sure EFS shared directory permissions are correct
+  directory efs_shared_dir do
```
enrico-usai: Do we really need it? It is the same code executed in the first step.

recipes/_master_base_config.rb
```diff
@@ -28,6 +28,38 @@
 # Get VPC CIDR
 node.default['cfncluster']['ec2-metadata']['vpc-ipv4-cidr-block'] = get_vpc_ipv4_cidr_block(node['macaddress'])
 
+# Get shared_dir path and mount EFS filesystem
```
enrico-usai: Most of the code is duplicated,  what about creating a new recipe, included by both `_compute_base_config.rb` and `_master_base_config.rb`?

recipes/_master_base_config.rb
```diff
@@ -28,6 +28,38 @@
...
+efs_shared_dir = "/" + node['cfncluster']['cfn_efs_shared_dir'].split(',')[0]
+
+# Check to see if EFS is created
+if efs_shared_dir != "/NONE"
+  # Create the EFS shared directory
+  directory efs_shared_dir do
+    owner 'root'
+    group 'root'
+    mode '1777'
+    recursive true
```
mauri-melato: why recursive on create?

recipes/_master_base_config.rb
mauri-melato: Isn't this needed on the compute node, or this operation performed on the master is enough?

recipes/_master_base_config.rb
mauri-melato: Agree, a new efs_mount reciped included both in the master and compute recipes.

recipes/_master_base_config.rb
rexcsn: so recursive will ensure all the root directory have the same owner group mode permissions. so if the user wants the efs directory to be "/dirs/efs" and recursive is true, both "/dirs" and "/dirs/efs" will have the owner group mode specified. If this is not a behavior that we want, we can take recursive out and that way only "/dirs/efs" will have the owner group mode specified.

Think about this now, I guess if we want to keep recursive in this block we should also set recursive to true in the directory block after the nfs mounting

recipes/_master_base_config.rb
rexcsn: This action is not needed on the compute for some reason. However since I am moving efs mounting to a separate recipe and including the recipe in the master and compute base config, we might as well just do this for the compute as well for extra assurance

PR #257, opened 2019-01-29
torque: wait before setting dummy node to offline

demartinofra:
Before setting the dummy node state to offline we need to wait for the updated okclients list
to be sent out to the cluster. Once this happens the state is set to free and that's when
we want to transition to offline.

### DIFF
--- a/templates/default/torque.setup.erb
+++ b/templates/default/torque.setup.erb
@@ -97,4 +97,18 @@ qmgr -c 'set server default_queue = batch'
 qmgr -c "set server query_other_jobs = True"
 
 qmgr -c 'create node <%= node['hostname'] %> np=1000'
+
+# Before setting the dummy node state to offline we need to wait for the updated okclients list
+# to be sent out to the cluster. Once this happens the state is set to free and that's when
+# we want to transition to offline.
+i=1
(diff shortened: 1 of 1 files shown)


templates/default/torque.setup.erb
```diff
@@ -97,4 +97,11 @@ qmgr -c 'set server default_queue = batch'
 qmgr -c "set server query_other_jobs = True"
 
 qmgr -c 'create node <%= node['hostname'] %> np=1000'
+
+# Before setting the dummy node state to offline we need to wait for the updated okclients list
+# to be sent out to the cluster. Once this happens the state is set to free and that's when
+# we want to transition to offline.
+while pbsnodes <%= node['hostname'] %> | grep "state = down,offline,MOM-list-not-sent" > /dev/null; do
```
mauri-melato: To make the test more robust I'd test specifically on the `MOM-list-not-sent` condition so to avoid a change in the order of the states would break the test, i.e. `while ... | grep "state = " | grep "MOM-list-not-sent" ...`.

Moreover, please consider the worst case where the status is not going to change, what happens, is there a timeout, e.g. CFN stack creation general timeout, to prevent never ending loop ?

demartinofra: Regarding the timeout: yes the initial idea was to rely on the cfn timeout. I set a shorter timeout of 2 minutes to fail faster in case of errors.

PR #259, opened 2019-02-11
Allow users to force deletion of jobs on SGE

lukeseawalker:
Change the default SGE configuration in order to allow regular users
to be able to force deletion of jobs

Note: Forced deletion for jobs is executed  differently depending on
whether users are Sun Grid Engine administrators or not. In case of
administrative  users,  the jobs are removed from the internal database
of Sun Grid Engine immediately. For regular users,  the  equivalent
of a normal qdel is executed first, and deletion is forced only if the
normal cancellation  was  unsuccessful.

Test
- after cluster creation
```
qconf -sconf | grep qmaster_params
qmaster_params               ENABLE_FORCED_QDEL
```


recipes/_master_sge_config.rb
```diff
@@ -28,6 +28,20 @@
   mode '0644'
 end
 
+# Copy qconf utils (Downloaded from http://arc.liv.ac.uk/SGE/downloads/qconf_scripts.tar.gz)
+cookbook_file 'qconf_scripts.tar.gz' do
+  path '/opt/sge/util/qconf_scripts.tar.gz'
```
enrico-usai: I'd like to define a variable for the `/opt/sge` and for the `/opt/sge/util` paths but it could part of another patch

recipes/_master_sge_config.rb
```diff
@@ -65,3 +79,10 @@
     rm $TMPFILE
   SETAS
 end
+
+bash "enable_forced_qdel" do
```
enrico-usai: I'd add a comment to explaining "why" we are doing this setting.

PR #264, opened 2019-02-25
Install Lustre on Centos 7.6 & 7.5

sean-smith:
Only install Lustre Modules on Centos 7.6 and 7.5.

If the centos version isn't supported, log it, but don't fail. This is because the `_lustre_install.rb` is run regardless of if FSx is used.

The `.cover?` function works like so:

```ruby
> (5..6).cover?(5)
True
> (5..6).cover?(6)
True
> (5..6).cover?(7)
False
```

#### Centos 7.6
Tested with `base_os = centos7`

```ini
[cluster fsx]
...
base_os = centos7
fsx_settings = fs

[fsx fs]
shared_dir = /fsx
fsx_fs_id = fs-073c3703dca3e28b6
```

#### Centos 7.5
I have not tested, as I couldn't find a working ami, but here's the strategy: set `custom_ami = ami-017b013ef58a362fb` (This is an old parallelcluster release)

```ini
[cluster fsx]
...
custom_ami = ami-017b013ef58a362fb
fsx_settings = fs

[fsx fs]
shared_dir = /fsx
fsx_fs_id = fs-073c3703dca3e28b6
```

### DIFF
--- a/recipes/_lustre_install.rb
+++ b/recipes/_lustre_install.rb
@@ -13,7 +13,8 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-if node['platform'] == 'centos'
+# Install only on Centos 7.6 and 7.5
+if node['platform'] == 'centos' && (5..6).cover?(node['platform_version'].split('.')[1].to_i)
   lustre_kmod_rpm = "#{node['cfncluster']['sources_dir']}/kmod-lustre-client-#{node['cfncluster']['lustre']['version']}.x86_64.rpm"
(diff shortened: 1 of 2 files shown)


recipes/_lustre_install.rb
```diff
@@ -44,4 +45,6 @@
   yum_package 'lustre_client' do
     source lustre_client_rpm
   end
+elsif node['platform'] == 'centos'
+  Chef::Log.error("Unsupported version of Centos, #{node['platform_version']}, supported versions are 7.6 and 7.5")
```
mauri-melato: warn message here since it's not an error for which the build should fail.

sean-smith: changed

PR #272, opened 2019-03-15
slurm: declare nodes in separate config file and use FUTURE for dummy nodes

demartinofra:
Slightly refactored slurm.conf in order to include a separate config files
where nodes are declared. This allows the sqswatcher to regenerate the nodes
file when it needs to add or remove an host rather than modifying based on
regexes the full slurm.conf file

Also using FUTURE state for dummy nodes so that Slurm does not try to contact
all dummy nodes at every reboot. For a cluster with max size equals 1000 this
was causing a reboot delay of about 40 seconds.

### DIFF
--- a/templates/default/slurm.conf.erb
+++ b/templates/default/slurm.conf.erb
@@ -102,15 +102,5 @@ JobCompType=jobcomp/none
 # when manually editing! Default node is there to allow
 # submission of jobs to an empty cluster.
 #PARTITION:compute
-<%
-def append_dummy
-    if node['cfncluster']['cfn_max_queue_size'].to_i > 1
-        "dummy-compute[1-" + node['cfncluster']['cfn_max_queue_size'] + "]"
-    else
-        "dummy-compute"
-    end
-end
-%>
-NodeName=<%= append_dummy %> Procs=2048 State=UNKNOWN
-#NodeName= Procs=1 State=UNKNOWN
-PartitionName=compute Nodes=<%= append_dummy %> Default=YES MaxTime=INFINITE State=UP
+include slurm_parallelcluster_nodes.conf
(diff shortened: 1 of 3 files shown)


templates/default/slurm.conf.erb
```diff
@@ -102,15 +102,5 @@ JobCompType=jobcomp/none
...
-        "dummy-compute"
-    end
-end
-%>
-NodeName=<%= append_dummy %> Procs=2048 State=UNKNOWN
-#NodeName= Procs=1 State=UNKNOWN
-PartitionName=compute Nodes=<%= append_dummy %> Default=YES MaxTime=INFINITE State=UP
\ No newline at end of file
+include slurm_parallelcluster_nodes.conf
+PartitionName=compute Nodes=ALL Default=YES MaxTime=INFINITE State=UP
```
enrico-usai: I really like `Nodes=ALL` 👍

templates/default/slurm_parallelcluster_nodes.conf.erb
```diff
@@ -0,0 +1,8 @@
+<%
+  def append_dummy
+    if node['cfncluster']['cfn_max_queue_size'].to_i > 0
+        "dummy-compute[1-" + node['cfncluster']['cfn_max_queue_size'] + "]"
+    end
+  end
+%>
+NodeName=<%= append_dummy %> CPUs=2048 State=FUTURE
```
enrico-usai: I see you changed from Procs to CPUs, is there any difference?
Do you think to set the right number of CPUs for the configured instance type in future?

demartinofra: No difference. From what I understood Procs is the legacy parameter, CPU the newer one.

Yes, I will address the CPU configuration separately.

PR #274, opened 2019-03-18
FSx Amazon Linux Support

sean-smith:
### Testing

1. Create an AMI with the latest Amazon Linux ami id `ami-0080e4c5bc078760e`
```bash
$ pcluster createami --ami-id ami-0080e4c5bc078760e --os alinux
```
2. Use the following config and the ami id from above to create a cluster
```ini
[cluster fsx]
custom_ami = [result_from_createami]
vpc_settings = public
key_name = [key_name]
fsx_settings = fs
custom_template = https://s3.amazonaws.com/seaam-pcluster/template/aws-parallelcluster.cfn.2.2.1.json # needed for custom_ami patches
custom_chef_cookbook = https://s3.amazonaws.com/seaam-pcluster/cookbooks/aws-parallelcluster-cookbook-2.2.1.tgz

[fsx fs]
shared_dir = /fsx
fsx_fs_id = fs-073c3703dca3e28b6
```

3. SSH in and check to see that the file system is mounted.
```bash
$ pcluster ssh mycluster
..
~> df -h
... 
/fsx   3.4TB   27GB   12G    10%     /fsx
```

### DIFF
--- a/recipes/_lustre_install.rb
+++ b/recipes/_lustre_install.rb
@@ -47,4 +47,12 @@ if node['platform'] == 'centos' && (5..6).cover?(node['platform_version'].split(
   end
 elsif node['platform'] == 'centos'
   Chef::Log.warn("Unsupported version of Centos, #{node['platform_version']}, supported versions are 7.6 and 7.5")
+elsif node['platform'] == 'amazon'
+
+  # Install lustre client module
+  package 'lustre-client' do
+    retries 3
+    retry_delay 5
+  end
+
 end


recipes/_lustre_install.rb
```diff
@@ -47,4 +47,11 @@
   end
 elsif node['platform'] == 'centos'
   Chef::Log.warn("Unsupported version of Centos, #{node['platform_version']}, supported versions are 7.6 and 7.5")
-end
+elsif node['platform'] == 'amazon'
+
+  # Install lustre client module
+  yum_package 'lustre-client' do
```
lukeseawalker: we should use "package" instead of the declined version for the OS, see https://docs.chef.io/resource_package.html and https://docs.chef.io/resource_yum_package.html
```
package lustre-client do
    retries 3
    retry_delay 5
  end
```

sean-smith: Good Idea 👍🏻

sean-smith: added

PR #279, opened 2019-03-21
Install EFA drivers

sean-smith:
Installs the EFA drivers. See [1]

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-enable


recipes/_efa_install.rb
```diff
@@ -0,0 +1,39 @@
...
+    mode '0644'
+    retries 3
+    retry_delay 5
+    not_if { ::File.exist?(efa_installer) }
+  end
+
+  bash "install efa" do
+    cwd node['cfncluster']['sources_dir']
+    code <<-NODE
+      yum remove -y openmpi
```
enrico-usai: Could you add a comment describing why we need to remove openmpi?

sean-smith: Added

recipes/_efa_install.rb
```diff
@@ -0,0 +1,41 @@
...
+    not_if { ::File.exist?(efa_installer) }
+  end
+
+  bash "install efa" do
+    cwd node['cfncluster']['sources_dir']
+    code <<-NODE
+      # default openmpi installation conflicts with new install
+      # new one is installed in /opt/amazon/efa/bin/
+      yum remove -y openmpi
+      tar -xzf aws-efa-installer-latest.tar.gz
```
lukeseawalker: use #{efa_installer} instead. Minor: rename the var efa_installer efa_tarball to be consistent with the other tarballs

sean-smith: done

recipes/base_install.rb
```diff
@@ -235,3 +235,6 @@
 
 # Install FSx options
 include_recipe "aws-parallelcluster::_lustre_install"
+
+# Install EFA
+include_recipe "aws-parallelcluster::_efa_install"
```
lukeseawalker: just for consistency I would move the OS check here like we do for the fsx recipe

attributes/default.rb
```diff
@@ -45,6 +45,8 @@
 default['cfncluster']['nvidia']['enabled'] = 'no'
 default['cfncluster']['nvidia']['driver_url'] = 'http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
 default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.nvidia.com/compute/cuda/10.0/Prod/local_installers/cuda_10.0.130_410.48_linux'
+# EFA
+default['cfncluster']['efa']['installer_url'] = 'https://s3-us-west-2.amazonaws.com/aws-efa-installer/aws-efa-installer-latest.tar.gz'
```
lukeseawalker: we need to understand the implication of having "us-west-2"

sean-smith: yep, china regions won't work we're working on resolving this

recipes/base_install.rb
sean-smith: I don't understand what you mean? FSx recipe checks the os in the `base_config.rb`, not in `base_install.rb`. We install efa drivers in `base_install.rb` so it'll be built into the ami, or installed at runtime for a custom ami.

lukeseawalker: yeah, I meant to move the check
`if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon'`
outside the recipe _efa_install, so it's not even called

`if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon'
  include_recipe "aws-parallelcluster::_efa_install"
end
`

PR #295, opened 2019-04-12
Pin supervisord to version 3.4.0

lukeseawalker:
This is the latest version with Python 2.6 support

Fix also the variable name, which is referenced in https://github.com/aws/aws-parallelcluster-cookbook/blob/9f25f6f44ff507bc873d349aa896e2c7c7e85f48/recipes/base_install.rb#L193


enrico-usai:
Could you please add in the description why are you changing the attribute name too?

PR #299, opened 2019-04-23
Avoid failure when partition changes require a reboot

lukeseawalker:
The patch will let the script continue also when the following error is
returned by the parted command:
"Error: Partition(s) Y on /dev/XXX have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use.  As a result, the old partition(s) will remain in use.  You should reboot now before making further changes."

This will avoid issue reported here https://github.com/aws/aws-parallelcluster/issues/990

### DIFF
--- a/files/default/setup-ephemeral-drives.sh
+++ b/files/default/setup-ephemeral-drives.sh
@@ -21,6 +21,13 @@ function error_exit () {
   exit 1
 }
 
+function exec_command() {
+  _command_output=$($@ 2>&1)
+  _exit_code=$?
+
+  # Do not set RC=1 if error says that changes have been written but a reboot is required to inform the kernel
+  [[ $_exit_code -ne 0 && $(echo "${_command_output}" | grep -i "you should reboot now") ]] && RC=1
+}
 
 # LVM stripe, format, mount ephemeral drives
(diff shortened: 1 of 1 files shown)


files/default/setup-ephemeral-drives.sh
```diff
@@ -21,6 +21,12 @@ function error_exit () {
   exit 1
 }
 
+function exec_command() {
+  _command_output=$($@ 2>&1)
+  _exit_code=$?
+
+  [[ $_exit_code -ne 0 && $(echo "${_command_output}" | grep -i "you should reboot now") ]] && RC=1
```
demartinofra: can we add a small comment on why we are doing this? especially why excluding this specific error? Probably you mentioned it in the PR but having it in the code might be useful for future reference.

files/default/setup-ephemeral-drives.sh
```diff
@@ -21,6 +21,12 @@ function error_exit () {
   exit 1
 }
 
+function exec_command() {
+  _command_output=$($@ 2>&1)
```
demartinofra: shall we double quote array expansions to avoid re-splitting elements?

PR #311, opened 2019-04-30
Skip Slurm installation on compute node

lukeseawalker:
Slurm installation folder is mounted from master node
The installation is done when "cfn_node_type" is "MasterServer" (at runtime)
or is "nil" (at packer time)

### DIFF
--- a/recipes/slurm_install.rb
+++ b/recipes/slurm_install.rb
@@ -16,43 +16,81 @@
 include_recipe 'aws-parallelcluster::base_install'
 include_recipe 'aws-parallelcluster::munge_install'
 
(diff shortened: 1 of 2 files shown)


recipes/slurm_install.rb
```diff
@@ -16,43 +16,64 @@
...
+      make -j $CORES
+      make install
+      make install-contrib
+    SLURM
+    # TODO: Fix, so it works for upgrade
+    creates '/opt/slurm/bin/srun'
+  end
+
+  # Setup slurm user
+  user "slurm" do
```
enrico-usai: The user must be added in the compute node too.

lukeseawalker: done

PR #316, opened 2019-05-15
Test for EFA Install

sean-smith:

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -99,6 +99,13 @@ if node['cfncluster']['cfn_scheduler'] == 'slurm'
   end
 end
 
+if node['cfncluster']['os'] == 'alinux' || node['cfncluster']['os'] == 'centos7'
+  execute 'check efa rpm installed' do
+    command "rpm -qa | grep libfabric && rpm -qa | grep efa-"
+    user node['cfncluster']['cfn_cluster_user']
+  end
+end
+
 unless node['cfncluster']['os'].end_with?("-custom")
   bash 'execute jq' do
     cwd Chef::Config[:file_cache_path]


recipes/tests.rb
```diff
@@ -99,6 +99,13 @@
   end
 end
 
+if node['cfncluster']['os'] == 'alinux' || node['cfncluster']['os'] == 'centos7'
+  execute 'check efa_install' do
```
lukeseawalker: execute 'ls efa root' do

recipes/tests.rb
```diff
@@ -99,6 +99,13 @@
...
+    command "ls /opt/amazon/efa || exit 1"
```
lukeseawalker: I think "exit 1" is redundant

recipes/tests.rb
```diff
@@ -99,6 +99,13 @@
...
+    user node['cfncluster']['cfn_cluster_user']
+  end
+end
+
```
lukeseawalker: What about the check on the driver (the one instance type dependent)?

recipes/tests.rb
sean-smith: done

recipes/tests.rb
sean-smith: yep, `ls` doesn't exit with a non-zero exit code if the directory doesn't exist, so I've changed it to use `-d` to check if the directory exists.

recipes/tests.rb
sean-smith: As we discussed last week, the driver needs to be enabled in the LaunchTemplate which we can't change for the kitchen tests. This'll be tested in the integration test.

recipes/tests.rb
sean-smith: I've changed this to check for the rpm being installed. `rpm -qa` lists installed rpms. See [1]

```bash
$ rpm -qa | grep libfabric
libfabric-devel-1.7.0amzn1.1-1.amzn2.x86_64
libfabric-1.7.0amzn1.1-1.amzn2.x86_64
$ rpm -qa | grep efa-
efa-0.9.2-1.amzn2.x86_64
```

[1] https://www.cyberciti.biz/howto/question/linux/linux-rpm-cheat-sheet.php

PR #317, opened 2019-05-23
Change download URL for CloudFormation Helper Scripts package

lukeseawalker:
This will add support for Ubuntu in China NorthWest region (cn-northwest-1)


sean-smith:
Is there any risk from pulling from `cn-north-1` just on amazon linux? From the code it looks like now, `alinux` is the odd one out. What if these are different versions?

lukeseawalker:
@sean-smith we aren't pulling anything on amazon-linux. Tools are already installed.

PR #323, opened 2019-05-30
Install EFA at Runtime

sean-smith:

### DIFF
--- a/recipes/_efa_fetch.rb
+++ b/recipes/_efa_fetch.rb
@@ -1,6 +1,6 @@
 #
 # Cookbook Name:: aws-parallelcluster
-# Recipe:: _efa_install
+# Recipe:: _efa_fetch
 #
 # Copyright 2013-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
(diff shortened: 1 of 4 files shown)


recipes/_efa_fetch.rb
```diff
@@ -0,0 +1,25 @@
+#
+# Cookbook Name:: aws-parallelcluster
+# Recipe:: _efa_install
```
lukeseawalker: fix name

recipes/base_install.rb
```diff
@@ -236,14 +236,5 @@
...
-    include_recipe "aws-parallelcluster::_efa_install"
-  else
-    package 'openmpi-devel' do
-      retries 3
-      retry_delay 5
-    end
-  end
-end
+# Fetch EFA Driver
+include_recipe "aws-parallelcluster::_efa_fetch"
```
lukeseawalker: Better keep the check for OS and China

recipes/base_config.rb
```diff
@@ -58,3 +58,8 @@
   # Mount FSx
   include_recipe 'aws-parallelcluster::fsx_mount'
 end
+
+# Install EFA
+if node['cfncluster']['enable_efa'] == 'compute' && node['cfncluster']['cfn_node_type'] == 'ComputeFleet'
+  include_recipe "aws-parallelcluster::_efa_install"
```
lukeseawalker: where are the checks for OS, China and instance type?

recipes/_efa_fetch.rb
sean-smith: fixed

recipes/base_install.rb
sean-smith: the driver should be downloaded on all os's just not installed on os's that don't support. Added china check back

recipes/base_config.rb
sean-smith: they're in the frontend

sean-smith: Ok, I've revised https://github.com/aws/aws-parallelcluster/pull/1081 so that it enforces the instance type check even when config_sanity check is disabled. This means we can rely on the enable_efa flag to signify that the instance_type and os are supported. Only check needed is to check `enable_efa  = compute` and  `cfn_node_type = ComputeFleet`.

recipes/base_config.rb
```diff
@@ -58,3 +58,8 @@
   # Mount FSx
   include_recipe 'aws-parallelcluster::fsx_mount'
 end
+
+# Install EFA
+if node['cfncluster']['enable_efa'] == 'compute' && node['cfncluster']['cfn_node_type'] == 'ComputeFleet'
```
lukeseawalker: what if node['cfncluster']['enable_efa'] is not set?

sean-smith: It's always set, see https://github.com/aws/aws-parallelcluster/pull/1081#discussion_r289290570

PR #327, opened 2019-05-31
Enable EFA at Runtime

sean-smith:
This builds the rpms into the ami, then set the limits at runtime.


recipes/_efa_enable.rb
```diff
@@ -13,13 +13,10 @@
...
-remote_file efa_tarball do
-  source node['cfncluster']['efa']['installer_url']
-  mode '0644'
-  retries 3
-  retry_delay 5
-  not_if { ::File.exist?(efa_tarball) }
+bash "enable efa limits" do
+  cwd Chef::Config[:file_cache_path]
+  code <<-EFAENABLE
+    cd aws-efa-installer/install
```
lukeseawalker: file_cache_path is a temp dir, so aws-efa-installer could be not present anymore, between the base_install and the base_config recipes

sean-smith: changed to sources dir: `/opt/parallelcluster/sources`

recipes/base_install.rb
```diff
@@ -236,7 +236,14 @@
...
 
-# Fetch EFA Driver
-unless node['cfncluster']['cfn_region'].start_with?("cn-")
-  include_recipe "aws-parallelcluster::_efa_fetch"
+# Install EFA
+if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon' || (node['platform'] == 'ubuntu' && node['platform_version'] == "16.04")
+  unless node['cfncluster']['cfn_region'].start_with?("cn-")
+    include_recipe "aws-parallelcluster::_efa_install"
+  else
+    package 'openmpi-devel' do
```
lukeseawalker: In Ubuntu there is no openmpi-devel, package name is different

sean-smith: ubuntu will be a separate patch, I've removed it for now

PR #330, opened 2019-06-05
Enable EFA on Ubuntu1604

sean-smith:
* Installs EFA on Ubuntu1604
* Default openmpi is specified in default/attributes.rb
  * no longer needs to be added in china regions

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -237,13 +237,15 @@ include_recipe "aws-parallelcluster::_nvidia_install"
 include_recipe "aws-parallelcluster::_lustre_install"
 
 # Install EFA
-if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon'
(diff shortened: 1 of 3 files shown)


recipes/base_install.rb
```diff
@@ -237,13 +237,14 @@
...
-if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon'
+if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon' || (node['platform'] == 'ubuntu' && node['platform_version'] == "16.04")
   unless node['cfncluster']['cfn_region'].start_with?("cn-")
     include_recipe "aws-parallelcluster::_efa_install"
-  else
-    package 'openmpi-devel' do
-      retries 3
-      retry_delay 5
-    end
+  end
```
lukeseawalker: this should be an "else":
if (not cn-) then execute efa_install
else execute mpi install

sean-smith: Fixed

PR #375, opened 2019-09-04
Support for Ubuntu 18.04

ddeidda:
Signed-off-by: ddeidda <ddeidda@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -186,7 +186,9 @@ include_recipe "aws-parallelcluster::_nvidia_install"
 include_recipe "aws-parallelcluster::_lustre_install"
 
 # Install EFA & Intel MPI
-if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon' || (node['platform'] == 'ubuntu' && node['platform_version'] == "16.04")
+if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) \
+  || node['platform'] == 'amazon' \
+  || (node['platform'] == 'ubuntu' && node['platform_version'].to_f > 16.04)
   unless node['cfncluster']['cfn_region'].start_with?("cn-")
     include_recipe "aws-parallelcluster::_efa_install"
     include_recipe "aws-parallelcluster::intel_mpi"
--- a/recipes/sge_install.rb
+++ b/recipes/sge_install.rb
@@ -29,6 +29,14 @@ when 'MasterServer', nil
     not_if { ::File.exist?(sge_tarball) }
   end
 
+  # Patch file for SGE sources
+  cookbook_file 'sge_source.patch' do
+    path '/tmp/sge_source.patch'
+    user 'root'
+    group 'root'
+    mode '0644'
+  end
+
   # Install SGE
   bash 'make install' do
     user 'root'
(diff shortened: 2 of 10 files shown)


recipes/base_install.rb
```diff
@@ -186,7 +186,9 @@
 include_recipe "aws-parallelcluster::_lustre_install"
 
 # Install EFA & Intel MPI
-if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon' || (node['platform'] == 'ubuntu' && node['platform_version'] == "16.04")
+if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7) || node['platform'] == 'amazon' \
+  || (node['platform'] == 'ubuntu' && node['platform_version'] == "16.04") \
+  || (node['platform'] == 'ubuntu' && node['platform_version'] == "18.04")
```
lukeseawalker: minor, is it possible to put the check `node['platform'] == 'ubuntu' ` in common?
or use something like we use for centos `>= 7`

recipes/sge_install.rb
```diff
@@ -29,20 +29,30 @@
...
     user 'root'
     group 'root'
     cwd Chef::Config[:file_cache_path]
     environment 'SGE_ROOT' => '/opt/sge'
     code <<-SGE
+      set -e
       tar xf #{sge_tarball}
-      cd sge-#{node['cfncluster']['sge']['version']}/source
+      cd sge-#{node['cfncluster']['sge']['version']}
+      patch -p1 < /tmp/sge_source.patch
```
lukeseawalker: I would like to apply the patch only for ubuntu18, same logic we do for torque. Let discuss about it

recipes/base_install.rb
ddeidda: Something like node['platform_version'] >= 16.04 you mean? Yes I like it

recipes/sge_install.rb
ddeidda: The idea is to get rid of "ifs" by leaving a unique flow and managing the different behaviors, if needed, in the specific file. By default the patch is empy, it contains data only for ubuntu 18

ddeidda: In this way, if a new OS will require a different patch we could just add the new file without having to change the main code

lukeseawalker: I don't like this approach for a couple of reasons:
* the patch for ubuntu18 is applied also for the other OSs which do not require it
* if a new OS will require a different patch, as you said, it could be possible that is not compatible with ubuntu18 or the other OSs
* in the long run I would like it solved upstream, so no patch would be required

ddeidda: - As I said, to the other OSs is applied an empty patch (the default one)
- If a new OS will require a different patch we will just need to put this different patch in the file files/<new OS>/sge_source.patch

Maybe it's helpful to remind that the patch is managed as a cookbook file, so each OS will have it's own patch (and just that one)

lukeseawalker: ok good, I didn't see the patch is located in files/ubuntu-18.04/sge_source.patch

recipes/base_install.rb
ddeidda: done

PR #376, opened 2019-09-04
Add installation package retry where missing

lukeseawalker:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/recipes/awsbatch_config.rb
+++ b/recipes/awsbatch_config.rb
@@ -57,5 +57,7 @@ elsif node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7
 else
   python_package "aws-parallelcluster" do
     version node['cfncluster']['cfncluster-version']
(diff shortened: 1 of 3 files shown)


recipes/awsbatch_config.rb
```diff
@@ -57,5 +57,7 @@
 else
   python_package "aws-parallelcluster" do
     version node['cfncluster']['cfncluster-version']
+    retries 3
```
demartinofra: are we sure this is supported for python_package?

lukeseawalker: It [inherits](https://github.com/poise/poise-python/blob/master/lib/poise_python/resources/python_package.rb#L102) from chef package, which has the retry.

Btw I also tested it manually with this snippet
```
python_package "aws-parallelcluster" do
    retrie 3
    retry_delay 5
end
```
and it worked

PR #378, opened 2019-09-05
Export /opt/intel

sean-smith:

### DIFF
--- a/recipes/_compute_base_config.rb
+++ b/recipes/_compute_base_config.rb
@@ -47,6 +47,15 @@ mount '/home' do
   action %i[mount enable]
 end
 
+# Mount /opt/intel over NFS
+mount '/opt/intel' do
+  device "#{nfs_master}:/opt/intel"
+  fstype 'nfs'
+  options 'hard,intr,noatime,vers=3,_netdev'
+  action %i[mount enable]
+  only_if { ::File.directory?("/opt/intel") }
+end
+
 # Configure Ganglia
 if node['cfncluster']['ganglia_enabled'] == "yes"
   template '/etc/ganglia/gmond.conf' do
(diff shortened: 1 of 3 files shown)


recipes/intel_install.rb
```diff
@@ -71,24 +71,6 @@
   modulefile "2019.4"
 end
 
-# intel performance primitives
-create_modulefile "#{node['cfncluster']['modulefile_dir']}/intelipp" do
```
demartinofra: was this supposed to be part of this PR/commit?

recipes/_compute_base_config.rb
```diff
@@ -47,6 +47,14 @@
   action %i[mount enable]
 end
 
+# Mount /opt/intel over NFS
+mount '/opt/intel' do
+  device "#{nfs_master}:/opt/intel"
+  fstype 'nfs'
+  options 'hard,intr,noatime,vers=4,_netdev'
```
demartinofra: so far we've been using nfs version 3 everywhere. I agree we could update if there are no drawbacks but I would rather do it for all the mount points (I recall there were some performance degradations with version 4 we want to double check)

recipes/_master_base_config.rb
```diff
@@ -112,6 +112,13 @@
   options ['no_root_squash']
 end
 
+# Export /opt/intel
+nfs_export "/opt/intel" do
```
demartinofra: are we exporting and mounting this on all OSes?

recipes/intel_install.rb
sean-smith: yes, this removes un-needed module files. These components are all loaded in the psxe module file.

recipes/_master_base_config.rb
sean-smith: Changed to only exporting if the `/opt/intel` directory exists, this is the case on

* alinux, ubuntu1604, centos7
* not in china

recipes/_compute_base_config.rb
sean-smith: Can you send me a link to that performance issue? I don't have any context for that.

PR #349, opened 2019-07-11
Create recipe for dcv on centos7

matteofiordarancio:
Signed-off-by: Matteo Fiordarancio <mfiordar@amazon.com>

This pull request adds dcv compatibility for centos7. Still WIP

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #386, opened 2019-09-17
Add recipes to install/configure NICE DCV on CentOS7

enrico-usai:
This patch adds a new set or recipes and utility files required to integrate NICE DCV with ParallelCluster.

We are preparing the environment to enable the user to connect through NICE DCV to the master instance.

NICE DCV will be installed at AMI creation time and configured/enabled at runtime according to the user's choiches.

We added `pcluster_dcv_connect.sh` script to manage the DCV session/creation for the ec2-user and a new External Authenticator for DCV (`pcluster_dcv_authenticator.py`).

The external authenticator is a python WebServer, managed as a service by supervisord, that will run in a private virtual environment.

DCV will be automatically configured to use the ext authenticator.

___

This patch replaces the https://github.com/aws/aws-parallelcluster-cookbook/pull/349 and adds the following changes:
+ Rebased the code with the latest changes
+ Reworded the commit messages 
+ Add support for GPU acceleration on NICE DCV
+ Improve security of hash creation
With this patch I'm differentiating `RequestTokenValues` and `SessionTokenValues`
named tuples, to have the access_file hash stored in memory with the token itself
The advantage is we can use avoid to define the "salt" as global var and use
different salt for each hash creation.
Before of this page the same SALT was used for the whole `DCVAuthenticator` lifecycle.
+ Add comments, refactor code and add unit tests
+ Renamed functions, variables and added newlines to split code blocks
+ Renamed `requiredFile` parameter to `accessFile`
+ Renamed `pyenv_install` resource to `install_pyenv` to avoid confusion
  with `pyenv_*` resources included in the pyenv cookbook recipes:
  https://supermarket.chef.io/cookbooks/pyenv
+ Add unit tests and add comments to them
+ Refactored `install_rpm_packages_from_path` function
+ Use definitive `dna.json` to retrieve base os and not the temporary one
+ Install Xorg at AMI creation time and improve X running verification
+ Improve certificate and private key creation by creating different files
+ Explicitly define private key size RSA signature alghoritm
They are already the defaults: 2048-bit private key size, SHA-256 with RSA signature algorithm
+ Add logging for pcluster_dcv_connect bash script
+ rename `pcluster_dcv_ext_auth` log file to be aligned with py module name
+ Fix code to cleanup old authorization files
The code was deleting the files and the AUTHORIZATION_DIR folder too so
the authentication flow didn't work after server restart.
+ Remove init method from test class
Pytest is not able to collect method from a test Class with init method
+ Define `_atomic_touch` function in `pcluster_dcv_connect.sh`
+ Use a simple bash log rotation
The log will rotate if the size is greater than 5MB.
A backup of the log will be saved.
+ Remove future and retry external dependency and use py logging library
+ Add log messages to understand who is doing what (and when)
+ Override `server.log_message` to not print `requestToken` and `sectionToken` GET details.
+ Rename `ext_auth_files` folder to dcv
+ Add missing init file for dcv Python folder
+ Improve naming of DCV related stuff
    + `pcluster_dcv_ext_auth` -> `pcluster_dcv_authenticator`
    + `/var/spool/dcv_ext_auth` -> `/var/spool/parallelcluster/pcluster_dcv_authenticator`
    + `/var/log/parallelcluster/pcluster_dcv_ext_auth.log` -> `/var/log/parallelcluster/pcluster_dcv_authenticator.log`
    + `['dcv']['ext_auth_*']` -> `['dcv']['authenticator']['*']`
+ Improve SSH configuration
    + Disable X11Forwarding
    + Limit Ciphers and MACs
+ Create ParallelCluster log folder in _prep_env recipe
+ Update NICE DCV version from 2019.1-7423 to 2019.1-7644
+ Move installation of pcluster_dcv_connect.sh into the dcv_install recipe
+ Disable lock screen by default for all the users
It is required since default EC2 users don't have a password
References: https://developer.gnome.org/gio//2.48/glib-compile-schemas.html

### DIFF
--- a/amis/packer_centos7.json
+++ b/amis/packer_centos7.json
@@ -10,6 +10,7 @@
     "build_date" : "{{env `BUILD_DATE`}}",
     "vendor_path" : "{{env `VENDOR_PATH`}}",
     "nvidia_enabled" : "{{env `NVIDIA_ENABLED`}}",
+    "dcv_installed": "{{env `DCV_INSTALLED`}}",
     "instance_type" : "{{env `AWS_FLAVOR_ID`}}",
     "subnet_id" : "{{env `AWS_SUBNET_ID`}}",
     "security_group_id" : "{{env `AWS_SECURITY_GROUP_ID`}}",
@@ -230,6 +231,9 @@
           "cfn_region": "{{user `region`}}",
           "nvidia" : {
             "enabled" : "{{user `nvidia_enabled`}}"
+          },
+          "dcv": {
+            "installed" : "{{user `dcv_installed`}}"
           },
           "custom_node_package" : "{{user `custom_node_package`}}"
         }
--- a/files/default/dcv/generate_certificate.sh
+++ b/files/default/dcv/generate_certificate.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+
+# Generate a 5 year self-signed certificate and a new RSA private key and copy in the given path.
+# The command must be executed as root.
+
+# Usage:
+# ./generate_certificate.sh \
+#     "/etc/parallelcluster/ext-auth-certificate.pem" \
+#     "/etc/parallelcluster/ext-auth-private-key.pem" \
+#     dcv-ext-auth-user \
+#     dcv-ext-auth-user-group
+
+_check_param() {
+    paramValue="$1"
+    errorMessage="$2"
+    if [[ -z "${paramValue}" ]]; then
+        >&2 echo "${errorMessage}"
+        exit 1
+    fi
+}
+
+main() {
+    certificatePath="$1"
+    privateKeyPath="$2"
+    user="$3"
+    group="$4"
+
+    _check_param "${certificatePath}" "Certificate file path required"
+    _check_param "${privateKeyPath}" "Private Key file path required"
+    _check_param "${user}" "User required"
+    _check_param "${group}" "Group required"
+
+    # Generate a new certificate and a new RSA private key
+    openssl req -newkey rsa:2048 -sha256 -x509 -days 1825 -subj "/CN=localhost" -out "${certificatePath}" -nodes -keyout "${privateKeyPath}"
+    chmod 440 "${certificatePath}" "${privateKeyPath}"
+    chown "${user}":"${group}" "${certificatePath}" "${privateKeyPath}"
+}
+
+main "$@"
--- a/files/default/dcv/pcluster_dcv_connect.sh
+++ b/files/default/dcv/pcluster_dcv_connect.sh
@@ -1,0 +1,172 @@
+#!/usr/bin/env bash
+#
+# Cookbook Name:: aws-parallelcluster
+#
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script must be executed by the user who want to connect to DCV to obtain a Session Token.
+#
+# It performs the following steps:
+# 1. Asks the ParallelCluster DCV External Authenticator for a Request Token and the Access File name
+# 2. Creates the "access file" in the AUTHORIZATION_FILE_DIR folder (to confirm user identity)
+# 3. Asks the ParallelCluster DCV External Authenticator for a SessionToken (the real token to access to the DCV session)
+# 4. Returns DCV Session Id, DCV Server Port and the Session Token
+
+# Requirements:
+# jq, curl, awk, xargs, shuf, cat, grep, systemctl
+
+# Usage:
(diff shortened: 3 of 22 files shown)


amis/packer_centos7.json
```diff
@@ -10,6 +10,7 @@
     "build_date" : "{{env `BUILD_DATE`}}",
     "vendor_path" : "{{env `VENDOR_PATH`}}",
     "nvidia_enabled" : "{{env `NVIDIA_ENABLED`}}",
+    "dcv_installed": "{{env `DCV_INSTALLED`}}",
```
lukeseawalker: please add this var into the "guide" of the `build_ami.sh` script

metadata.rb
```diff
@@ -20,13 +20,13 @@
 depends 'poise-python', '~> 1.7.0'
 depends 'tar', '~> 2.1.1'
 depends 'selinux', '~> 2.1.0'
-depends 'nfs', '~> 2.6.3'
+depends 'nfs', '~> 2.5.1'
```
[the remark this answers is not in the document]
enrico-usai: yes, it is a temporary fix to make the branch work

metadata.rb
```diff
@@ -20,13 +20,13 @@
...
 depends 'yum', '~> 5.1.0'
 depends 'yum-epel', '~> 3.1.0'
 depends 'openssh', '~> 2.6.3'
 depends 'apt', '~> 7.0.0'
 depends 'hostname', '~> 0.4.2'
-depends 'line', '~> 2.4.1'
+depends 'line', '~> 1.0.6'
```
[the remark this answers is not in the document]
enrico-usai: yes, it is a temporary fix to make the branch work

libraries/helpers.rb
```diff
@@ -65,6 +65,14 @@ def get_vpc_ipv4_cidr_block(eth0_mac)
   vpc_ipv4_cidr_block
 end
 
+def get_instance_type()
+  uri = URI("http://169.254.169.254/latest/meta-data/instance-type")
+  res = Net::HTTP.get_response(uri)
+  master_instance_type = res.body if res.code == '200'
```
lukeseawalker: generalize `master_instance_type`

recipes/base_install.rb
```diff
@@ -204,3 +201,9 @@
 if (node['platform'] == 'centos' && node['platform_version'].to_i >= 7)
   include_recipe "aws-parallelcluster::intel_install"
 end
+
+# Disable selinux
+selinux_state "SELinux Disabled" do
+  action :disabled
+  only_if 'which getenforce'
+end
```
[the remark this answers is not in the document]
enrico-usai: It was already here, I just moved after DCV installation.

recipes/dcv_install.rb
```diff
@@ -0,0 +1,129 @@
...
+    # Extract DCV packages
+    unless File.exist?(dcv_tarball)
+      remote_file dcv_tarball do
+        source node['cfncluster']['dcv']['url']
+        mode '0644'
+        retries 3
+        retry_delay 5
+      end
+
+      bash 'extract dcv packages' do
```
lukeseawalker: use set -e if we need to catch failures

recipes/_setup_python.rb
```diff
@@ -14,40 +14,21 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
-
-pyenv_user_install 'root'
-
-pyenv_python node['cfncluster']['python-version'] do
-  user 'root'
+install_pyenv "root" do
```
lukeseawalker: this is not idempotent and will be re-executed at runtime

recipes/dcv_install.rb
```diff
@@ -0,0 +1,129 @@
...
+  # Override default GSettings to disable lock screen for all the users
+  cookbook_file "/usr/share/glib-2.0/schemas/10_org.gnome.desktop.screensaver.gschema.override" do
+    source 'dcv/10_org.gnome.desktop.screensaver.gschema.override'
+    owner 'root'
+    group 'root'
+    mode '0755'
+  end
+
+  # compile gsettings schemas
+  execute 'Compile gsettings schema' do
```
lukeseawalker: minor: this is not idempotent and will be re-executed at runtime

libraries/helpers.rb
enrico-usai: done

files/default/dcv/generate_certificate.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Usage:
+# ./generate_certificate.sh \
+#     "/etc/parallelcluster/ext-auth-certificate.pem" \
+#     "/etc/parallelcluster/ext-auth-private-key.pem" \
+#     dcv-ext-auth-user \
+#     dcv-ext-auth-user-group
+
+_check_set() {
+    file="$1"
```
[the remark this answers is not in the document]
enrico-usai: fixed

recipes/dcv_config.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  # dcvgl package must be installed after NVIDIA and before starting up X
+  dcv_gl = "#{Chef::Config[:file_cache_path]}/nice-dcv-#{node['cfncluster']['dcv']['version']}-el7/#{node['cfncluster']['dcv']['gl']}"
+  package dcv_gl do
+    action :install
+    source dcv_gl
+  end
+
+  # Configure the X server to start automatically when the Linux server boots and start the X server in background
+  bash 'Launch X' do
```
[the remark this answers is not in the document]
enrico-usai: added, thank you

recipes/dcv_install.rb
enrico-usai: added

amis/packer_centos7.json
enrico-usai: added

files/default/dcv/pcluster_dcv_connect.sh
```diff
@@ -0,0 +1,172 @@
...
+
+    log_time=$(date "+%Y-%m-%d %H:%M:%S")
+
+    file_size=$(du -b ${LOG_FILE_PATH} | awk '{print $1}')
+
+    [[ -f "${LOG_FILE_PATH}" ]] || _atomic_touch "${LOG_FILE_PATH}"
+
+    if [[ ${file_size} -gt ${LOG_FILE_MAX_SIZE} ]]; then
+        mv "${LOG_FILE_PATH}" "${LOG_FILE_PATH}.1"
+        echo "[${log_time}]: ${text}" > "${LOG_FILE_PATH}"
```
lukeseawalker: once you move the file, you can do the append >> in place of the >. 
This will factor out with the line in the else case, hence the else could be removed

PR #392, opened 2019-09-19
Regionalize Nvidia download URL

lukeseawalker:
curl -L http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run -o nvidia-commercial
curl -L http://cn.download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run -o nvidia-china

md5sum nvidia-*
d1cf6dc33af48b53edbae5d6f31a6f89  nvidia-commercial
d1cf6dc33af48b53edbae5d6f31a6f89  nvidia-china

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -57,6 +57,9 @@ default['cfncluster']['ganglia']['web_url'] = 'https://github.com/ganglia/gangli
 # NVIDIA
 default['cfncluster']['nvidia']['enabled'] = 'no'
 default['cfncluster']['nvidia']['driver_url'] = 'http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
+if node['cfncluster']['cfn_region'].start_with?("cn-")
+  default['cfncluster']['nvidia']['driver_url'] = 'http://cn.download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
+end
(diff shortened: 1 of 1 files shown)


attributes/default.rb
```diff
@@ -58,6 +58,9 @@
 default['cfncluster']['nvidia']['enabled'] = 'no'
 default['cfncluster']['nvidia']['driver_url'] = 'http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
 default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.nvidia.com/compute/cuda/10.0/Prod/local_installers/cuda_10.0.130_410.48_linux'
```
enrico-usai: Could we do the same for cuda driver?

attributes/default.rb
```diff
@@ -58,6 +58,9 @@
...
+if node['cfncluster']['cfn_region'].start_with?("cn-")
```
enrico-usai: We could move the "if" closer to the driver_url initialization or maybe use an if/else block

attributes/default.rb
lukeseawalker: Tried but that same version is not present in the mirror

enrico-usai: ok, thank you!

PR #400, opened 2019-09-30
Enable flock and user_xattr Lustre options by default

sean-smith:
From the [lustre docs](http://wiki.lustre.org/Mounting_a_Lustre_File_System_on_Client_Nodes):

> `flock`: This has no measurable performance impact in modern (2.x) versions of Lustre (this page previously stated there was a performance impact; this is not correct for modern versions). This is the recommended mode, and is enabled by default in Lustre 2.13 and newer.

> `user_xattr`: Enable support for user extended attributes. Extended attributes are name:value pairs associated permanently with files and directories, similar to the environment strings associated with a process.

> `noatime` This option turns off inode access time updates.

For `systemd` based systems I've added two more options:

> `x-systemd.automount` adds to `fstab` so filesystem is automatically mounted after reboots.

> `x-systemd.requires=lnet.service` requires lustre kernel module

### DIFF
--- a/recipes/fsx_mount.rb
+++ b/recipes/fsx_mount.rb
@@ -29,13 +29,19 @@ if fsx_shared_dir != "NONE"
     action :create
   end
 
+  mount_options = %w[defaults _netdev flock user_xattr noatime]
+
+  if node['init_package'] == 'systemd'
+    mount_options.push(%w[noauto x-systemd.automount x-systemd.requires=lnet.service])
+  end
+
   # Mount FSx over NFS
   mount fsx_shared_dir do
     device "#{node['cfncluster']['cfn_fsx_fs_id']}.fsx.#{node['cfncluster']['cfn_region']}.amazonaws.com@tcp:/fsx"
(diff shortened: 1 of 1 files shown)


sean-smith:
@lukeseawalker that's Centos 7, Ubuntu 16/18, and Amazon Linux 2

Centos 6 and Amazon Linux 1 are non-systemd.

recipes/fsx_mount.rb
```diff
@@ -29,13 +29,19 @@
     action :create
   end
 
+  mount_options = %w[defaults _netdev flock user_xattr noauto]
```
lukeseawalker: I would follow the doc moving the noauto into the systemd case

sean-smith:
@lukeseawalker The `noatime` is in the FSx docs: https://docs.aws.amazon.com/fsx/latest/LustreGuide/mounting-ec2-instance.html

lukeseawalker:
@sean-smith ` noauto != noatime`

PR #403, opened 2019-10-08
Upgrade NVIDIA Driver, CUDA. Added kitchen tests.

rexcsn:
- Updated download url for NVIDIA driver and CUDA
- Added kitchen test to test for NVIDIA driver and CUDA install. Note: need to change version in kitchen tests when we upgrade. 

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -57,8 +57,9 @@ default['cfncluster']['ganglia']['web_version'] = '3.7.2'
 default['cfncluster']['ganglia']['web_url'] = 'https://github.com/ganglia/ganglia-web/archive/3.7.2.tar.gz'
 # NVIDIA
 default['cfncluster']['nvidia']['enabled'] = 'no'
-default['cfncluster']['nvidia']['driver_url'] = 'http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
-default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.nvidia.com/compute/cuda/10.0/Prod/local_installers/cuda_10.0.130_410.48_linux'
+# domain has dynamic DNS resolution, will resolve to a server in Tokyo when called from China
+default['cfncluster']['nvidia']['driver_url'] = 'https://us.download.nvidia.com/tesla/418.87/NVIDIA-Linux-x86_64-418.87.00.run'
+default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda_10.1.243_418.87.00_linux.run'
 # EFA
 default['cfncluster']['efa']['installer_url'] = 'https://s3-us-west-2.amazonaws.com/aws-efa-installer/aws-efa-installer-1.5.4.tar.gz'
 # ENV2 - tool to capture environment and create modulefiles
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -138,3 +138,69 @@ unless node['cfncluster']['os'].end_with?("-custom")
     JQMERGE
   end
 end
+
+# only test on GPU instances
+if node['cfncluster']['nvidia']['enabled'] == 'yes'
+  # extract driver and CUDA version from url
+  url_driver_ver = node['cfncluster']['nvidia']['driver_url'][/(\d+.\d+)/, 0]
+  url_cuda_ver = node['cfncluster']['nvidia']['cuda_url'][/(\d+.\d+)/, 0]
+
+  bash 'test nvidia driver install' do
+    cwd Chef::Config[:file_cache_path]
+    code <<-TESTNVIDIA
+      has_gpu=$(lspci | grep -o "NVIDIA")
+      if [ -z "$has_gpu" ]; then
+        echo "No GPU detected, no test needed."
+        exit 0
+      fi
+
+      set -e
+      driver_ver="#{url_driver_ver}"
+      export PATH="/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/aws/bin"
+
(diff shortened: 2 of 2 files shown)


attributes/default.rb
```diff
@@ -57,8 +57,9 @@
 default['cfncluster']['ganglia']['web_url'] = 'https://github.com/ganglia/ganglia-web/archive/3.7.2.tar.gz'
 # NVIDIA
 default['cfncluster']['nvidia']['enabled'] = 'no'
-default['cfncluster']['nvidia']['driver_url'] = 'http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
-default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.nvidia.com/compute/cuda/10.0/Prod/local_installers/cuda_10.0.130_410.48_linux'
+# Is there a non-US domain?
```
sean-smith: remove

recipes/tests.rb
```diff
@@ -138,3 +138,68 @@
...
+    cwd Chef::Config[:file_cache_path]
+    code <<-TESTNVIDIA
+      has_gpu=$(lspci | grep -o "NVIDIA")
+      if [ -z $has_gpu ]; then
+        echo "No GPU detected, no test needed."
+        exit 0
+      fi
+
+      set -e
+      driver_ver="418.87"     #Set this to NVIDIA driver version that we are currently installing
```
sean-smith: move this to `attributes/default.rb`

recipes/tests.rb
```diff
@@ -138,3 +138,68 @@
...
+        exit 1
+      else
+        echo "Correctly installed NVIDIA $driver_output"
+      fi
+    TESTNVIDIA
+  end
+end
+
+# only test on GPU instances
+if node['cfncluster']['nvidia']['enabled'] == 'yes'
```
sean-smith: Can these blocks be combined?

recipes/tests.rb
```diff
@@ -138,3 +138,68 @@
...
+    cwd Chef::Config[:file_cache_path]
+    code <<-TESTCUDA
+      has_gpu=$(lspci | grep -o "NVIDIA")
+      if [ -z $has_gpu ]; then
+        echo "No GPU detected, no test needed."
+        exit 0
+      fi
+
+      set -e
+      cuda_ver="10.1"         #Set this to CUDA version that we are currently installing
```
sean-smith: same here

attributes/default.rb
demartinofra: actually is there a non us domain? :D

recipes/tests.rb
```diff
@@ -138,3 +138,68 @@
     JQMERGE
   end
 end
+
+# only test on GPU instances
+if node['cfncluster']['nvidia']['enabled'] == 'yes'
+  bash 'test nvidia driver install' do
+    cwd Chef::Config[:file_cache_path]
+    code <<-TESTNVIDIA
```
demartinofra: I think it might be a good idea to have `set -e` in tests so that we fail if any command fails

rexcsn: `set -e` on line 153 and line 182. If we do `set -e` before the line 147, in the case of the instance not having GPU hardware and we are not expected to run this test, the program will `exit 1` after line 147, which is something we don't want. Hence `set -e` only after the check for GPU hardware

recipes/tests.rb
rexcsn: fdm wanted them separate, so we can see which block(driver or CUDA) is failing in kitchen test

attributes/default.rb
rexcsn: I couldn't find it :(

lukeseawalker: for cn
* https://cn.download.nvidia.com/tesla/418.87/NVIDIA-Linux-x86_64-418.87.00.run
* https://developer.download.nvidia.cn/compute/cuda/10.1/Prod/local_installers/cuda_10.1.243_418.87.00_linux.run

recipes/tests.rb
lukeseawalker: you can put the two "bash" resources in the same "if node['cfncluster']['nvidia']['enabled'] == 'yes'"

attributes/default.rb
```diff
@@ -57,8 +57,11 @@
 default['cfncluster']['ganglia']['web_url'] = 'https://github.com/ganglia/ganglia-web/archive/3.7.2.tar.gz'
 # NVIDIA
 default['cfncluster']['nvidia']['enabled'] = 'no'
-default['cfncluster']['nvidia']['driver_url'] = 'http://download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run'
-default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.nvidia.com/compute/cuda/10.0/Prod/local_installers/cuda_10.0.130_410.48_linux'
+default['cfncluster']['nvidia']['driver_version'] = '418.87'
```
lukeseawalker: it would be nice to use this variable as part of the download URL, so we could change the version only in one place and don't forget to keep it aligned

attributes/default.rb
```diff
@@ -57,8 +57,11 @@
...
+# domain has dynamic DNS resolution, will resolve to a server in Tokyo when called from China
+default['cfncluster']['nvidia']['driver_url'] = 'https://us.download.nvidia.com/tesla/418.87/NVIDIA-Linux-x86_64-418.87.00.run'
+default['cfncluster']['nvidia']['cuda_version'] = '10.1'
+default['cfncluster']['nvidia']['cuda_url'] = 'https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda_10.1.243_418.87.00_linux.run'
```
lukeseawalker: or another possibility is to remove completely the definition of driver_version and cuda_version and extract them from the download URL, in order to use them in the "bash" tests

PR #405, opened 2019-10-17
Upgrade slurm version to 19.05.3-2

rexcsn:
* Change slurm verison and url in attributes/default.rb
* Remove deprecated option FastSchedule from slurm.conf


templates/default/slurm.conf.erb
```diff
@@ -69,10 +69,11 @@ SchedulerType=sched/backfill
 SelectType=select/cons_res
 SelectTypeParameters=CR_CPU
 
+# FastSchedule is deprecated in release 19.05.3-x
```
demartinofra: let's just remove the part about FastSchedule and do that in a separate commit that contains this explanation

PR #409, opened 2019-10-18
Enable Slurm GPU options

rexcsn:
* Added run_nvidiasmi to _compute_slurm_config if instance has GPU so that GPU devices are installed
* Added gres.conf.erb as base gres.conf file
* Modified slurm.conf.erb, changed SelectType plugin to cons_tres and added GresTypes=gpu.
* Note that this configuration will be used for all instance types. The modification will not affect non-GPU instance because resources on a node is used if specified by slurm_parallelcluster_nodes.conf and gres.conf, which are written by the sqswatcher. The sqswatcher has the logic to make sure the right resources are discovered by the scheduler, so we do not have to make the differentiation in cookbook package


templates/default/gres.conf.erb
```diff
@@ -0,0 +1 @@
+NodeName=dummy-compute Name=gpu Type=tesla File=/dev/nvidia0
```
sean-smith: Is this how we want the default node to appear? How does it effect sinfo?

rexcsn: If instance does not have GPU, when we update `slurm_parallelcluster_nodes.conf` in `sqswatcher` we will not include GPU information for a node. As long as GPU information is not mentioned in `slurm_parallelcluster_nodes.conf`, slurm will not look at `gres.conf`. So having this default node here will not affect existing behavior.

recipes/_compute_slurm_config.rb
```diff
@@ -22,6 +22,14 @@
   action %i[mount enable]
 end
 
+# Check to see if there is GPU on the instance, only execute run_nvidiasmi if there is GPU
+has_gpu = `lspci | grep -i -o 'NVIDIA'`
```
demartinofra: what is this for?

templates/default/slurm.conf.erb
```diff
@@ -66,7 +66,8 @@ SchedulerType=sched/backfill
 #SchedulerAuth=
 #SchedulerPort=
 #SchedulerRootFilter=
-SelectType=select/cons_res
+GresTypes=gpu
```
demartinofra: Let's start organizing the config into sections, can we group the GPU related options under something like 
```
### TRES and GPU config options
```

templates/default/gres.conf.erb
```diff
@@ -0,0 +1,3 @@
+# WARNING!!! The GRES entries below is managed by the sqswatcher process
```
demartinofra: more than the GRES entries is the `slurm_parallelcluster_gres.conf` file.

templates/default/slurm_parallelcluster_gres.conf.erb
```diff
@@ -0,0 +1,2 @@
+# Sample dummy node gres entry
```
demartinofra: shall we add the same comment set by sqswatcher or just say that this file will be generated by sqswatcher at runtime?

recipes/_compute_slurm_config.rb
rexcsn: we need to run `nvidia-smi` on the compute nodes at least once so that the GPU devices are mounted in `/dev/nvidia0` etc.

demartinofra: with `if !has_gpu.strip.empty?` aren't we just checking that the string is not empty and not the result of the command?

rexcsn: Yes, we are only checking if the string is empty. If the instance does not have GPU the grep on `lspci` will fail and return an empty string, if instance has at least 1 GPU the grep will return something. So if the grep is not empty(instance has GPU) we run `nvidia-smi` which will automatically mount the GPU devices so they are available for use.

lukeseawalker: what if instance has GPU but Nvidia drivers are not installed, e.g. creation of a custom AMI without drivers installation? Can the execution of `nvidia-smi` skipped in this case?

PR #428, opened 2019-11-11
Enable Tox linters and Python unit tests

enrico-usai:
* configure logger in the main function because when importing _pcluster_dcv_authenticator_, the code was trying to initialize the _LOGGER_ by writing a file in the log dir (that is not present in the test environment)
* format code

I tested the code in a running cluster, by replacing the _pcluster_dcv_authenticator.py_ and restarting the supervisord daemon.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,25 @@
+[flake8]
+ignore =
+    D100, # Missing docstring in public module
+    D104, # Missing docstring in public package
+    D107, # Missing docstring in __init__
+    D103, # Missing docstring in public function
+    W503, # line break before binary operator => Conflicts with black style.
+    D413, # Missing blank line after last section
+exclude =
+    .tox,
+    .git,
+    .pytest_cache,
+    docs/source/conf.py,
+    build,
+    dist,
+    tests/fixtures/*,
+    *.pyc,
+    *.egg-info,
+    .cache,
+    .eggs
+max-complexity = 10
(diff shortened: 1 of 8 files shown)


.flake8
```diff
@@ -0,0 +1,41 @@
...
+    W503, # line break before binary operator => Conflicts with black style.
+    D413, # Missing blank line after last section
+# D103 Missing docstring in public function
+# E402 module level import not at top of file
+# D101 Missing docstring in public class
+# D102 Missing docstring in public method
+# D205 1 blank line required between summary line and description
+# D400 First line should end with a period
+# D401 First line should be in imperative mood
+per-file-ignores =
```
demartinofra: this can be removed

.isort.cfg
```diff
@@ -0,0 +1,14 @@
...
+# 3 - Vertical Hanging Indent
+# from third_party import (
+#     lib1,
+#     lib2,
+#     lib3,
+#     lib4,
+# )
+multi_line_output=3
+include_trailing_comma=true
+skip=pcluster/resources/batch/custom_resources_code/crhelper
```
demartinofra: can be removed

.flake8
enrico-usai: removed

.isort.cfg
enrico-usai: done

PR #433, opened 2019-11-13
Install libnuma

sean-smith:

### DIFF
--- a/recipes/intel_install.rb
+++ b/recipes/intel_install.rb
@@ -15,6 +15,17 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


recipes/intel_install.rb
```diff
@@ -15,6 +15,17 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+libnuma = "#{node['cfncluster']['sources_dir']}/libnuma.tar.gz"
```
demartinofra: I think this needs to go in the dir where the license is fetched (by intelmpi installer?)

enrico-usai: let's move it in the `/usr/lib64` folder that contains the `/usr/lib64/libnuma.so.1` library

sean-smith: ok updated

PR #439, opened 2019-11-21
Fix gpg command parsing issues

tilne:
The recipe for installing the CloudWatch agent verifies that the public
key provided by the service has the expected fingerprint. To do this, the
output of the `gpg` command was being parsed. As it turns out, there are
some differences in the way the output is formatted on ubuntu18 as
opposed to the other supported distros.

On ubuntu18, the output of the command that was being used looks like
this:
```
root@ip-172-31-27-82:~# gpg --list-keys 'Amazon CloudWatch Agent'
pub   rsa2048 2017-11-14 [SC]
      937616F3450B7D806CBD9725D58167303B789C72
uid           [ unknown] Amazon CloudWatch Agent
```

On the other supported distros (namely, amazon linux in the following
example) the output looks like this:
```
[root@ip-172-31-23-223 ~]# gpg --list-keys 'Amazon CloudWatch Agent'
pub   2048R/3B789C72 2017-11-14
uid       [ unknown] Amazon CloudWatch Agent
```

Instead of trying to modify the parsing logic, I thought it best to handle it
using the [python-gnupg](https://pypi.org/project/python-gnupg/) module.

*Non-technical question*: is this allowed, since we distribute the
packages we install in the cookbook via our AMIs? It uses a BSD license.

### DIFF
--- a/files/default/verify_cloudwatch_agent_public_key_fingerprint.py
+++ b/files/default/verify_cloudwatch_agent_public_key_fingerprint.py
@@ -1,39 +1,26 @@
 #!/usr/bin/env python
 """Verify the fingerprint for the CloudWatch agent's public key is as expected."""
 
-import re
-import subprocess
 import sys
+import gnupg
 
 # This value comes from the CloudWatch agent's documentation:
 # https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/verify-CloudWatch-Agent-Package-Signature.html
-EXPECTED_FINGERPRINT = "9376 16F3 450B 7D80 6CBD  9725 D581 6730 3B78 9C72"
+EXPECTED_FINGERPRINT = "937616F3450B7D806CBD9725D58167303B789C72"
 
 
-def get_key_identifier():
-    """Get identifier for cloudwatch agent's key."""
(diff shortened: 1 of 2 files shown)


files/default/verify_cloudwatch_agent_public_key_fingerprint.py
```diff
@@ -1,39 +1,26 @@
...
-    """Get the fingerprint for the key with ID key_id."""
-    output = subprocess.check_output(['gpg', '--fingerprint', key_id]).decode('utf-8')
-    for line in output.splitlines():
-        match = re.search(r'Key fingerprint = (.*)', line.strip())
-        if match:
-            return match.group(1)
-    sys.exit('Unable to get fingerprint for cloudwatch agent public key')
+def get_key_fingerprint():
+    """Get identifier for CloudWatch agent's key."""
+    keys = gnupg.GPG().list_keys()
```
lukeseawalker: question, is there a way to retrieve just the key needed?

tilne: Good point. There definitely is.

tilne: The returned value is still a list, so the loop is still needed. But the list should have only a single entry.

PR #440, opened 2019-11-21
Fix CloudWatch agent kitchen test

tilne:
This PR fixes two issues with the CloudWatch agent kitchen test:
1. The original version uses the `[[` conditional operator, which is not
provided in every shell. This was causing test failures when running on
ubuntu AMIs.
2. The original version also assumes that `jq` is installed. After
looking at the process used to select the custom AMIs used in the
kitchen tests, this is definitely not a valid assumption.

The fix uses the return code from `grep`ing the following JSON format:
```
ubuntu@ip-172-31-3-190:~$ /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a status
{
  "status": "running",
  "starttime": "2019-11-21T21:08:13+00:00",
  "version": "1.231221.0"
}
```

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -220,11 +220,11 @@ case node['cfncluster']['cfn_cluster_cw_logging_enabled']
 when 'true'
   execute 'cloudwatch-agent-status-running' do
     user 'root'
-    command "[[ $(/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a status | jq --raw-output .status) = running ]] || exit 1"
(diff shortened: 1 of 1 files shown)


lukeseawalker:
minor comment for 2:
-  jq should be installed, at list for non -custom AMIs. There is a test using it in this same file, see https://github.com/aws/aws-parallelcluster-cookbook/pull/440/files#diff-f8c99fdb0f930cc719bac8e80e7a6504R139. Probably didn't worked out because PATH wasn't specified.

tilne:
> * jq should be installed, at list for non -custom AMIs

Yeah it was breaking kitchen tests for custom AMIs. I could've kept it the same and wrapped it in a guard like the test you pointed me towards, but I had to change the executed command anyway because of the first issue.

PR #442, opened 2019-11-27
Add DCV logs to CloudWatch logs integration config

tilne:
This PR implements two new pieces of functionality:
1. DCV logs are now logged via CloudWatch when enabled for a cluster
2. A script for more easily adding to and validating the JSON file used
to create the CloudWatch agent config file. The CloudWatch agent
configuration recipe makes use of it.

Of particular note for this review are the implications of using a
wildcard in the `file_path` value of the `log_configs` entries with
`log_stream_name`s `dcv-xsession` and `dcv-agent`. According to the
[CloudWatch agentdocumentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html)
this results in only data for the latest file (by modification time)
being pushed to CloudWatch. Since the wildcard used in the `file_path`s
for those two entries represent a \<username>.\<session ID> pair, this
motivates two questions:
1. Are simultaneous sessions supported?
2. Once a session is closed, can it be reopened?

If the answer to both of those is "no", then it's probably fine. If not,
then generating configs on the fly, regenerating a CloudWatch agent
config file, and restarting the agent would be worth considering.


files/default/cloudwatch_logs/cloudwatch_log_configs_util.py
```diff
@@ -0,0 +1,180 @@
+"""
+Validate and modify the data in the cloudwatch_log_files.json cookbook file.
+
+This file is used to validate and add data to the JSON file that's used to
+configure the CloudWatch agent on a cluster's EC2 instances. The structure of
+the new and/or existing data is validated (to an extent).
```
sean-smith: Can you add how it's validated?

files/default/cloudwatch_logs/cloudwatch_log_configs_util.py
```diff
@@ -0,0 +1,180 @@
...
+    return _read_json_at(LOG_CONFIGS_PATH)
+
+
+def _validate_json_schema(input_json):
+    """Ensure the structure of input_json matches the schema."""
+    schema = _read_schema()
+    try:
+        jsonschema.validate(input_json, schema)
+    except jsonschema.exceptions.ValidationError as validation_err:
+        sys.exit(str(validation_err))
```
sean-smith: Pull this out into a `fail()` function

files/default/cloudwatch_logs/cloudwatch_log_files.json
```diff
@@ -293,6 +309,175 @@
...
+      "platforms": [
+        "centos",
+        "ubuntu"
+      ],
+      "node_roles": [
+        "MasterServer"
+      ],
+      "feature_conditions": [
+        {
+          "dna_key": "dcv_enabled",
```
sean-smith: Nice work 👍

files/default/cloudwatch_logs/cloudwatch_log_configs_util.py
tilne: Done.

files/default/cloudwatch_logs/cloudwatch_log_configs_util.py
tilne: Done.

PR #446, opened 2019-12-03
Add set -e on top of each bash script in cookbooks

ddeidda:
This commit adds the missing "set -e" instructions to all bash scripts
to make them fail in case of errors.

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -29,11 +29,12 @@ end
 bash 'check awscli regions' do
   cwd Chef::Config[:file_cache_path]
   code <<-AWSREGIONS
+    set -e
     export PATH="/usr/local/bin:/usr/bin/:$PATH"
     regions=($(#{node['cfncluster']['cookbook_virtualenv_path']}/bin/aws ec2 describe-regions --region us-east-1 --query "Regions[].{Name:RegionName}" --output text))
     for region in "${regions[@]}"
     do
-      #{node['cfncluster']['cookbook_virtualenv_path']}/bin/aws ec2 describe-regions --region "${region}" >/dev/null 2>&1 || exit 1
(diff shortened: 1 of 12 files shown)


recipes/tests.rb
```diff
@@ -157,6 +159,7 @@
   bash 'test nvidia driver install' do
     cwd Chef::Config[:file_cache_path]
     code <<-TESTNVIDIA
+      set -e
```
lukeseawalker: must be not put here, it's few lines later

recipes/tests.rb
```diff
@@ -183,6 +186,7 @@
   bash 'test CUDA install' do
     cwd Chef::Config[:file_cache_path]
     code <<-TESTCUDA
+      set -e
```
lukeseawalker: must be not put here, it's few lines later

recipes/tests.rb
ddeidda: OK :+1:

recipes/tests.rb
ddeidda: OK :+1:

recipes/tests.rb
```diff
@@ -29,6 +29,7 @@
 bash 'check awscli regions' do
   cwd Chef::Config[:file_cache_path]
   code <<-AWSREGIONS
+    set -e
```
lukeseawalker: we could remove the exit 1

recipes/tests.rb
```diff
@@ -139,6 +140,7 @@
   bash 'execute jq' do
     cwd Chef::Config[:file_cache_path]
     code <<-JQMERGE
+      set -e
```
lukeseawalker: we could remove the exit 1

recipes/tests.rb
ddeidda: Done

recipes/tests.rb
ddeidda: Done

PR #449, opened 2019-12-04
Change nfs export to work with VPC containing multiple CIDR blocks

rexcsn:
* Changed from getting vpc-ipv4-cidr-block, which only returns the first CIDR block, to getting vpc-ipv4-cidr-blocks, which returns all CIDR blocks for a VPC, from metadata call.
* Changed all nfs export to export to all CIDR blocks in a VPC

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -95,14 +95,15 @@ def get_1st_partition(device)
 end
 
 #
-# Get vpc-ipv4-cidr-block
+# Get vpc-ipv4-cidr-blocks
 #
-def get_vpc_ipv4_cidr_block(eth0_mac)
-  uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{eth0_mac.downcase}/vpc-ipv4-cidr-block")
+def get_vpc_ipv4_cidr_blocks(eth0_mac)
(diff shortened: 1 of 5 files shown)


libraries/helpers.rb
```diff
@@ -95,14 +95,15 @@ def get_1st_partition(device)
...
 #
-def get_vpc_ipv4_cidr_block(eth0_mac)
-  uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{eth0_mac.downcase}/vpc-ipv4-cidr-block")
+def get_vpc_ipv4_cidr_blocks(eth0_mac)
+  uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{eth0_mac.downcase}/vpc-ipv4-cidr-blocks")
   res = Net::HTTP.get_response(uri)
-  vpc_ipv4_cidr_block = res.body if res.code == '200'
-
-  vpc_ipv4_cidr_block
+  vpc_ipv4_cidr_blocks = res.body if res.code == '200'
```
demartinofra: I know it's not part of your change but I'm wondering what happens if res.code != '200'

rexcsn: I am looking into this: https://docs.ruby-lang.org/en/2.0.0/Net/HTTP.html#class-Net::HTTP-label-Following+Redirection.
Merging this PR first, will open another one to address this issue.

PR #460, opened 2019-12-09
Remove dead case from CW agent kitchen test

tilne:
Since kitchen tests run without inputs from CFN, the agent will never
be up-and-running. Remove the test case for that scenario accordingly.


sean-smith:
Can we enable it in the kitchen tests?

tilne:
> Can we enable it in the kitchen tests?

it doesn't matter that much in my opinion. the main point is just to make sure that the agent's status can be queried. also, i'm not sure if starting the agent would result in log groups being created or not.

PR #484, opened 2020-01-09
DCV on Ubuntu 18

rexcsn:
* Extend DCV support to Ubuntu18
* Install/configure DCV packages on Ubuntu18, following the same basic logic as Centos7 installation

### DIFF
--- a/amis/packer_ubuntu1804.json
+++ b/amis/packer_ubuntu1804.json
@@ -10,6 +10,7 @@
     "build_date" : "{{env `BUILD_DATE`}}",
     "vendor_path" : "{{env `VENDOR_PATH`}}",
     "nvidia_enabled" : "{{env `NVIDIA_ENABLED`}}",
+    "dcv_installed": "{{env `DCV_INSTALLED`}}",
     "instance_type" : "{{env `AWS_FLAVOR_ID`}}",
     "subnet_id" : "{{env `AWS_SUBNET_ID`}}",
     "security_group_id" : "{{env `AWS_SECURITY_GROUP_ID`}}",
@@ -233,6 +234,9 @@
           "cfn_region": "{{user `region`}}",
           "nvidia" : {
             "enabled" : "{{user `nvidia_enabled`}}"
+          },
+          "dcv": {
+            "installed" : "{{user `dcv_installed`}}"
           },
           "custom_node_package" : "{{user `custom_node_package`}}"
         }
--- a/files/default/dcv/pcluster_dcv_authenticator.py
+++ b/files/default/dcv/pcluster_dcv_authenticator.py
@@ -376,14 +376,14 @@ class DCVAuthenticator(BaseHTTPRequestHandler):
         """Check if there is a dcvagent process running for the given user and for the given session_id."""
         # row example:
         # centos 63 0.0 0.0 4348844 3108   ??  Ss   23Jul19   2:32.46  /usr/libexec/dcv/dcvagent --session-id mysession
+        # ubuntu 2949 0.3 0.4 860568 34328 ? Sl 20:10 0:18 /usr/lib/x86_64-linux-gnu/dcv/dcvagent --session-id mysession
         fields = row.split()
         command_index = 10
         session_name_index = 12
         user_index = 0
-        dcv_agent_path = "/usr/libexec/dcv/dcvagent"
 
         return (
-            fields[command_index] == dcv_agent_path
+            fields[command_index].endswith("/dcv/dcvagent")
             and fields[user_index] == user
             and fields[session_name_index] == session_id
         )
--- a/recipes/dcv_config.rb
+++ b/recipes/dcv_config.rb
@@ -34,11 +34,20 @@ def allow_gpu_acceleration
   end
 
   # dcvgl package must be installed after NVIDIA and before starting up X
+  # DO NOT install dcv-gl on non-GPU instances, or will run into a black screen issue
   dcv_gl = "#{node['cfncluster']['sources_dir']}/#{node['cfncluster']['dcv']['package']}/#{node['cfncluster']['dcv']['gl']}"
-  package dcv_gl do
-    action :install
-    source dcv_gl
+  case node['platform']
+  when 'centos'
+    package dcv_gl do
+      action :install
+      source dcv_gl
+    end
+  when 'ubuntu'
+    execute 'apt install dcv-gl' do
+      command "apt -y install #{dcv_gl}"
+    end
   end
+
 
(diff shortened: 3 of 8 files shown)


amis/packer_ubuntu1804.json
```diff
@@ -123,6 +124,7 @@
         {
           "device_name" : "/dev/sda1",
           "volume_type" : "gp2",
+          "volume_size" : "25",
```
lukeseawalker: we should not force a volume-size when using a provided custom AMI, because we cannot know in advance the size of the default volume.
Suppose that the provided custom AMI has a default volume size of 100GB, you'll get an error because the volume cannot be shrunk back to 25 GB

files/default/dcv/pcluster_dcv_authenticator.py
```diff
@@ -376,14 +376,14 @@ def check_dcv_process(row, user, session_id):
...
+        # ubuntu 2949 0.3 0.4 860568 34328 ? Sl 20:10 0:18 /usr/lib/x86_64-linux-gnu/dcv/dcvagent --session-id mysession
         fields = row.split()
         command_index = 10
         session_name_index = 12
         user_index = 0
-        dcv_agent_path = "/usr/libexec/dcv/dcvagent"
 
         return (
-            fields[command_index] == dcv_agent_path
+            fields[command_index].endswith("dcvagent")
```
lukeseawalker: what about `.endswith("/dcv/dcvagent")`?

recipes/dcv_config.rb
```diff
@@ -59,7 +68,7 @@ def allow_gpu_acceleration
   end
 end
 
-if node['platform'] == 'centos' && node['platform_version'].to_i == 7 && node['cfncluster']['cfn_node_type'] == "MasterServer"
+if node['cfncluster']['dcv']['supported_os'].include?("#{node['platform']}#{node['platform_version'].to_i}") && node['cfncluster']['cfn_node_type'] == "MasterServer"
```
lukeseawalker: I don't like this check. Can we be more precise?
Also I don't understand how could it work when `platform_version = 18.04`. Is it included in ['centos7', 'ubuntu18']?

recipes/dcv_install.rb
```diff
@@ -15,11 +15,18 @@
 # See the License for the specific language governing permissions and limitations under the License.
 
 # Utility function to install a list of rpm packages
```
lukeseawalker: minor change the comment as well: "_rpm_"

recipes/dcv_install.rb
```diff
@@ -15,11 +15,18 @@
...
-      source package_name
+    case node['platform']
+    when 'centos'
+      package package_name do
+        action :install
+        source package_name
+      end
+    when 'ubuntu'
+      execute 'apt install dcv package' do
+        command "apt -y install #{package_name}"
```
lukeseawalker: Just for reference why `source` is not working for `apt`
see https://github.com/chef/chef/issues/9014 and https://github.com/chef/chef/issues/5760

recipes/dcv_install.rb
```diff
@@ -66,21 +73,41 @@ def disable_lock_screen
   not_if { ::File.exist?("#{node['cfncluster']['scripts_dir']}/pcluster_dcv_connect.sh") }
 end
 
-
-if node['platform'] == 'centos' && node['platform_version'].to_i == 7 && !File.exist?("/etc/dcv/dcv.conf")
+if node['cfncluster']['dcv']['supported_os'].include?("#{node['platform']}#{node['platform_version'].to_i}") && !File.exist?("/etc/dcv/dcv.conf")
```
lukeseawalker: Some from the comment above

recipes/dcv_install.rb
```diff
@@ -66,21 +73,41 @@ def disable_lock_screen
...
+      end
+      # Install X Window System (required when using GPU acceleration)
+      package "xorg-x11-server-Xorg"
+
+    when 'ubuntu'
+      # Install the desktop environment and the desktop manager packages
+      # Must purge ifupdown before creating the AMI or the instance will have an ssh failure
+      bash 'install pre-req' do
+        user 'root'
+        group 'root'
```
lukeseawalker: minor I think `user root` and `group root` are the default, so could be omitted

recipes/dcv_install.rb
```diff
@@ -101,7 +128,7 @@ def disable_lock_screen
     dcv_packages_path = "#{node['cfncluster']['sources_dir']}/#{node['cfncluster']['dcv']['package']}/"
     # Rewrite dcv_packages object by cycling each package file name and appending the path to them
     dcv_packages.map! { |package| dcv_packages_path + package }
-    install_rpm_packages(dcv_packages)
+    install_list_packages(dcv_packages)
```
lukeseawalker: minor `install_packages_list` or `install_packages`

templates/default/parallelcluster_supervisord.conf.erb
```diff
@@ -15,7 +15,7 @@ redirect_stderr = true
 stdout_logfile = /var/log/jobwatcher
 <% end -%>
 
-<% if node['cfncluster']['dcv_enabled'] == "master" && node['platform'] == 'centos' && node['platform_version'].to_i == 7 -%>
+<% if node['cfncluster']['dcv_enabled'] == "master" && node['cfncluster']['dcv']['supported_os'].include?("#{node['platform']}#{node['platform_version'].to_i}") -%>
```
lukeseawalker: some comment above

recipes/dcv_install.rb
```diff
@@ -66,21 +73,41 @@ def disable_lock_screen
...
+          apt -y install ubuntu-desktop
+          apt -y purge ifupdown
+          DEBIAN_FRONTEND=noninteractive apt -y install lightdm
+          apt -y install mesa-utils
+          wget https://d1uj6qtbmh3dt5.cloudfront.net/NICE-GPG-KEY
+          gpg --import NICE-GPG-KEY
+        PREREQ
+      end
     end
     disable_lock_screen
```
enrico-usai: please double check if the lock screen disable works for ubuntu.

recipes/dcv_config.rb
rexcsn: `node['platform_version'].to_i` converts `18.04` to `18`, but yes we can be more precise and just use `"#{node['platform']}#{node['platform_version']}"`

recipes/dcv_install.rb
rexcsn: The same procedure works for Centos and Ubuntu: https://askubuntu.com/a/1184810
Also confirming that lock screen is disabled with manual testing

PR #485, opened 2020-01-10
Intel MPI U6

sean-smith:
Upgrade to Intel MPI U6.

### DIFF
--- a/recipes/intel_mpi.rb
+++ b/recipes/intel_mpi.rb
@@ -16,17 +16,11 @@
 # limitations under the License.
 
 intelmpi_modulefile = "#{node['cfncluster']['modulefile_dir']}/intelmpi/#{node['cfncluster']['intelmpi']['version']}"
-intelmpi_installer = "#{node['cfncluster']['sources_dir']}/aws_impi.sh"
-if node['cfncluster']['cfn_region'].start_with?("cn-")
-  s3_suffix = '.cn'
-else
-  s3_suffix = ''
-end
-intelmpi_url = "https://#{node['cfncluster']['cfn_region']}-aws-parallelcluster.s3.#{node['cfncluster']['cfn_region']}.amazonaws.com#{s3_suffix}/#{node['cfncluster']['intelmpi']['s3_path']}"
+intelmpi_installer = "#{node['cfncluster']['sources_dir']}/l_mpi_#{node['cfncluster']['intelmpi']['version']}.tgz"
 
 # fetch intelmpi installer script
 remote_file intelmpi_installer do
-  source intelmpi_url
+  source node['cfncluster']['intelmpi']['url']
   mode '0744'
   retries 3
   retry_delay 5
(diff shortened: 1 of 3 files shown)


recipes/intel_mpi.rb
```diff
@@ -39,7 +39,10 @@
   cwd node['cfncluster']['sources_dir']
   code <<-INTELMPI
     set -e
-    ./aws_impi.sh install -check_efa 0 -version #{node['cfncluster']['intelmpi']['version']}
+    tar -xf l_mpi_#{node['cfncluster']['intelmpi']['version']}.tgz
```
demartinofra: can we fetch this as part of a remote_file directive and have it downloaded in the sources dir or tmp dir?
you could just update the part above that downloads the installer which is not used anymore.

sean-smith: This does download into the sources directory already. Not sure what you mean.

sean-smith:
Yes, I handled that in https://github.com/aws/aws-parallelcluster-cookbook/pull/487

recipes/intel_mpi.rb
demartinofra: I think you can ignore my comment :)

recipes/intel_mpi.rb
```diff
@@ -39,10 +33,19 @@
...
+    cd l_mpi_#{node['cfncluster']['intelmpi']['version']}/
+    ./install.sh -s silent.cfg --accept_eula
   INTELMPI
-  creates '/opt/intel/impi'
+  creates "/opt/intel/impi/#{node['cfncluster']['intelmpi']['version']}"
+end
+
+bash "create modulefile" do
+  cwd node['cfncluster']['modulefile_dir']
+  code <<-MODULEFILE
```
demartinofra: set -e flag or use command

sean-smith: done

recipes/tests.rb
```diff
@@ -136,11 +136,27 @@
       command "rpm -qa | grep libfabric && rpm -qa | grep efa-"
       user node['cfncluster']['cfn_cluster_user']
     end
+    execute 'check intel mpi installed' do
```
demartinofra: Leaving the comment here but it refers to L132: Aren't we installing EFA and intel mpi in cn regions too?

recipes/tests.rb
```diff
@@ -136,11 +136,27 @@
...
   when 'ubuntu1604', 'ubuntu1804'
     execute 'check efa rpm installed' do
       command "dpkg -l | grep libfabric && dpkg -l | grep 'efa '"
       user node['cfncluster']['cfn_cluster_user']
     end
+    execute 'check intel mpi installed' do
+      command "dpkg -l | grep intel-mpi"
+      user node['cfncluster']['cfn_cluster_user']
+    end
+    execute 'check intel mpi version' do
```
demartinofra: minor: you could move this outside of when and avoid the duplication

recipes/tests.rb
sean-smith: yes, I've removed not `cn` check

recipes/tests.rb
sean-smith: done

PR #501, opened 2020-01-21
Install Ganglia from package repositories for ALinux, Centos6/7

lukeseawalker:
For ALinux, CentOS6 and CentOS7 install Ganglia from package repositories instead of installing it from source.
This avoid compilation issues that could raise errors when using a custom AMI.

Tested manually creating an AMI starting from both the default AMI and a custom one like DeepLearning ALinux AMI and enabling ganglia at cluster creation time.

Ganglia version installed from repositories is 3.7.2 for ALinux, CentOS6 and CentOS7, which is the same as the one previously downloaded frm GitHub.


demartinofra:
Do you think we can install ganglia from package repositories for all OSes?

lukeseawalker:
> Do you think we can install ganglia from package repositories for all OSes?

This requires further investigation. I will have a look

PR #507, opened 2020-01-27
Add support for Amazon Linux 2

tilne:
This PR contains the changes required to support on amazon linux 2 all of the features for which we have integration tests (with the exception of the intel HPC suite which is only supported for centos 7).

Testing:
* Running all of the integration tests (except for the scaling and benchmark tests) locally

A couple of notes:
* There is a workaround for an issue in setting hostnames in the base_config recipe. The hostname cookbook we use for the other OSes doesn't work the same on alinux2. Specifically, it sets the hostname to the fqdn, which breaks some things (torque, for example).
* The changes required for the kitchen tests are (probably) not all in this PR. My plan is to make those when attempting to get the kitchen tests running.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -78,24 +78,20 @@ default['cfncluster']['env2']['url'] = 'https://sourceforge.net/projects/env2/fi
 # NICE DCV
 default['cfncluster']['dcv']['installed'] = 'yes'
 default['cfncluster']['dcv']['version'] = '2019.1-7644'
-default['cfncluster']['dcv']['supported_os'] = %w[centos7 ubuntu18]
-case node['platform']
-when 'centos'
-  if node['platform_version'].to_i == 7
-    default['cfncluster']['dcv']['package'] = "nice-dcv-#{node['cfncluster']['dcv']['version']}-el7"
-    default['cfncluster']['dcv']['server'] = "nice-dcv-server-2019.1.7644-1.el7.x86_64.rpm" # NICE DCV server package
-    default['cfncluster']['dcv']['xdcv'] = "nice-xdcv-2019.1.226-1.el7.x86_64.rpm" # required to create virtual sessions
-    default['cfncluster']['dcv']['gl'] = "nice-dcv-gl-2019.1.544-1.el7.x86_64.rpm" # required to enable GPU sharing
-    default['cfncluster']['dcv']['sha256sum'] = "ab50323c92b3584ac88f697fd45e68b98da0b1b334a3e1f7eef6343df3aa4d91"
-  end
-when 'ubuntu'
-  if node['platform_version'] == '18.04'
-    default['cfncluster']['dcv']['package'] = "nice-dcv-#{node['cfncluster']['dcv']['version']}-ubuntu1804"
-    default['cfncluster']['dcv']['server'] = "nice-dcv-server_2019.1.7644-1_amd64.ubuntu1804.deb" # NICE DCV server package
-    default['cfncluster']['dcv']['xdcv'] = "nice-xdcv_2019.1.226-1_amd64.ubuntu1804.deb"  # required to create virtual sessions
-    default['cfncluster']['dcv']['gl'] = "nice-dcv-gl_2019.1.544-1_amd64.ubuntu1804.deb"  # required to enable GPU sharing
-    default['cfncluster']['dcv']['sha256sum'] = "41a0b4112c435b3a57de3ae46cfe8cdf90c33a216f488e3bfab944f034f6067f"
-  end
+default['cfncluster']['dcv']['supported_os'] = %w[centos7 ubuntu18 amazon2]
+case "#{node['platform']}#{node['platform_version'].to_i}"
+when 'centos7', 'amazon2'
+  default['cfncluster']['dcv']['package'] = "nice-dcv-#{node['cfncluster']['dcv']['version']}-el7"
+  default['cfncluster']['dcv']['server'] = "nice-dcv-server-2019.1.7644-1.el7.x86_64.rpm" # NICE DCV server package
+  default['cfncluster']['dcv']['xdcv'] = "nice-xdcv-2019.1.226-1.el7.x86_64.rpm" # required to create virtual sessions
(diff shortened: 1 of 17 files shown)


attributes/default.rb
```diff
@@ -96,6 +96,14 @@
     default['cfncluster']['dcv']['gl'] = "nice-dcv-gl_2019.1.544-1_amd64.ubuntu1804.deb"  # required to enable GPU sharing
     default['cfncluster']['dcv']['sha256sum'] = "41a0b4112c435b3a57de3ae46cfe8cdf90c33a216f488e3bfab944f034f6067f"
   end
+when 'amazon'
```
lukeseawalker: since the following definition are the same used for centos7, can we try to put them in common?
Also, given that we define
```
default['cfncluster']['dcv']['supported_os'] = %w[centos7 ubuntu18 amazon2]
```
we could also double check if ```if node['platform_version'].to_i```  is necessary or could be omitted and just leave the switch case for platform

attributes/default.rb
```diff
@@ -185,11 +193,24 @@
...
-    default['cfncluster']['base_packages'] = %w[vim ksh tcsh zsh openssl-devel ncurses-devel pam-devel net-tools openmotif-devel
-                                                libXmu-devel hwloc-devel db4-devel tcl-devel automake autoconf pyparted libtool
-                                                httpd boost-devel redhat-lsb mlocate mpich-devel R atlas-devel fftw-devel
-                                                libffi-devel dkms mysql-devel libedit-devel postgresql-devel postgresql-server
-                                                sendmail cmake byacc libglvnd-devel mdadm libgcrypt-devel]
+    case node['platform_version'].to_i
+    when 2
+      default['cfncluster']['base_packages'] = %w[vim ksh tcsh zsh openssl-devel ncurses-devel pam-devel net-tools openmotif-devel
+                                                  hwloc-devel libdb-devel libxml2-devel libgcrypt-devel boost-devel perl-devel
+                                                  dpkg-dev tar gzip python-pip]
```
lukeseawalker: is ```python-pip``` really needed?
Also let's double check the list, because for example I see mdadm missing, which should be needed for RAID

recipes/_lustre_install.rb
```diff
@@ -87,7 +87,7 @@
   end
 
   kernel_module 'lnet'
-elsif node['platform'] == 'amazon'
+elsif node['platform'] == 'amazon' and node['platform_version'].to_i == 1
```
lukeseawalker: As you wrote some diff above
>       # Platform_version is a year rather than a version for the first major version of amazon linux.
so this ```node['platform_version'].to_i == 1``` would not work, since `platform_version.to_i` should be `2018`

recipes/base_install.rb
```diff
@@ -18,7 +18,12 @@
 case node['platform_family']
 when 'rhel', 'amazon'
   include_recipe 'yum'
-  include_recipe "yum-epel" if node['platform_version'].to_i < 7
+  if node['platform_family'] == 'amazon' and node['platform_version'].to_i == 2
+    execute "amazon-linux-extras install -y epel"
```
lukeseawalker: do we need `epel` for alinux2?

tilne: Yes. It's used to install a lot of the base packages.

recipes/_lustre_install.rb
tilne: Ah yes, missed this.

attributes/default.rb
tilne: Yeah i didn't handle the conditionals together because the case statement keys on the platform family without respect to the version. But you are right, it should be done that way. I'll change the case statement to key on the same <platform family><platform version int> combo used in the supported_os list.

attributes/default.rb
tilne: I tried to keep this list as minimal as possible. I only included a package if it was needed to execute compilation recipes or run integration tests. I believe python-pip was needed, but i'll need to check for which package.

For mdam or other missisng packages, this set was sufficient for building AMIs and successfully running all of the integration tests on those built AMIs.

tilne: The following are installed as part of the first provisioner to run when building the AMIs:
```
autoconf
automake
bison
flex
gcc
gcc-c++
libtool
patch
rpm-build
rpm-sign
system-rpm-config
byacc
cscope
ctags
diffstat
doxygen
elfutils
gcc-gfortran
git
indent
intltool
patchutils
rcs
subversion
swig
systemtap
curl
install
jq
wget
```

so those should probably be in this list to handle the use of non-pcluster AMIs.

tilne: python-pip wasn't needed as it turns out. i've removed it.

tilne: After trying to start a cluster with the modified cookbook, it looks like python-pip is actually required by the python_package resource used to install the batch CLI. Adding it back now.

attributes/default.rb
```diff
@@ -190,6 +186,22 @@
                                                 httpd boost-devel redhat-lsb mlocate mpich-devel R atlas-devel fftw-devel
                                                 libffi-devel dkms mysql-devel libedit-devel postgresql-devel postgresql-server
                                                 sendmail cmake byacc libglvnd-devel mdadm libgcrypt-devel]
+    if node['platform_version'].to_i == 2
+      # Swap out some packages for their alinux2 equivalents
+      [%w[db4-devel libdb-devel], %w[mpich-devel openmpi-devel], %w[redhat-lsb system-lsb]].each do |al1, al2equiv|
```
lukeseawalker: let's sync on the swap [mpich-devel openmpi-devel]

resources/alinux_extras_topic.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+resource_name :alinux_extras_topic
```
lukeseawalker: nice!

attributes/default.rb
tilne: I messaged you about this but for anyone that's interested: neither of the two will be installed since we already install two different MPI libraries.

PR #510, opened 2020-01-29
Install whoopsie before ubuntu-desktop

tilne:
When `whoopsie` is installed as part of the installation of
`ubuntu-desktop` it crashes and results in a popup that says "System
program problem detected" when a user first connects to the first DCV
session. When it's already installed this does not to happen.

### DIFF
--- a/recipes/dcv_install.rb
+++ b/recipes/dcv_install.rb
@@ -94,6 +94,7 @@ if node['cfncluster']['dcv']['supported_os'].include?("#{node['platform']}#{node
         cwd Chef::Config[:file_cache_path]
         code <<-PREREQ
           set -e
+          apt -y install whoopsie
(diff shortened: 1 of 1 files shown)


recipes/dcv_install.rb
```diff
@@ -94,6 +94,7 @@ def disable_lock_screen
         cwd Chef::Config[:file_cache_path]
         code <<-PREREQ
           set -e
+          apt -y install whoopsie
```
sean-smith: can these two lines be a one-liner?

tilne: I did it on separate lines because I didn't want apt to see that whoopsie would be installed as part of ubuntu-desktop anyways, and then not install it first. That would make it effectively the same as it is now. I'm not sure if that's actually how it would work.

PR #514, opened 2020-01-30
Remove intel mpi installer after install

sean-smith:
Removes the installer source but keeps `EULA.txt` in the `/opt/intel/impi` directory.


recipes/intel_mpi.rb
```diff
@@ -36,6 +36,9 @@
     tar -xf l_mpi_#{node['cfncluster']['intelmpi']['version']}.tgz
     cd l_mpi_#{node['cfncluster']['intelmpi']['version']}/
     ./install.sh -s silent.cfg --accept_eula
+    mv rpm/EULA.txt /opt/intel/impi
```
tilne: should this path also include the intel MPI version like it does below?

sean-smith: Yep - nice catch

PR #518, opened 2020-01-31
More robust downloading of SGE archive signing key

tilne:
The builds for alinux2 AMIs in China are consistently failing while
building SGE. They are failing when trying to import the public key used
to sign the debian archive that is used. This change uses a (hopefully)
more reliable method of importing the public key.

I didn't use a remote_file resource because this only needs to be done
for alinux2. Doing this via a shell command allows me to keep it out of
the general-purpose recipe used for all of the other OSes.

### DIFF
--- a/files/amazon-2/sge_preinstall.sh
+++ b/files/amazon-2/sge_preinstall.sh
@@ -1,5 +1,19 @@
-#!/bin/sh
+#!/bin/bash
 set -e
+
+if [ -z ${TARBALL_URL} ]; then
+  echo "TARBALL_URL must be set"
+  exit 1
+elif [ -z ${TARBALL_ROOT_DIR} ]; then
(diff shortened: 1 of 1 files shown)


files/amazon-2/sge_preinstall.sh
```diff
@@ -1,6 +1,20 @@
-#!/bin/sh
+#!/bin/sh -x
```
demartinofra: what about using /bin/bash so that we pick a specific implementation and do not rely on what is symlinked by sh? Also do we want to keep the -x?

tilne: Yeah it should definitely be bash. That was an oversight on my part.

tilne: Done.

PR #525, opened 2020-02-07
Fix Chef::Mount error on rerunning the recipes.

joeydumont:
*Description of changes:* 

When trying to re-run Chef in order to update the configuration of the master, I stumbled onto an error with the `mount shared_dir_array` block. Because the `ruby_block "setup_disk_#{index}` does not run on a second execution, `dev_uuids` is not never set, so the device in the mount block is `nil`. 

Here, I manually run the `get_uuid` to fill the `dev_uuids` array. This fixes the issue.


joeydumont:
Sorry, this should have been an issue, not a PR.

PR #529, opened 2020-02-11
Add kitchen test to check chrony service is running

lukeseawalker:

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -189,7 +189,17 @@ end
 ###################
 # Amazon Time Sync
 ###################
-execute 'check chrony' do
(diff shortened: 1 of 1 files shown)


recipes/tests.rb
```diff
@@ -189,7 +189,17 @@
 ###################
 # Amazon Time Sync
 ###################
-execute 'check chrony' do
+if node['init_package'] == 'init'
```
enrico-usai: What about using the [service resource](the https://docs.chef.io/resource_service.html)?

lukeseawalker: could not be used for that purpose

PR #531, opened 2020-02-11
Restore know host file management for CentOS6

lukeseawalker:
The OpenSSH version in CentOS6 doesn't support the Match directive that allows to conditionally disable the StrictHostKeyChecking parameter

Match directive in ssh_config is available since
OpenSSH 6.5/6.5p1 (2014-01-30)

OpenSSH version of CentOS6 is
OpenSSH 5.3/5.3p1 (2009-10-01)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -124,12 +124,14 @@ else
   default['openssh']['server']['m_a_cs'] = 'hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512,hmac-sha2-256'
 end
 default['openssh']['client']['gssapi_authentication'] = 'yes'
-default['openssh']['client']['match'] = 'exec "ssh_target_checker.sh %h"'
-# Disable StrictHostKeyChecking for target host in the cluster VPC
-default['openssh']['client']['  _strict_host_key_checking'] = 'no'
-# Do not store server key in the know hosts file to avoid scaling clashing
-# that is when an new host gets the same IP of a previously terminated host
-default['openssh']['client']['  _user_known_hosts_file'] = '/dev/null'
+unless node['platform'] == 'centos' && node['platform_version'].to_i < 7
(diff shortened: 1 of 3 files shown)


attributes/default.rb
```diff
@@ -124,12 +124,14 @@
   default['openssh']['server']['m_a_cs'] = 'hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512,hmac-sha2-256'
 end
 default['openssh']['client']['gssapi_authentication'] = 'yes'
-default['openssh']['client']['match'] = 'exec "ssh_target_checker.sh %h"'
-# Disable StrictHostKeyChecking for target host in the cluster VPC
-default['openssh']['client']['  _strict_host_key_checking'] = 'no'
-# Do not store server key in the know hosts file to avoid scaling clashing
-# that is when an new host gets the same IP of a previously terminated host
-default['openssh']['client']['  _user_known_hosts_file'] = '/dev/null'
+unless node['platform'] == 'centos' && node['platform_version'].to_i < 7
```
demartinofra: Optional: we could define a condition in conditions.rb for centos6

recipes/base_install.rb
```diff
@@ -86,12 +86,14 @@
 # Manage SSH via Chef
 include_recipe "openssh"
 
-# Install SSH target checker
-cookbook_file 'ssh_target_checker.sh' do
-  path "/usr/local/bin/ssh_target_checker.sh"
-  owner "root"
-  group "root"
-  mode "0755"
+unless node['platform'] == 'centos' && node['platform_version'].to_i < 7
```
demartinofra: Optional: We could also install this everywhere to avoid additional custom logic

attributes/default.rb
lukeseawalker: the value of the condition is not yet avail in the default.rb

recipes/base_install.rb
lukeseawalker: ok

PR #534, opened 2020-02-12
Eliminate unwanted side effects from pyenv resources

tilne:
This PR is meant to:
* Revert the temporary workaround for a slurm build issue introduced in #521 
* Eliminate the unwanted side effects that result from using the pyenv cookbook
* Change the names of some of the resources and their properties to more clearly express what they're intended to do


resources/pyenv_command.rb
```diff
@@ -0,0 +1,27 @@
...
+default_action :run
+
+action :run do
+  cmd_env = {
+    'PYENV_ROOT' => new_resource.pyenv_path,
+    'PATH' => "#{new_resource.pyenv_path}/bin:#{ENV['PATH']}",
+    'USER' => new_resource.user,
+    'HOME' => new_resource.home_dir
+  }
+  bash "pyenv #{new_resource.command}" do
```
sean-smith: Please move the Python Centos 6 to use this instead of pyenv cookbook

tilne: I looked into this and it's more work than I thought it would be. Unless there's a compelling reason to invest the time I'd like to leave it as is and drop the centos6 case when that OS is deprecated.

tilne:
> Is it possible to add kitchen tests to do some sanity checks like for example verify that the right version of python is installed in the virtualenv we are creating?

tilne:
Use solution in #543 instead

PR #541, opened 2020-02-14
Restart X server during DCV configuration

tilne:
The previous version of the DCV configuration recipe was using the
`set-default` unit file command in an attempt to stop the X server. This
sets the target to boot into, but does not stop the X server. This
commit uses the isolate unit command, which stops all units not enabled
in the specified unit (including the X server).

A kitchen test is also added to ensure that local users can access the X
server. This test is recommended in the [DCV admin guide](https://docs.aws.amazon.com/dcv/latest/adminguide/setting-up-installing-linux-checks.html#checks-xserver)
when the DCV GL package has been installed.

### DIFF
--- a/recipes/dcv_config.rb
+++ b/recipes/dcv_config.rb
@@ -22,7 +22,7 @@
 def allow_gpu_acceleration
   # Turn off X
   execute "Turn off X" do
-    command "systemctl set-default multi-user.target"
+    command "systemctl isolate multi-user.target"
   end
 
   # Update the xorg.conf to set up NVIDIA drivers.
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -220,6 +220,11 @@ if node['cfncluster']['cfn_node_type'] == "MasterServer" &&
     command 'dcv version'
     user node['cfncluster']['cfn_cluster_user']
   end
+  if graphic_instance?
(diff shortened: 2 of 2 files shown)


recipes/dcv_config.rb
```diff
@@ -22,7 +22,7 @@
 def allow_gpu_acceleration
   # Turn off X
   execute "Turn off X" do
-    command "systemctl set-default multi-user.target"
+    command "systemctl isolate multi-user.target"
```
rexcsn: I think it maybe best to include both `systemctl set-default multi-user.target` and `systemctl isolate multi-user.target` according to this guide: https://docs.aws.amazon.com/dcv/latest/adminguide/setting-up-installing-linux-prereq.html#rhel-7.x,-centos-7.x,-amazon-linux-2,-ubuntu-18.x,-and-suse-linux-enterprise-12.x

tilne: The only call to `set-default` I can find on that guide is the one that sets it to `graphical.target` which is done later in this function after DCV GL is installed.

rexcsn: By doing `systemctl isolate` on `multi-user.target` we are turning off X for `graphical.target`, which we turn on after setting up DCV GL later as Tim mentioned.

PR #543, opened 2020-02-19
Use single system pyenv installation

tilne:
Use a single system-level pyenv installation instead of multiple user
installations


recipes/tests.rb
```diff
@@ -29,6 +29,12 @@
   environment('PATH' => '/usr/local/bin:/usr/bin/:$PATH')
 end
 
+[node['cfncluster']['cookbook_virtualenv_path'], node['cfncluster']['node_virtualenv_path']].each do |venv_path|
```
lukeseawalker: can we add ['dcv']['authenticator']['virtualenv'] too?

PR #550, opened 2020-02-20
Start cloudwatch agent in _prep_env

rexcsn:
* Start cloudwatch agent at the start of chef run at run time
* This would avoid missing error during early installation stages that would happen before we initiate cloudwatch agent

### DIFF
--- a/recipes/base_config.rb
+++ b/recipes/base_config.rb
@@ -17,6 +17,9 @@
 
 include_recipe 'aws-parallelcluster::base_install'
 
+# Write cloudwatch log config and start it.
+include_recipe "aws-parallelcluster::cloudwatch_agent_config"
+
 if node['platform_family'] == 'amazon' && node['platform_version'] == '2'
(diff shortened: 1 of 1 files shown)


recipes/cloudwatch_agent_config.rb
```diff
@@ -14,6 +14,10 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+
+# Install cloudwatch agent if not already installed
+include_recipe 'aws-parallelcluster::cloudwatch_agent_install'
```
lukeseawalker: what I do not like here is that the cloudwatch_agent_install is re-executed at runtime when using a PCluster AMI, and some actions inside this recipe doesn't have a guard and are re-executed

PR #551, opened 2020-02-21
Disable pyenv rehash

lukeseawalker:
Since changing the contents of pyenv is not needed, disable the rehash to avoid getting the message "pyenv: cannot rehash: /usr/local/pyenv/shims isn't writable"

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -28,5 +28,6 @@ depends 'apt', '~> 7.0.0'
 depends 'hostname', '~> 0.4.2'
 depends 'line', '~> 2.4.1'
 depends 'ulimit', '~> 1.0.0'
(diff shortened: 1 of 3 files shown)


files/default/pyenv.sh
```diff
@@ -0,0 +1,12 @@
+# Prefer a user pyenv over a system wide install
+if [ -s "${HOME}/.pyenv/bin" ]; then
+    pyenv_root="${HOME}/.pyenv"
+elif [ -s "/usr/local/pyenv" ]; then
```
tilne: This is an attribute defined in defaults. We could make it a template and pass it in, but I also doubt if we'll ever change the system default root.

lukeseawalker: sure thing

PR #565, opened 2020-03-09
Check Packer Version >= 1.4.0

sean-smith:

### DIFF
--- a/amis/build_ami.sh
+++ b/amis/build_ami.sh
@@ -20,9 +20,15 @@
 #   build-date: timestamp to append to the AMIs names (optional)
 
 requirements_check() {
+    currentver="$(packer --version)"
+    requiredver="1.4.0"
     packer build --help >/dev/null 2>&1
     if [ $? -ne 0 ] ; then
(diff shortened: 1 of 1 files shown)


amis/build_ami.sh
```diff
@@ -20,9 +20,10 @@
 #   build-date: timestamp to append to the AMIs names (optional)
 
 requirements_check() {
-    packer build --help >/dev/null 2>&1
-    if [ $? -ne 0 ] ; then
-      echo "packer command not found. Is Packer installed?"
+    currentver="$(packer --version)"
```
lukeseawalker: We used "packer build" command because there is another "packer" tool out there, probably accepting the "--version" option

sean-smith: Good point. Let me check `packer build --version`

sean-smith: Ok, I've updated to do two checks:
1. Check `packer build` (this is to distinguish from the default cracklib packer binary present on RHEL based AMI's)
2. Check `packer --version` >= 1.4.0.

PR #567, opened 2020-03-15
Change Slurm process tracking config to use cgroup plugin

rexcsn:
* Change ProctrackType from proctrack/gpid to proctrack/cgroup in slurm.conf
* Using proctrack/gpid could leave stray processes when running MPI applications
* Added cgroup configuration file to enable cgroup plugin


attributes/conditions.rb
```diff
@@ -22,3 +22,8 @@
   || node['platform'] == 'amazon' || node['platform'] == 'ubuntu'
 
 default['conditions']['ami_bootstrapped'] = ami_bootstrapped?
+
+# alinux1 and centos6 use an old cgroup directory: /cgroup, all other OSs use /sys/fs/cgroup, which is the default
+# "platform_verision" on alinux1 is "2018.03" and "platform_verision" on alinux2 is "2"
+default['conditions']['old_cgroup_directory'] = (node['platform'] == 'centos' && node['platform_version'].to_i == 6) \
+  || (node['platform'] == 'amazon' && node['platform_version'] != "2")
```
demartinofra: can we match alinux1 here `node['platform_version'] != "2"` so that it is forward compatible?

templates/default/slurm.conf.erb
```diff
@@ -27,7 +27,7 @@ SwitchType=switch/none
 MpiDefault=none
 SlurmctldPidFile=/var/run/slurmctld.pid
 SlurmdPidFile=/var/run/slurmd.pid
-ProctrackType=proctrack/pgid
+ProctrackType=proctrack/cgroup
```
demartinofra: not sure how much it helps but we could have a kitchen test verifying at least that cgroups are available. you could run `cat /proc/mounts` and grep for cgroups?

also support for cgroups becomes a new requirement for customami creation. Are we tracking this anywhere in our docs?

attributes/conditions.rb
```diff
@@ -22,3 +22,8 @@
   || node['platform'] == 'amazon' || node['platform'] == 'ubuntu'
 
 default['conditions']['ami_bootstrapped'] = ami_bootstrapped?
+
+# alinux1 and centos6 use an old cgroup directory: /cgroup, all other OSs use /sys/fs/cgroup, which is the default
+# "platform_verision" on alinux1 is "2018.03" and "platform_verision" on alinux2 is "2"
```
enrico-usai: minor: double typo in the comment: "verision" vs "version"

templates/default/cgroup.conf.erb
```diff
@@ -0,0 +1,7 @@
+###
+# Slurm cgroup support configuration file
+###
+CgroupAutomount=yes
+<% if node['conditions']['old_cgroup_directory'] -%>
```
lukeseawalker: use https://docs.chef.io/resources/template/#template_specificity instead of condition based on platform

templates/default/slurm.conf.erb
lukeseawalker: if within the requirements there is a package to be installed, we must add it to the list of package we install in the AMI

rexcsn: Don't think we can do `cat /proc/mounts` and grep for cgroups. `alinux1` and `centos6` originally come with no `cgroup` mount points, `slurmd` actually mounts the `cgroup/freezer` directory for these 2 OSs.

For kitchen tests on compute we are not really starting `slurm`, so I would not expect `cgroup/freezer` to be there.

I agree that we need to add to our documentation to make sure users are aware that their kernel need to be compatible with `cgroup`, but I think the the majority of the OSs out there should already be compatible.
`cgroup` should come with any OS using kernel version after Linux 2.6.24, which was released 24 January 2008.

templates/amazon-2018.03/cgroup.conf.erb
```diff
@@ -0,0 +1,5 @@
+###
```
lukeseawalker: can we be more generic in the directory name, e.g. amazon-2018?

templates/centos-6.9/cgroup.conf.erb
```diff
@@ -0,0 +1,5 @@
+###
```
lukeseawalker: can we be more generic in the directory name, e.g. centos-6?

templates/amazon-2018.03/cgroup.conf.erb
rexcsn: Yes, and for future reference, using only the major version will work for both template and cookbook file resources

PR #576, opened 2020-04-07
Separate Ganglia installation from configuration

lukeseawalker:
ganglia is now managed by two recipes, one for the installation and one for the configuration.
The installation is executed at AMI build, while the configuration is executed at cluster creation time if ganglia is enabled through the ['ganglia_enabled'] variable.

This solves a regression introduced by https://github.com/aws/aws-parallelcluster-cookbook/pull/563, where the ganglia installation recipe wasn't executed at runtime.

*Issue #, if available:*

*Description of changes:*


recipes/ganglia_install.rb
```diff
@@ -0,0 +1,45 @@
...
+case node['cfncluster']['cfn_node_type']
+when 'MasterServer', nil
+  case node['platform']
+  when "redhat", "centos", "amazon", "scientific" # ~FC024
+    package %w[ganglia ganglia-gmond ganglia-gmetad ganglia-web httpd php php-gd rrdtool] do
+      retries 3
+      retry_delay 5
+    end
+  when "ubuntu"
+    package %w[ganglia-monitor rrdtool gmetad ganglia-webfrontend] do
```
ddeidda, line 27: Good catch. web frontend was being installed in compute nodes too :+1:

recipes/ganglia_install.rb
```diff
@@ -0,0 +1,45 @@
...
+      retries 3
+      retry_delay 5
+    end
+  end
+when 'ComputeFleet'
+  case node['platform']
+  when "redhat", "centos", "amazon", "scientific" # ~FC024
+    package %w[ganglia-gmond] do
```
ddeidda, line 35: This is needed because it was previously installed alongside ganglia-web. Just taking note for my review.

recipes/ganglia_install.rb
```diff
@@ -0,0 +1,45 @@
...
+      retries 3
+      retry_delay 5
+    end
+  when "ubuntu"
+    package %w[ganglia-monitor] do
```
ddeidda, line 40: Same as before, but only for Ubuntu. OK

PR #578, opened 2020-04-07
Omit cfn-init-cmd.log and cfn-wire.log from CW Logs integration

tilne:
In an effort to reduce the cost of enabling the CloudWatch Logs
integration, two of the cfn-* logs are being removed from the config
file used to ultimately instruct the CloudWatch agent which files to
log data for on a cluster's instances.

### DIFF
--- a/files/default/cloudwatch_logs/cloudwatch_log_files.json
+++ b/files/default/cloudwatch_logs/cloudwatch_log_files.json
@@ -50,50 +50,8 @@
     },
     {
       "timestamp_format_key": "default",
-      "file_path": "/var/log/cfn-init-cmd.log",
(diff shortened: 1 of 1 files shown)


files/default/cloudwatch_logs/cloudwatch_log_files.json
```diff
@@ -48,27 +48,6 @@
       ],
       "feature_conditions": []
     },
-    {
-      "timestamp_format_key": "default",
-      "file_path": "/var/log/cfn-init-cmd.log",
```
rexcsn, line 53: Might need to revisit in the future to see which `cfn-init` log to keep if we were to add timestamps to track individual recipe's execution time. Ok for now.

PR #581, opened 2020-04-10
Upgrade Slurm to version 20.02

ddeidda:
Upgrade Slurm to version 20.02


recipes/slurm_install.rb
```diff
@@ -60,6 +60,8 @@
       IFS=${_OLD_IFS}
       export PATH=${NEW_PATH}
 
+      source #{node['cfncluster']['cookbook_virtualenv_path']}/bin/activate
```
demartinofra, line 63: shall we just leave a comment to explain why we need to do this?

PR #585, opened 2020-04-18
Handling EC2 Health Scheduled Events

rexcsn:
* Added health queue info to `/etc/sqswatcher.cfg` so that nodewatcher can use this info


templates/default/cfnconfig.erb
```diff
@@ -15,6 +15,7 @@ cfn_node_type=<%= node['cfncluster']['cfn_node_type'] %>
 cfn_cluster_user=<%= node['cfncluster']['cfn_cluster_user'] %>
 <% if node['cfncluster']['cfn_node_type'] == 'ComputeFleet' -%>
 cfn_sqs_queue=<%= node['cfncluster']['cfn_sqs_queue'] %>
+cfn_health_sqs_queue=<%= node['cfncluster']['cfn_health_sqs_queue'] %>
```
demartinofra, line 18: is this needed? I think the other queue is here to send the compute_ready event

templates/default/sqswatcher.cfg.erb
```diff
@@ -1,6 +1,7 @@
 [sqswatcher]
 region = <%= node['cfncluster']['cfn_region'] %>
 sqsqueue = <%= node['cfncluster']['cfn_sqs_queue'] %>
+healthsqsqueue = <%= node['cfncluster']['cfn_health_sqs_queue'] %>
```
demartinofra, line 4: just because I'm picky :D we could rename this to `health_sqsqueue` just to use the same naming convention

templates/default/cfnconfig.erb
rexcsn: I thought this file need to be consistent with `attributes/default.rb` base on [this comment](https://github.com/aws/aws-parallelcluster-cookbook/blob/v2.6.1/attributes/default.rb#L266)

templates/default/sqswatcher.cfg.erb
rexcsn: Done

templates/default/cfnconfig.erb
demartinofra: I don't think it's needed but I'm ok with leaving it there

PR #587, opened 2020-04-20
Fix installation of Lustre client on Centos 7.7

demartinofra:
The newer clients for Centos 7.8 are already available in the same repo.
Without pinning the kernel version we are going to download the latest ones that are not compatible.
Also this should be compatible with Centos 7.8 kernel as soon as it gets released.

### DIFF
--- a/recipes/_lustre_install.rb
+++ b/recipes/_lustre_install.rb
@@ -49,7 +49,7 @@ if node['platform'] == 'centos' && (5..6).cover?(node['platform_version'].split(
   end
 
   kernel_module 'lnet'
-elsif node['platform'] == 'centos' && node['platform_version'].split('.')[1].to_i == 7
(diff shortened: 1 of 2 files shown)


recipes/_lustre_install.rb
```diff
@@ -49,7 +49,7 @@
   end
 
   kernel_module 'lnet'
-elsif node['platform'] == 'centos' && node['platform_version'].split('.')[1].to_i == 7
+elsif node['platform'] == 'centos' && node['platform_version'].split('.')[1].to_i >= 7
```
lukeseawalker, line 52: minor - Please change the log below
"Chef::Log.warn("Unsupported version of Centos, #{node['platform_version']}, supported versions are 7.5, 7.6 and 7.7")"

demartinofra: done

PR #589, opened 2020-04-24
Generate slurm config files

rexcsn:


files/default/slurm/generate_slurm_config.py
```diff
@@ -0,0 +1,104 @@
+import json
+from os import makedirs, path
+
+import argparse
+
+from jinja2 import Environment, FileSystemLoader
+from jsonschema import validate
```
demartinofra, line 7: although jsonschema is useful here I would drop all non-necessary dependencies on third party libraries.

files/default/slurm/templates/slurm_parallelcluster_partitions.conf
```diff
@@ -0,0 +1,9 @@
+# slurm_parallelcluster_partitions.conf is managed by the pcluster processes.
```
demartinofra, line 1: wondering if we should rename this file to `slurm_parallelcluster.conf` or something that is not partition specific since we decided to include additional config options in it.

recipes/_master_slurm_config.rb
```diff
@@ -43,25 +43,25 @@
...
+  source 'slurm'
+  mode '0700'
+  action :create
+  recursive true
+end
+
+# TO-DO: verify directories and input file path
+# Generate pcluster specific configs
+execute "generate_pcluster_slurm_configs" do
+  command "python #{node['cfncluster']['scripts_dir']}/generator_slurm_config.py --output-directory /opt/slurm/etc/ --template-directory #{node['cfncluster']['scripts_dir']}/slurm/templates/ --input-file #{node['cfncluster']['scripts_dir']}/slurm/templates/sample_input.json"
```
demartinofra, line 64: we need to use the python from the cookbook virtualenv

templates/default/slurm.conf.erb
```diff
@@ -8,50 +8,29 @@
...
-#JobCredentialPublicCertificate=
 StateSaveLocation=/var/spool/slurm.state
 SlurmdSpoolDir=/var/spool/slurmd
 SwitchType=switch/none
-MpiDefault=none
 SlurmctldPidFile=/var/run/slurmctld.pid
 SlurmdPidFile=/var/run/slurmd.pid
-ProctrackType=proctrack/cgroup
-#PluginDir=
+SlurmctldParameters=cloud_dns,idle_on_node_suspend
```
demartinofra, line 22: I would still keep SlurmctldParameters and CommunicationParameters under `CLOUD CONFIGS OPTIONS` since they are very specific to that and also mentioned in the Cloud Scheduling Guide: https://slurm.schedmd.com/elastic_computing.html

files/default/slurm/templates/slurm_parallelcluster_queue_gres.conf
```diff
@@ -0,0 +1,14 @@
+# This file is automatically generated by pcluster
+
+{% for instance in queue.instances %}
+{%- if instance.gpus > 0 -%}
+{%- set has_static_nodes = true if instance.static_size != 0 else false -%}
+{%- set has_dynamic_nodes = true if instance.dynamic_size > 0 else false -%}
+{% if has_static_nodes -%}
+NodeName={{ queue.queue_name }}-static-{{ instance.type }}[1-{{ instance.static_size }}] Name=gpu Type=tesla File=/dev/nvidia[0-{{ instance.gpus - 1 }}]
```
demartinofra, line 8: let's add a `-` after instance.type to facilitate parsing

files/default/slurm/templates/slurm_parallelcluster.conf
```diff
@@ -0,0 +1,27 @@
...
+# Do not modify.
+# Please add user-specific slurm configuration options in slurm.conf
+{% set has_static_nodes = namespace(has_static=false) -%}
+
+ControlMachine={{ master_hostname }}
+SuspendTime={{ scaling_config.scaledown_idletime }}
+
+{% for queue in queues_config %}
+include {{ queues_config[queue].queue_partition_filename }}
+{%- for instance in queues_config[queue].instances -%}
```
demartinofra, line 11: it's a common practice to indent jinja directives

files/default/slurm/templates/slurm_parallelcluster_queue_partition.conf
```diff
@@ -0,0 +1,30 @@
...
+{{ queue.queue_name }}-static-{{ instance.type }}[1-{{ instance.static_size }}],{{ queue.queue_name }}-dynamic-{{ instance.type }}[1-{{ instance.dynamic_size }}]
+{%- elif has_static_nodes -%}
+{{ queue.queue_name }}-static-{{ instance.type }}[1-{{ instance.static_size }}]
+{%- elif has_dynamic_nodes -%}
+{{ queue.queue_name }}-dynamic-{{ instance.type }}[1-{{ instance.dynamic_size }}]
+{%- endif -%}
+{{- "," if not loop.last else "" -}}
+{%- endfor %}
+
+PartitionName={{ queue.queue_name }} Nodes={{ queue.queue_name }}_nodes MaxTime=INFINITE State=UP
```
demartinofra, line 29: we probably want to set the first queue/partition as default

templates/default/slurm.conf.erb
```diff
@@ -88,22 +59,9 @@ SlurmctldLogFile=/var/log/slurmctld.log
 SlurmdDebug=3
```
demartinofra, line 59: let's update the debug config options to use the proper values (I believe setting the log level as a number comes from very old versions of slurm)

files/default/slurm/templates/slurm_parallelcluster.conf
```diff
@@ -0,0 +1,27 @@
+# slurm_parallelcluster_partitions.conf is managed by the pcluster processes.
+# Do not modify.
+# Please add user-specific slurm configuration options in slurm.conf
+{% set has_static_nodes = namespace(has_static=false) -%}
+
+ControlMachine={{ master_hostname }}
+SuspendTime={{ scaling_config.scaledown_idletime }}
```
demartinofra, line 7: SuspendTime is in seconds, scaledown_idletime in minutes.

files/default/slurm/templates/sample_input.json
```diff
@@ -0,0 +1,52 @@
...
+            "compute_type":"spot"
+        },
+        "queue2":{
+            "queue_name" : "Queue2",
+            "instances":[
+                {
+                    "type":"g3.8xlarge",
+                    "static_size":0,
+                    "dynamic_size":2,
+                    "vcpus":32,
```
demartinofra, line 41: we need to take into account disable_hyperthreading

files/default/slurm/templates/slurm_parallelcluster_queue_partition.conf
rexcsn: Setting default in `slurm_parallelcluster.conf`

recipes/_master_slurm_config.rb
```diff
@@ -43,25 +43,38 @@
...
+
+# Copy queue config file from S3 URI
+execute "copy_queue_config_from_s3" do
+  command "aws s3 cp #{node['cfncluster']['slurm']['queue_config_s3_uri']} #{node['cfncluster']['slurm']['queue_config_path']}"
+end
+
+# TO-DO: verify directories and input file path
+# Generate pcluster specific configs
+execute "generate_pcluster_slurm_configs" do
+  command "/usr/local/pyenv/versions/cookbook_virtualenv/bin/python #{node['cfncluster']['scripts_dir']}/pcluster_slurm_config_generator.py --output-directory /opt/slurm/etc/ --template-directory #{node['cfncluster']['scripts_dir']}/slurm/templates/ --input-file #{node['cfncluster']['slurm']['queue_config_path']}"
```
demartinofra, line 77: cookbook_virtualenv path from variable

recipes/_master_slurm_config.rb
```diff
@@ -43,25 +43,38 @@
...
-  mode '0644'
+  mode '0755'
+end
+
+# Copy queue config file from S3 URI
+execute "copy_queue_config_from_s3" do
+  command "aws s3 cp #{node['cfncluster']['slurm']['queue_config_s3_uri']} #{node['cfncluster']['slurm']['queue_config_path']}"
+end
+
+# TO-DO: verify directories and input file path
```
demartinofra, line 74: is this still a valid todo?

recipes/_master_slurm_config.rb
```diff
@@ -87,6 +100,20 @@
   mode '0755'
 end
 
+cookbook_file '/home/slurm/slurm_resume' do
```
demartinofra, line 103: we should move this to a more appropriate directory. Also we might want to use 0700 as mode?

templates/default/slurm.conf.erb
```diff
@@ -8,50 +8,29 @@
 #
```
demartinofra, line 8: I'm wondering if this can be moved to the files folder now

templates/default/slurm.conf.erb
```diff
@@ -62,48 +41,27 @@ MinJobAge=300
...
-#JobAcctGatherType=jobacct_gather/linux
-#JobAcctGatherFrequency=30
-#
-#AccountingStorageType=accounting_storage/slurmdbd
-#AccountingStorageHost=
-#AccountingStorageLoc=
-#AccountingStoragePass=
-#AccountingStorageUser=
 #
 # WARNING!!! The Nodename and PartionName entries below
```
demartinofra, line 63: requires to be updated or removed

files/default/slurm/pcluster_slurm_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+import logging
+from logging.handlers import RotatingFileHandler
+from os import makedirs, path
+from socket import gethostname
+
+import argparse
+
+from jinja2 import Environment, FileSystemLoader
+
+LOG_FILE_PATH = "/var/log/parallelcluster/pcluster_slurm_config_generator.log"
```
demartinofra, line 12: do we want to have a log file specific for this script or simply log to stdout and let the caller have it logged? cfn-init already logs stdout by default if I'm not wrong.

files/default/slurm/pcluster_slurm_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+    logfile_handler.setFormatter(formatter)
+
+    logger = logging.getLogger("pcluster_slurm_config_generator")
+    logger.addHandler(logfile_handler)
+
+    logger.setLevel("INFO")
+    return logger
+
+
+if __name__ == "__main__":
```
demartinofra, line 139: nit: let's move whatever is under `if __name__ == "__main__":` to a `main()` function

files/default/slurm/pcluster_slurm_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+    makedirs(pcluster_subdirectory, exist_ok=True)
+    env = _get_jinja_env(args.template_directory)
+
+    # Generate slurm_parallelcluster_{QueueName}_partitions.conf and slurm_parallelcluster_{QueueName}_gres.conf
+    for queue in queues_info["queues_config"]:
+        for file_type in ["partition", "gres"]:
+            logger.info("Generating slurm_parallelcluster_{}_{}.conf".format(queue, file_type))
+            rendered_template = _render_queue_configs(queues_info, queue, file_type, env, pcluster_subdirectory)
+            if not args.dryrun:
+                filename = queues_info["queues_config"][queue]["queue_{}_filename".format(file_type)]
```
demartinofra, line 42: optional: since we are using python 3.6 to run this scripts we could use f-strings rather than format in case you find it more readable

attributes/default.rb
```diff
@@ -56,6 +57,9 @@
 default['cfncluster']['slurm']['version'] = '20.02.1'
 default['cfncluster']['slurm']['url'] = 'https://download.schedmd.com/slurm/slurm-20.02.1.tar.bz2'
 default['cfncluster']['slurm']['sha1'] = '91704d8c73bf1a6eab64ccde3f3d03b85188672a'
+# ParallelCluster slurm variables
+default['cfncluster']['slurm']['queue_config_s3_uri'] = nil
+default['cfncluster']['slurm']['queue_config_path'] = "#{node['cfncluster']['configs_dir']}/slurm_config.json"
```
demartinofra, line 62: nit: in the end this is not a slurm_config. Shall we find a more generic name for it? `queues_config` or something else that does not have slurm in it?

files/default/slurm/templates/slurm_parallelcluster_queue_partition.conf
```diff
@@ -0,0 +1,32 @@
+# This file is automatically generated by pcluster
+{% set has_node = namespace(has_node=false) %}
```
demartinofra, line 2: is a namespace required to set a variable from a different scope or a normal variable is enough here?
We might want to change the name of the namespace if we need to keep it to something like ns so that we don't have something like `has_node.has_node`

recipes/_master_slurm_config.rb
```diff
@@ -43,25 +43,38 @@
...
+  source 'slurm/pcluster_slurm_config_generator.py'
   owner 'root'
   group 'root'
-  mode '0644'
+  mode '0755'
+end
+
+# Copy queue config file from S3 URI
+execute "copy_queue_config_from_s3" do
+  command "aws s3 cp #{node['cfncluster']['slurm']['queue_config_s3_uri']} #{node['cfncluster']['slurm']['queue_config_path']}"
```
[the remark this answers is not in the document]
demartinofra: now that I think more about the proxy, we already export the proxy variables in the userdata, which makes me thing they are available when executing this piece of code. Can you confirm that and in the case I'd say we drop the proxy variables

test/unit/slurm/test_slurm_config_generator.py
```diff
@@ -0,0 +1,97 @@
...
+from assertpy import assert_that
+from jsonschema import validate
+from slurm.pcluster_slurm_config_generator import (
+    _get_jinja_env,
+    _load_queues_info,
+    _render_queue_configs,
+    _render_slurm_parallelcluster_configs,
+)
+
+INPUT_SCHEMA = {
```
demartinofra, line 13: we'll need to update this

PR #596, opened 2020-05-06
Support for arm instances

tilne:

### DIFF
--- a/files/ubuntu-16.04/openmpi-modulefile-via-alternatives
+++ b/files/ubuntu-16.04/openmpi-modulefile-via-alternatives
@@ -1,3 +1,7 @@
 #%Module
+
+# Call update-alternatives when this module is loaded to ensure that the
+# system is configured to use open MPI binaries and libraries by default
+
 exec sudo update-alternatives --set mpi /usr/lib/openmpi/include
 exec sudo update-alternatives --set mpirun /usr/bin/mpirun.openmpi
--- a/files/ubuntu-16.04/openmpi-modulefile-via-alternatives
+++ b/files/ubuntu-16.04/openmpi-modulefile-via-alternatives
@@ -1,0 +1,3 @@
+#%Module
+exec sudo update-alternatives --set mpi /usr/lib/openmpi/include
+exec sudo update-alternatives --set mpirun /usr/bin/mpirun.openmpi
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -160,3 +160,19 @@ end
 def arm_instance?
   node['kernel']['machine'] == 'aarch64'
 end
+
+#
+# Check if this is an OS on which EFA is supported
+#
+def platform_supports_efa?
+  [node['platform'] == 'centos' && node['platform_version'].to_i >= 7,
+   node['platform'] == 'amazon',
+   node['platform'] == 'ubuntu'].any?
+end
+
+#
+# Check if this is an architecture on which EFA is supported
+#
+def arch_supports_efa?
+  !arm_instance?
+end
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -171,8 +171,17 @@ def platform_supports_efa?
 end
 
 #
-# Check if this is an architecture on which EFA is supported
+# Check if the platform supports intel MPI
 #
-def arch_supports_efa?
-  !arm_instance?
+def platform_supports_impi?
+  [node['platform'] == 'centos' && node['platform_version'].to_i >= 7,
+   node['platform'] == 'amazon',
+   node['platform'] == 'ubuntu'].any?
+end
+
+#
+# Check if this platform supports intel's HPC platform
+#
+def platform_supports_intel_hpc_platform?
+  node['platform'] == 'centos' && node['platform_version'].to_i >= 7
 end
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -185,3 +185,10 @@ end
 def platform_supports_intel_hpc_platform?
   node['platform'] == 'centos' && node['platform_version'].to_i >= 7
 end
+
+#
+# Check if DCV is supported on this OS
+#
+def platform_supports_dcv?
+  node['cfncluster']['dcv']['supported_os'].include?("#{node['platform']}#{node['platform_version'].to_i}")
+end
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -192,3 +192,12 @@ end
 def platform_supports_dcv?
   node['cfncluster']['dcv']['supported_os'].include?("#{node['platform']}#{node['platform_version'].to_i}")
 end
+
+#
+# Check if Lustre is supported on this OS
+#
+def platform_supports_lustre?
+  [node['platform'] == 'centos' && node['platform_version'].to_i >= 7,
+   node['platform'] == 'amazon',
+   node['platform'] == 'ubuntu'].any?
+end
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -153,3 +153,10 @@ def ami_bootstrapped?
 
   'aws-parallelcluster-' + node['cfncluster']['cfncluster-version'] == version && node['cfncluster']['skip_install_recipes'] == 'yes'
 end
+
+#
+# Check if this is an ARM instance
+#
+def arm_instance?
+  node['kernel']['machine'] == 'aarch64'
+end
--- a/recipes/arm_openmpi_install.rb
+++ b/recipes/arm_openmpi_install.rb
@@ -15,15 +15,13 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-return if node['conditions']['ami_bootstrapped'] || !arm_instance?
-
 openmpi_modules_dir = "#{node['cfncluster']['moduleshome']}/modulefiles/openmpi"
 
 case node['platform_family']
 when 'amazon'
   if node['platform_version'].to_i == 2
     package 'openmpi-devel' do
-      retries 10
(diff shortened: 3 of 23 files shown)


tilne:
changing base branch to feature branch and marking this ready for review

libraries/helpers.rb
```diff
@@ -153,3 +153,10 @@ def ami_bootstrapped?
 
   'aws-parallelcluster-' + node['cfncluster']['cfncluster-version'] == version && node['cfncluster']['skip_install_recipes'] == 'yes'
 end
+
+#
+# Check if this is an ARM instance
+#
+def arm_instance?
+  node['kernel']['machine'] == 'aarch64'
```
lukeseawalker, line 161: is this value common across all the supported platforms (e.g. for Ubuntu)?

recipes/cloudwatch_agent_install.rb
```diff
@@ -30,18 +30,17 @@
...
-  package_path = "#{node['cfncluster']['sources_dir']}/amazon-cloudwatch-agent.deb"
-when 'amazon'
-  # NOTE: the URL used to get amazon linux's package does not use node['platform']
-  package_url = "#{package_url_prefix}/amazon_linux/amd64/latest/amazon-cloudwatch-agent.rpm"
-  package_path = "#{node['cfncluster']['sources_dir']}/amazon-cloudwatch-agent.rpm"
-when 'centos'
-  package_url = "#{package_url_prefix}/#{node['platform']}/amd64/latest/amazon-cloudwatch-agent.rpm"
-  package_path = "#{node['cfncluster']['sources_dir']}/amazon-cloudwatch-agent.rpm"
-end
+arch_url_component = arm_instance? ? 'arm64' : 'amd64' # TODO: helper function for detecting ARM
```
lukeseawalker, line 33: remove #TODO

recipes/cloudwatch_agent_install.rb
```diff
@@ -30,18 +30,17 @@
...
+platform_url_component = node['platform'] == 'amazon' ? 'amazon_linux' : node['platform']
+package_suffix = node['platform'] == 'ubuntu' ? 'deb' : 'rpm'
```
lukeseawalker, line 35: - minor - what about package_extension?

recipes/dcv_install.rb
```diff
@@ -70,7 +70,7 @@ def disable_lock_screen
   not_if { ::File.exist?("#{node['cfncluster']['scripts_dir']}/pcluster_dcv_connect.sh") }
```
lukeseawalker, line 70: with the.new logic I think this cookbook_file resource is skipped, but afaik it was intended to be there even for not supported OS, like comment says "# Install pcluster_dcv_connect.sh script in all the OSes to use it for error handling"

recipes/arm_openmpi_install.rb
```diff
@@ -0,0 +1,46 @@
...
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+return if node['conditions']['ami_bootstrapped'] || !arm_instance?
```
lukeseawalker, line 18: if there are no other place from where `arm_openmpi_install` is included, then `if node['conditions']['ami_bootstrapped']` can be removed, because it's already included in the caller

recipes/arm_openmpi_install.rb
```diff
@@ -0,0 +1,46 @@
...
+
+case node['platform_family']
+when 'amazon'
+  if node['platform_version'].to_i == 2
+    package 'openmpi-devel' do
+      retries 10
```
lukeseawalker, line 26: why 10 retries?

recipes/arm_openmpi_install.rb
```diff
@@ -0,0 +1,46 @@
...
+    append_if_no_line "append mpi modulefiles line" do
+      path "#{node['cfncluster']['moduleshome']}/init/.modulespath"
+      line "/etc/modulefiles/mpi"
+    end
+  else
+    Chef::Log.Info("ARM instances are currently only supported for Amazon Linux 2")
+  end
+when 'debian'
+  package %w[libopenmpi-dev openmpi-bin openmpi-doc] do
+    retries 10
```
lukeseawalker, line 47: same here, why 10? I think in other place we are using 3

files/ubuntu-16.04/openmpi-modulefile-via-alternatives
```diff
@@ -0,0 +1,3 @@
+#%Module
```
lukeseawalker, line 1: can you add more comment (even in the commit log is ok) on how this works?
Would the `update-alternatives` command. be called every time the module is loaded?

recipes/tests.rb
```diff
@@ -242,51 +242,53 @@
...
-  execute 'check intel mpi installed' do
-    command "rpm -qa | grep intel-mpi"
-    user node['cfncluster']['cfn_cluster_user']
-  end
-when 'ubuntu1604', 'ubuntu1804'
-  case node['cfncluster']['cfn_node_type']
-  when 'MasterServer'
-    execute 'check ptrace protection enabled' do
-      command "sysctl kernel.yama.ptrace_scope | grep 'kernel.yama.ptrace_scope = 1'"
+if node['conditions']['efa_supported'] && node['conditions']['intel_mpi_supported']
```
lukeseawalker, line 251: I can be wrong but I don't think that `node['conditions']` is not available here, please correct me

recipes/tests.rb
```diff
@@ -90,23 +90,29 @@
 # SGE
 ###################
 if node['cfncluster']['cfn_scheduler'] == 'sge'
+  sge_bin_suffix = if arm_instance?
+                     "arm64"
+                   else
+                     "amd64"
+                   end
+  sge_bin_paths = "/opt/sge/bin:/opt/sge/bin/lx-#{sge_bin_suffix}"
```
lukeseawalker, line 98: not related to this line, but in the sge_install recipe there is a guard on `creates '/opt/sge/bin/lx-amd64/sge_qmaster'`. I think that should be managed as well otherwise the installation will be repeated at runtime too

libraries/helpers.rb
tilne: It's worked so far in my testing.

recipes/cloudwatch_agent_install.rb
tilne: Done.

recipes/cloudwatch_agent_install.rb
tilne: Done.

recipes/dcv_install.rb
tilne: Wouldn't it only be skipped if it already exists?

The guards at the top of the file only prevent the recipe from running again if the AMI's bootstrapped or if /etc/dcv/dcv.conf exists (which I believe implies this has already run anyway).

The recipe itself is source in two recipes:
* from dcv_config, 
* from default

The include in default doesn't have any guards around it, so that shouldn't be a problem. I believe this is the entrypoint to the cookbook for both the AMI build process and for runtime AMI baking.

The only process where this would be an issue would be the kitchen tests. Does it matter that it's not being included there? If DCV isn't supported on the platform it's going to skip all of those tests anyways.

recipes/arm_openmpi_install.rb
tilne: Makes sense.

recipes/arm_openmpi_install.rb
tilne: lol i don't have a good answer

tilne: we have 10 retries for package installations randomly scattered throughout the recipes. agreed it's definitely not needed for these.

files/ubuntu-16.04/openmpi-modulefile-via-alternatives
tilne: Yes. I'll add a comment to each of the files. Let me know if you think it needs more information.

recipes/tests.rb
tilne: This is counter to what I've seen. Why would it not be available?

recipes/tests.rb
tilne: good catch i very much missed that one.

recipes/arm_openmpi_install.rb
lukeseawalker: `retries 10` is used when a list is passed to the package manager
`retries 3` is used when a single package is required to be installed

recipes/dcv_install.rb
lukeseawalker: ok you are right. Probably the check on `/etc/dcv/dcv.conf"` could be even removed since there is the one on the `ami_bootstrapped `.

I didn't get why this would be an issue in the kitchen tests

tilne: It would be an issue during kitchen tests because the include path for that scenario is through dcv_config, which would prevent this recipe from running because it checks conditions.

I don't think that would be an issue since I assume none of the kitchen tests are exercising that script.

recipes/arm_openmpi_install.rb
tilne: ah i see.

recipes/arm_openmpi_install.rb
tilne: so along the lines of the previous answer, this is a list passed to the package manager. Use  10 retries here then for the sake of consistency?

My plan is to keep it at 3 retries due to the fact that it's a small list, but i see that dcv_install does something differently.

PR #598, opened 2020-05-07
Pin cfn-bootstrap on Centos 6

ddeidda:
This commit pins cfn-bootstrap to 1.4-31 to avoid python errors with cfn-bootstrap on centos 6.


tilne:
Where are the versions documented? The only documentation I could find is here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/releasehistory-aws-cfn-bootstrap.html. The latest version is 1.4-30.

ddeidda:
> Where are the versions documented? The only documentation I could find is here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/releasehistory-aws-cfn-bootstrap.html. The latest version is 1.4-30.

I found them here: https://s3.amazonaws.com/cloudformation-examples

ddeidda:
Side note: the fix won't work for AMIs baked at runtime since in that case cfn-bootstrap is downloaded from the URL specified in the UserData seciton of the cfn template.

PR #600, opened 2020-05-11
Download Chef client from our s3 buckets

ddeidda:
This patch moves the download of Chef client from Chef website to our s3
buckets.

### DIFF
--- a/amis/packer_alinux.json
+++ b/amis/packer_alinux.json
@@ -173,7 +173,8 @@
     {
       "type" : "shell",
       "inline" : [
-        "curl --retry 3 -L https://www.chef.io/chef/install.sh | sudo bash -s -- -v {{user `chef_version`}}"
+        "region=\"{{user `region`}}\"",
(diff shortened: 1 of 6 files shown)


amis/packer_alinux.json
```diff
@@ -173,7 +173,7 @@
     {
       "type" : "shell",
       "inline" : [
-        "curl --retry 3 -L https://www.chef.io/chef/install.sh | sudo bash -s -- -v {{user `chef_version`}}"
+        "curl --retry 3 -L https://{{user `region`}}-aws-parallelcluster.s3.{{user `region`}}.amazonaws.com$([ \"$_region\" != \"${_region#cn-*}\" ] && echo \".cn\")/archives/chef/chef-install.sh | sudo bash -s -- -v {{user `chef_version`}}"
```
lukeseawalker, line 176: $_region is not set anywhere

amis/packer_alinux.json
```diff
@@ -173,7 +173,8 @@
     {
       "type" : "shell",
       "inline" : [
-        "curl --retry 3 -L https://www.chef.io/chef/install.sh | sudo bash -s -- -v {{user `chef_version`}}"
+        "_region={{user `region`}}",
```
lukeseawalker, line 176: two things:
* quote assignment, e.g. `"_region=\"{{user `region`}}\"",`
* I would use `region` instead of `_region`, to be consistent with other places in this same file where `region` is defined as well

PR #604, opened 2020-05-12
Workaround for MODULESHOME

rexcsn:
* Hardcoding MODULESHOME as temporary workaround

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -147,6 +147,8 @@ when 'rhel', 'amazon'
 
   # Modulefile Directory
   default['cfncluster']['modulefile_dir'] = "/usr/share/Modules/modulefiles"
+  # MODULESHOME
+  default['cfncluster']['moduleshome'] = "/usr/share/Modules"
 
   case node['platform']
   when 'centos', 'redhat', 'scientific' # ~FC024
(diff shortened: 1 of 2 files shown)


rexcsn:
> * Can you drop a comment that links to the PR that made the change that uses the environment variable, and explain why this one is needed given those changes.

The dependency on `MODULESHOME` env variable was added as part of [this commit](https://github.com/aws/aws-parallelcluster-cookbook/pull/602/files#diff-a61d00bd12cc4893ec1686a978419932R47). The idea is we want to append the path for the IntelMPI module file to `$MODULESHOME/init/.modulespath` (something like `/usr/share/modules/init/.modulespath`).
However, related to [this chef issue](https://github.com/chef-cookbooks/chef-client/issues/473), `MODULESHOME` env variable cannot be resolved by chef, and chef tries to append to `/init/.modulespath` causing an error when executing this recipe.

> * What does the longer-term fix for this look like, if it is indeed an issue with chef not being able to see the environment properly?

Given that we are hardcoding things like `default['cfncluster']['modulefile_dir']` for different platforms, I think hardcoding `default['cfncluster']['moduleshome']` is aligned with current practice. If we want to completely solve the problem with chef environment variables, we might want to look into ways to set PATH for chef client, so that chef client can access all the same environment variables as normal users.

PR #608, opened 2020-05-14
Purge pending packages before installing new ones

tilne:
`apt-get install` was failing when provisioning an AMI that had
pending, uninstalled packages. Avoid this by purging pending packages
prior to installing new ones.


tilne:
closing this because we dealt with the bad AMI rather than always guarding against incomplete installations.

PR #612, opened 2020-05-14
Fix Kernel pin in Ubuntu18

ddeidda:
- Linux headers package was not installed
- Missing declaration was breaking build in ubuntu16


tilne:
The description mentions ubuntu16, but the title mentions ubuntu18. Which does it affect?

PR #616, opened 2020-05-15
Avoid dpkg lock during Packer provisioner shell execution

lukeseawalker:
Instead of disabling the Update Package List and Unattended Upgrade in the apt configuration,
wait until dpkg lock is released and temporarily stop the apt-daily services for the Packer duration.

```
$ apt-config shell StateDir Dir::State/d
StateDir='/var/lib/apt/'
```

```
$ echo $(apt-config shell StateDir Dir::State/d | sed -r "s/.*'(.*\/?)'$/\1/")daily_lock
/var/lib/apt/daily_lock
```

*Issue #, if available:*

*Description of changes:*


amis/packer_ubuntu1604.json
```diff
@@ -153,12 +153,15 @@
         "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done"
       ]
     },
+    {
+      "type": "shell",
+      "inline": [
+        "sudo flock $(apt-config shell StateDir Dir::State/d | sed -r \"s/.*'(.*\\/?)'$/\\1/\")daily_lock systemctl stop apt-daily.timer apt-daily.service"
```
tilne, line 159: I think the regex in the sed command implies that either `StateDir='/var/lib/apt'` or `StateDir='/var/lib/apt/'` would match. In the first case, I think there'd need to be a `/` in front of `daily_lock`.

tilne: Does not disabling these services mean that there could be problems during AMI configuration time, like when chef is attempting to install base packages?

We haven't ever hit this issue so I assume the risk is very low. Just curious.

tilne: Regarding the firs comment, I saw from the apt-config man page that the trailing /d ensures the the trailing / is already there.

PR #622, opened 2020-05-29
Start slurmd with nodename/hostname

rexcsn:
* Add recipe to finalize compute configuration for slurm and start slurmd with correct nodename/hostname of the compute instance
* Disable supervisord if scheduler is slurm

### DIFF
--- a/files/default/cloudwatch_logs/cloudwatch_log_files.json
+++ b/files/default/cloudwatch_logs/cloudwatch_log_files.json
@@ -91,6 +91,25 @@
       "feature_conditions": []
     },
     {
+      "timestamp_format_key": "month_first",
+      "file_path": "/var/log/cloud-init-output.log",
+      "log_stream_name": "cloud-init-output",
+      "schedulers": [
+        "sge",
+        "slurm",
+        "torque"
+      ],
+      "platforms": [
+        "amazon",
+        "centos",
+        "ubuntu"
+      ],
+      "node_roles": [
+        "ComputeFleet"
+      ],
+      "feature_conditions": []
+    },
+    {
       "timestamp_format_key": "default",
       "file_path": "/var/log/supervisord.log",
       "log_stream_name": "supervisord",
--- a/recipes/_compute_slurm_finalize.rb
+++ b/recipes/_compute_slurm_finalize.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+#
+# Cookbook Name:: aws-parallelcluster
+# Recipe:: _compute_slurm_finalize
+#
+# Copyright 2013-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+ruby_block 'get_compute_nodename' do
+  block do
+    # Retrieve private ip from Metadata V2
+    require 'mixlib/shellout'
+    node.run_state['compute_ip'] = shell_out!('curl http://169.254.169.254/latest/meta-data/local-ipv4', :user => 'root').stdout.strip
+    # Retrieve NodeName from scontrol
+    node.run_state['slurm_compute_nodename'] = shell_out!("/opt/slurm/bin/scontrol show nodes | awk \"/\\y#{node.run_state['compute_ip']}\\y/\" RS= | grep -oP '^NodeName=\\K(\\S+)'", :user => 'root').stdout.strip
+  end
+  retries 3
+  retry_delay 5
+end
+
+directory '/etc/sysconfig' do
+  user 'root'
+  group 'root'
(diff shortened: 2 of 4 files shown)


recipes/_compute_slurm_finalize.rb
```diff
@@ -0,0 +1,54 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+ruby_block 'get_compute_nodename' do
+  block do
+    # Retrieve private hostname from Metadata V2
+    require 'mixlib/shellout'
+    get_hostname_cmd = Mixlib::ShellOut.new('TOKEN=`curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600"` && curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/local-hostname', :user => 'root')
```
demartinofra, line 22: nit: you could make this more compact by using shellout the same way it's used in fsx_mount:
```
mountname = shell_out!("#{node['cfncluster']['cookbook_virtualenv_path']}/bin/aws fsx --region #{node['cfncluster']['cfn_region']} describe-file-systems --file-system-ids #{node['cfncluster']['cfn_fsx_fs_id']} --query 'FileSystems[0].LustreConfiguration.MountName' --output text", :user=>'root').stdout.strip
```

demartinofra: for now you could keep using metadata v1 without the token as done elsewhere

demartinofra: I would use the private ip rather than the hostname to retrieve the nodename. it might work better with custom dns setup

recipes/_compute_slurm_finalize.rb
```diff
@@ -0,0 +1,54 @@
...
+    get_hostname_cmd.run_command
+    node.run_state['compute_hostname'] = get_hostname_cmd.stdout.strip.split('.')[0]
+    # Raise if cannot retrieve hostname
+    get_hostname_cmd.error!
+    # Retrieve NodeName from scontrol
+    get_slurm_nodename_cmd = Mixlib::ShellOut.new("/opt/slurm/bin/scontrol show nodes -F | awk \"/\\y#{node.run_state['compute_hostname']}\\y/\" RS= | grep -oP '^NodeName=\\K(\\S+)'", :user => 'root')
```
demartinofra, line 28: if `-F` is for future nodes we don't need it anymore

recipes/_compute_slurm_finalize.rb
```diff
@@ -0,0 +1,54 @@
...
+  retry_delay 5
+end
+
+directory '/etc/sysconfig' do
+  user 'root'
+  group 'root'
+  mode '0755'
+end
+
+file '/etc/sysconfig/slurmd' do
```
demartinofra, line 44: we might want to check if using sysconfig is compatible with all OSes

recipes/_compute_slurm_finalize.rb
```diff
@@ -0,0 +1,54 @@
...
+    get_slurm_nodename_cmd.error!
+  end
+  retries 3
+  retry_delay 5
+end
+
+directory '/etc/sysconfig' do
+  user 'root'
+  group 'root'
+  mode '0755'
```
demartinofra, line 41: probably 644 is more appropriate for directories.

recipes/finalize.rb
```diff
@@ -15,14 +15,29 @@
...
 # limitations under the License.
 
-# Restart supervisord
-service "supervisord" do
-  supports restart: true
-  action %i[enable start]
+case node['cfncluster']['cfn_scheduler']
+when 'slurm'
+  # disable supervisord when scheduler is slurm
+  service "supervisord" do
```
demartinofra, line 21: user could use supervisord for other needs. I would rather skip the creation of the parallelcluster supervisord config rather than explicitly disabling and stopping the service

demartinofra: also the master will most likely have a daemon for slurm as well, just a different one.

recipes/finalize.rb
```diff
@@ -15,14 +15,29 @@
...
+return unless node['cfncluster']['cfn_node_type'] == 'ComputeFleet'
+
+case node['cfncluster']['cfn_scheduler']
+when 'slurm'
+  include_recipe 'aws-parallelcluster::_compute_slurm_finalize'
+else
+  execute "compute_ready" do
+    command "/opt/parallelcluster/scripts/compute_ready"
+    environment('PATH' => '/usr/local/bin:/usr/bin/:$PATH')
+    only_if { node['cfncluster']['cfn_node_type'] == 'ComputeFleet' }
```
demartinofra, line 41: this is a duplicated condition. We could simplify all of this by removing the return and the case and moving this in the previous case on scheduler by simply adding an if on the node type.

files/default/cloudwatch_logs/cloudwatch_log_files.json
```diff
@@ -90,6 +90,25 @@
       ],
       "feature_conditions": []
     },
+    {
+      "timestamp_format_key": "month_first",
```
demartinofra, line 94: we might want to add this change to a separate commit

demartinofra, line 94: we'll have to check if this format applies to this logfile. It can also be done in a separate PR

PR #634, opened 2020-06-25
Don't expose pcluster pyenv installation by default

tilne:
* Install python versions and virtualenvs used by cookbook and node
  daemons in a pyenv system installation at /opt/parallelcluster/pyenv
  (instead of /usr/local/pyenv) in order to demonstrate that these
  aren't intended to be modified.
* On CentOS 6, where it's necessary to provide a system installation in
  order to provide a python version recent enough to install AWS CLI, do
  so in a system installation separate from that containing the
  ParallelCluster-specific packages (/usr/local/pyenv).
* Only create pyenv profile.d script that gets sourced on login on
  CentOS 6, so that the python version needed to support the AWS CLI is
  used by default.


templates/default/pyenv.sh.erb
```diff
@@ -1,12 +1,6 @@
 # Prefer a user pyenv over a system wide install
```
demartinofra, line 1: do we still want to create this file?

tilne: The thought was that it wouldn't hurt and it's an easy way to ease the use of pyenv for a user, but now I'm reconsidering because the purpose of this change was to make it more clear that we're not responsible for supporting a user's pyenv. I'll make this change.

PR #635, opened 2020-06-25
Fix installation of custom pcluster cli

demartinofra:
The logic was incompatible with the concurrent custom installation of node package because when executing `cd *aws-parallelcluster-*` it was entering the node directory and not the cli one. Because of this `pip install cli/` was failing.

### DIFF
--- a/recipes/awsbatch_config.rb
+++ b/recipes/awsbatch_config.rb
@@ -42,9 +42,10 @@ if !node['cfncluster']['custom_awsbatchcli_package'].nil? && !node['cfncluster']
     cwd Chef::Config[:file_cache_path]
     code <<-CLI
       set -e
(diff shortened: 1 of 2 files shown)


recipes/awsbatch_config.rb
```diff
@@ -42,9 +42,10 @@
     cwd Chef::Config[:file_cache_path]
     code <<-CLI
       set -e
-      curl --retry 3 -v -L -o aws-parallelcluster.tgz #{node['cfncluster']['custom_awsbatchcli_package']}
-      tar -xzf aws-parallelcluster.tgz
-      cd *aws-parallelcluster-*
+      curl --retry 3 -L -o aws-parallelcluster-custom-cli.tgz #{node['cfncluster']['custom_awsbatchcli_package']}
+      mkdir aws-parallelcluster-custom-cli
```
lukeseawalker, line 46: [MINOR] can we have the same change (create dir) on the custom node installation?

PR #643, opened 2020-07-02
Install PMIx v3.1.5

tilne:
This enables the slurm be built against it and use it when launching
jobs.

Avoid issues like [this one](https://github.com/aws/aws-parallelcluster/issues/1182) by passing `--mpi=pmix` to `srun`.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -59,6 +59,10 @@ default['cfncluster']['torque']['url'] = 'https://github.com/adaptivecomputing/t
 default['cfncluster']['slurm']['version'] = '19.05.5'
 default['cfncluster']['slurm']['url'] = 'https://download.schedmd.com/slurm/slurm-19.05.5.tar.bz2'
 default['cfncluster']['slurm']['sha1'] = '055adca91e555cc124b1ecac5f3c45e66c17a8ba'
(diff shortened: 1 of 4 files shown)


tilne:
Updating this with two changes:
* missed adding `libevent-devel` to the packages list for centos6
* the apt package equivalent is named `libevent-dev` instead of `libevent-devel`

Also rebasing on develop

attributes/default.rb
```diff
@@ -161,7 +165,7 @@
                                                 httpd boost-devel redhat-lsb mlocate mpich-devel openmpi-devel R atlas-devel
                                                 blas-devel fftw-devel libffi-devel openssl-devel dkms mysql-devel libedit-devel
                                                 libical-devel postgresql-devel postgresql-server sendmail mdadm python python-pip
-                                                libgcrypt-devel]
+                                                libgcrypt-devel libevent-devel]
```
demartinofra, line 168: isn't this getting applied to centos 6 too?

tilne: Yes. This is the package list for centos6 right?

demartinofra: sorry just me misreading your update :D

PR #646, opened 2020-07-06
Use PMIx-compatible autotools versions on CentOS 6

tilne:
PMIx builds require autoconf >= 2.69, automake >= 1.13.4, and libtool >=
2.4.2. The latest versions of these installable via yum on Cent OS 6 do
not meet these criteria. The first commit  in this PR downloads and installs
the newest stable versions of these packages from the GNU archive.

The second commit does the same for libevent because PMIx requires a
more recent version than the latest installable via yum (1.4.X).


tilne:
We're going to instead just not include PMIx on CentOS 6. Closing this PR and opening a new one.

PR #672, opened 2020-07-29
Add support for clustermgtd

rexcsn:
* Add logic needed to support clustermgtd
* Use supervisord to run clustermgtd when scheduler is Slurm
* Always try to resume node from compute in compute finalize

### DIFF
--- a/templates/default/parallelcluster_supervisord.conf.erb
+++ b/templates/default/parallelcluster_supervisord.conf.erb
@@ -3,7 +3,13 @@
 <% case node['cfncluster']['cfn_node_type'] -%>
 <%# MasterServer -%>
 <% when 'MasterServer' -%>
-<% unless node['cfncluster']['cfn_scheduler'] == 'awsbatch' -%>
+<% if node['cfncluster']['cfn_scheduler'] == 'slurm' -%>
+[program:clustermgtd]
+command = <%= node['cfncluster']['node_virtualenv_path'] %>/bin/clustermgtd
+redirect_stderr = true
+stdout_logfile = /var/log/parallelcluster/clustermgtd
+
+<% elsif node['cfncluster']['cfn_scheduler'] != 'awsbatch' -%>
 [program:sqswatcher]
 command = <%= node['cfncluster']['node_virtualenv_path'] %>/bin/sqswatcher
 redirect_stderr = true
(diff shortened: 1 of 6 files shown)


templates/default/slurm/parallelcluster_clustermgtd.conf.erb
```diff
@@ -0,0 +1,5 @@
+[clustermgtd]
+cluster_name = <%= node['cfncluster']['stack_name'].sub("parallelcluster-", "") %>
+region = <%= node['cfncluster']['cfn_region'] %>
+proxy = <%= node['cfncluster']['cfn_proxy'] %>
+heartbeat_file_path = /home/<%= node['cfncluster']['cfn_cluster_user'] %>/clustermgtd_heartbeat
```
demartinofra, line 5: I'd use something like /home/<%= node['cfncluster']['cfn_cluster_user'] %>/.parallelcluster/.metadata/clustermgtd_heartbeat

templates/default/parallelcluster_supervisord.conf.erb
```diff
@@ -3,7 +3,13 @@
 <% case node['cfncluster']['cfn_node_type'] -%>
 <%# MasterServer -%>
 <% when 'MasterServer' -%>
-<% unless node['cfncluster']['cfn_scheduler'] == 'awsbatch' -%>
+<% if node['cfncluster']['cfn_scheduler'] == 'slurm' -%>
+[program:clustermgtd]
+command = <%= node['cfncluster']['node_virtualenv_path'] %>/bin/clustermgtd
+redirect_stderr = true
+stdout_logfile = /var/log/clustermgtd
```
demartinofra, line 10: For the new logic we are writing some of the logs to `/var/log` and some others to `/var/log/parallelcluster/`.
What about unifying the behavior by having everything under the latter? (of course only for the new daemons/scripts)

templates/default/slurm/parallelcluster_clustermgtd.conf.erb
rexcsn: Done, using `/home/<%= node['cfncluster']['cfn_cluster_user'] %>/.parallelcluster/.slurm_plugin/clustermgtd_heartbeat`

PR #675, opened 2020-08-04
Add support for computemgtd

rexcsn:
* Add computemgtd specific configs, start computemgtd on slurm compute nodes
* Mount /opt/parallelcluster to all computes
* Use /opt/parallelcluster/slurm/clustermgtd_heartbeat for clustermgtd heartbeat

### DIFF
--- a/recipes/_compute_slurm_config.rb
+++ b/recipes/_compute_slurm_config.rb
@@ -23,6 +23,7 @@ directory '/var/spool/slurmd' do
 end
 
 # Mount /opt/slurm over NFS
+# Computemgtd config is under /opt/slurm/etc/pcluster; all compute nodes share a config
 mount '/opt/slurm' do
   device(lazy { "#{node['cfncluster']['cfn_master_private_ip']}:/opt/slurm" })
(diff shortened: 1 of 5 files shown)


recipes/_compute_base_config.rb
```diff
@@ -64,6 +64,16 @@
   retry_delay 5
 end
 
+# Mount pcluster base dir, /opt/parallelcluster, over NFS
+mount "#{node['cfncluster']['base_dir']}" do
```
demartinofra, line 68: why do we need this?

demartinofra: don't we need to remove this?

recipes/_master_base_config.rb
```diff
@@ -133,6 +133,13 @@
   options ['no_root_squash']
 end
 
+# Export pcluster base dir, /opt/parallelcluster
+nfs_export "#{node['cfncluster']['base_dir']}" do
```
demartinofra, line 137: same here. We decided not to use this extra mount point right?

PR #678, opened 2020-08-07
Enable nfsv4

yuleiwan:
* Enable nfsv4 for computing instances file system mounting 
* Amend the comment for pcluster 2.8.0

### DIFF
--- a/recipes/_compute_base_config.rb
+++ b/recipes/_compute_base_config.rb
@@ -36,7 +36,7 @@ if raid_shared_dir != "NONE"
   mount raid_shared_dir do
     device "#{nfs_master}:#{raid_shared_dir}"
     fstype 'nfs'
-    options 'hard,intr,noatime,vers=3,_netdev'
+    options 'hard,intr,noatime,_netdev'
     action %i[mount enable]
     retries 3
     retry_delay 5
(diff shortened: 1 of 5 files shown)


recipes/_compute_base_config.rb
```diff
@@ -36,7 +36,7 @@
   mount raid_shared_dir do
     device "#{nfs_master}:#{raid_shared_dir}"
     fstype 'nfs'
-    options 'hard,intr,noatime,vers=3,_netdev'
+    options 'hard,intr,noatime,_netdev'
```
demartinofra, line 39: shall we pin version 4 rather than defaulting to it? my concern is that if we don't we always default to latest without any control over the version we are using.

Also, and this is something extra, are there any options that can we use to improve our NFS performance? For example here are the ones recommended for an EFS file system.

lukeseawalker: - we are defaulting to the latest version offered by the server and somehow we are pinning the server version on the master side, installing "the platform-specific server services for an effective protocol level of NFSv4", done by including the recipes [nfs::server4](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/recipes/base_install.rb#L122). I think it's reasonable to leave the version in the client unspecified

- this could be investigated and left as another PR

demartinofra: I think as long as we can control the version we use without automatically upgrading to latest I'm perfectly fine with that. Would it make sense to add at this point a kitchen test or modify the integ tests to assert that the used protocol is v4 when mounting from the compute nodes?

PR #686, opened 2020-08-19
Configure hostname and DNS cluster domain to be part of the Route53 hosted zone

enrico-usai:
Within the cluster we're creating a private hosted zone named `<clustername>.parallelcluster`
With this patch we're extending DNS resolver configuration by appending the Route53 domain created within the cluster.

The private domain will be added as a `search` domain in the `resolv.conf` file by `dhclient` in Centos6, CentOS7, Alinux, Alinux2 and Ubuntu16 and by `systemd-resolved` in Ubuntu18. In this way we'll be able to resolve the short version of the FQDN.

In the HIT case we're also configuring the hostname to be something in the form: `$QUEUE-static-$INSTANCE_TYPE_1-[1-$MIN1]`
NOTE: we're replacing the "." in the instance type with the "-" character because the period is used in hostnames to split subdomains.

# Relevant changes
* I added the `bind-utils` package to Centos7 to have the `dig` utility,  useful to check DNS resolution.
* In Centos6 there is a dhclient configuration file for eth0 so we're  setting the domain tin the `/etc/dhcp/dhclient-eth0.conf` file too.
+ Refactored the code to configure the hostname to be a function.
+ Included `cloudwatch_agent config` (and install) in prep_env to have `dns_config` (and pre-install) logs available on cloudwatch.
+ Included `_setup_python` recipe in `prep_env` to have the python virtual env available for `dns_config` and `cloudwatch_agent install` actions.
+ Added `hosted_zone` configuration parameter to `slurm_resume.conf`
+ Add missing changes from: https://github.com/aws/aws-parallelcluster/commit/4deb4dc8c85ee0d031354320b0afb116cc06ff76
+ NetworkManager is not available in Alinux2 Karoo, so we are using ignore_failure
  at network restart

# Important notes about Chef
First all recipes on the run list are compiled in to resources in the resource collection, then all resources in the collection are converged. Any code outside lazy, only_if / not_if, or a ruby_block's block property is run at compile time, which is before any resource runs its actions.
From: https://coderanger.net/two-pass/

The ohai_hint and hostname resources run at compile time by default. This is due to the fact that later resources may consume the node attributes which are set by those resources leading to excessive use of lazy in subsequent resources.
From: https://docs.chef.io/resource_common/

# References
* shell_out: https://docs.chef.io/dsl_recipe/#shell_out-1
* lazy: https://docs.chef.io/resource_common/#lazy-evaluation
* replace_or_add: https://github.com/sous-chefs/line/blob/master/documentation/resources/replace_or_add.md
* dhclient: https://wiki.archlinux.org/index.php/NetworkManager#DHCP_client
* systemd-resolved: https://wiki.archlinux.org/index.php/Systemd-resolved
* dynamodb get-item: https://docs.aws.amazon.com/cli/latest/reference/dynamodb/get-item.html

# Tests
* Retry in case of missing permission
* Retry in case of wrong key to search
* Head node and compute fleet configuration


recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+# Utility function to retrieve the assigned hostname from DynamoDB
+def retrieve_assigned_hostname
+  instance_id = shell_out!('curl -s http://169.254.169.254/latest/meta-data/instance-id', user: 'root').stdout.strip
+  Chef::Log.info("Instance id is '#{instance_id}'")
+
+  assigned_hostname = shell_out!("#{node['cfncluster']['cookbook_virtualenv_path']}/bin/aws dynamodb " \
```
demartinofra, line 24: this can be a problem for runtime baked AMIs where cookbook virtualenv is not available

recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+      "--region #{node['cfncluster']['cfn_region']} query --table-name #{node['cfncluster']['stack_name']} " \
```
demartinofra, line 25: let's use `cfn_ddb_table` instead of `stack_name`

recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+      "--index-name InstanceId --key-condition-expression 'InstanceId = :instanceid' " \
+      "--expression-attribute-values '{\":instanceid\": {\"S\":\"#{instance_id}\"}}' " \
+      "--output text --query 'Items[0].Id.S'", user: 'root').stdout.strip
```
demartinofra, line 28: just to confirm: when item is not in the table the query return None?

recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+      retry
+    end
+
+  else
+    # Head node
+    configure_hostname(node['ec2']['local_hostname'].split('.')[0].to_s)
+  end
+
+else
+  # Single Instance Type or master instance for Heterogeneous Instance Type
```
demartinofra, line 146: master instance for HIT is not handled here

recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+  end
+end
+
+if node['cfncluster']['cfn_scheduler'] == 'slurm'
+  # Heterogeneous Instance Type
+  if node['cfncluster']['cfn_dns_domain'].nil? ||
+     node['cfncluster']['cfn_dns_domain'].empty? ||
+     node['cfncluster']['cfn_dns_domain'] == "NONE"
+    Chef::Log.info("DNS domain not configured")
+    return
```
demartinofra, line 115: I think we can't just return. We should consider the following:
- option 1: the user disables DNS resolution (which will more likely be a configuration parameter). In this case we still configure the hostname as the Slurm nodename but DNS resolution won't work. Users with custom DNS might want to use this option
- option 2: the user disables DNS resolution but also wants to preserve the old way of configuring hostnames (this can be set with an extra_json parameter). In this case we need to still configure hostname the old fashion.

recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  restart_network_service
+
+  if node['cfncluster']['cfn_node_type'] == "ComputeFleet"
+    # For compute node retrieve assigned hostname from DynamoDB and configure it
+    # - hostname: $QUEUE-static-$INSTANCE_TYPE_1-[1-$MIN1]
+    # - fqdn: $QUEUE-static-$INSTANCE_TYPE_1-[1-$MIN1].$CLUSTER_NAME.pcluster
+    Chef::Log.info("Retrieving assigned hostname")
+    begin
+      retries ||= 0
```
demartinofra, line 128: to simplify retries you can use a ruby_block

recipes/dns_config.rb
```diff
@@ -0,0 +1,148 @@
...
+      "--output text --query 'Items[0].Id.S'", user: 'root').stdout.strip
+  Chef::Log.info("Assigned_hostname is '#{assigned_hostname}'")
+
+  assigned_hostname
+end
+
+# Utility function to configure hostname according to the OS
+def configure_hostname(hostname)
+  Chef::Log.info("Configuring hostname to '#{hostname}'")
+  hostname node['ec2']['local_hostname']
```
demartinofra, line 37: what is this line for?

recipes/dns_config.rb
enrico-usai: Yes

recipes/dns_config.rb
enrico-usai: Right, old comment

recipes/dns_config.rb
enrico-usai: Yes, I'm going to add and test the two configuration parameters

recipes/dns_config.rb
enrico-usai: Interesting comment, so it seems we have a problem in multiple places:
https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/hit/recipes/_master_slurm_config.rb#L63
https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/hit/recipes/_master_slurm_config.rb#L78
https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/hit/recipes/_master_slurm_finalize.rb#L20
https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/hit/recipes/fsx_mount.rb#L35
https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/hit/recipes/setup_raid_on_master.rb#L42
https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/hit/recipes/slurm_install.rb#L55

right?

recipes/dns_config.rb
enrico-usai: I started using the `ruby_block` but there are issues passing the retrieved value.

I might use `run_state` but it can only be used between chef resources and the `configure_hostname` function cannot be used in a chef resource.

recipes/dns_config.rb
enrico-usai: Fixed by moving _setup_python.rb recipe in prep_env. To be tested with not backed AMIs

recipes/dns_config.rb
enrico-usai: This line is to set the fqdn and update the /etc/hosts file, but thanks to your comment I saw I was using the ec2 local_hostname. Fixed it by using the right domain when provided.

demartinofra (approved):

recipes/dns_config.rb
enrico-usai: Fixe by using ruby_blocks, helpers and passing `compile_time = false` to the `hostname` resource that by default it is executed a t compile time.

libraries/helpers.rb
```diff
@@ -171,6 +171,43 @@ def master_address(region, stack_name)
...
+    %w[ubuntu debian] => {
+      '16.04' => 'networking',
+      '>=18.04' => 'systemd-resolved'
+    },
+    'default' => 'NetworkManager'
+  )
+  Chef::Log.info("Restarting '#{network_service_name}' service, platform #{node['platform']} '#{node['platform_version']}'")
+  service network_service_name.to_s do
+    action %i[restart]
+    ignore_failure true
```
demartinofra, line 207: can this hide potential bugs? shall we at least add a comment to explain why we want to ignore failures?

demartinofra (approved):

PR #690, opened 2020-08-20
Scaling Optimization

rexcsn:
* Set slurm.conf SuspendRate=0 (unlimited), the origin default was SuspendRate=60/min
* Only resume static nodes from compute finalize. This does not change existing behaviors because dynamic nodes do not need to be resumed anyway, but will lighten load on slurmctld
* Use /etc/parallelcluster/slurm_plugin for slurm cloud bursting integration related configs instead of /opt/parallelcluster
* Increase /proc/sys/net/core/somaxconn from default 128 to 1024 to handle large scale


recipes/_compute_slurm_finalize.rb
```diff
@@ -70,8 +70,10 @@
 end
 
 execute 'resume_node' do
-  # Always try to resume the node on start up
+  # Always try to resume a static node on start up
   # Command will fail if node is already in IDLE, ignoring failure
   command(lazy { "/opt/slurm/bin/scontrol update nodename=#{node.run_state['slurm_compute_nodename']} state=resume reason='Node start up'" })
   ignore_failure true
+  # Only resume static nodes
+  only_if { "#{node.run_state['slurm_compute_nodename']}".include? "static" }
```
demartinofra, line 78: minor: let's match "-static-"

do we need lazy also in the only_if?

PR #705, opened 2020-08-28
Slurm: enable cloud_dns option when DNS is configured

demartinofra:
Since all nodes are resolvable over DNS we can enable cloud_dns option and leave TreeWidth to its default (50). This can improve improve communications when jobs request lots of nodes and when there are a lot of active nodes in
the system. With TreeWidth=65533, that means that slurmctld will ping every node individually. With TreeWidth=50, slurmctld only needs to ping 50 of them and it will fan it out to the other nodes.

Tested a slurm restart and scontrol reconfigure to check if nodeaddr and nodehostname are preserved.

### DIFF
--- a/templates/default/slurm/slurm.conf.erb
+++ b/templates/default/slurm/slurm.conf.erb
@@ -23,13 +23,17 @@ CacheGroups=0
 #
 # CLOUD CONFIGS OPTIONS
 SlurmctldParameters=idle_on_node_suspend
+<% if node['cfncluster']['use_private_hostname'] == 'true' or node['cfncluster']['cfn_dns_domain'].nil? or node['cfncluster']['cfn_dns_domain'].empty? -%>
 CommunicationParameters=NoAddrCache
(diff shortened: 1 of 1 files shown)


templates/default/slurm/slurm.conf.erb
```diff
@@ -23,13 +23,17 @@ CacheGroups=0
 #
 # CLOUD CONFIGS OPTIONS
 SlurmctldParameters=idle_on_node_suspend
+<% if node['cfncluster']['use_private_hostname'] == 'true' or node['cfncluster']['cfn_dns_domain'].nil? or node['cfncluster']['cfn_dns_domain'].empty? -%>
 CommunicationParameters=NoAddrCache
+TreeWidth=65533
+<% else -%>
+CommunicationParameters=NoAddrCache,cloud_dns
```
rexcsn, line 30: cloud_dns is a SlurmctldParameters

PR #710, opened 2020-08-31
Slurm config optimization

rexcsn:
* Increase ResumeTimeout to 1 hr, in case nodes have longer pre/post-install scripts
* Enable cloud_dns and removed TreeWidth to reduce pinging between slurmctld and compute nodes
* Change TaskPlugin to task/cgroup to enforce better resource affinity when scheduling jobs

### DIFF
--- a/templates/default/slurm/cgroup.conf.erb
+++ b/templates/default/slurm/cgroup.conf.erb
@@ -2,3 +2,5 @@
 # Slurm cgroup support configuration file
 ###
 CgroupAutomount=yes
+TaskAffinity=no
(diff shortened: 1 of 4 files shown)


templates/default/slurm/cgroup.conf.erb
```diff
@@ -2,3 +2,5 @@
 # Slurm cgroup support configuration file
```
demartinofra, line 2: we also have other cgroup.conf files for centos6 and alinux

demartinofra (approved):

PR #733, opened 2020-09-10
Use > 8 NFS server kernel threads on instance types with many cores

tilne:
Provide an attribute so that it may be disabled via extra_json.

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -116,6 +116,7 @@ cookbook_file 'AWS-ParallelCluster-License-README.txt' do
   mode '0644'
 end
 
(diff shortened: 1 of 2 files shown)


tilne:
I'm closing this because the nfs cookbook doesn't appear to work for AL2.

recipes/base_install.rb
```diff
@@ -116,6 +116,7 @@
   mode '0644'
 end
 
+node.default['nfs']['threads'] = node['cfncluster']['nfs_server_threads'].to_i
```
tilne, line 119: If we do revisit this, we need to do this configuration at runtime, not at AMI build time.

PR #735, opened 2020-09-11
Disallow using AMIs created by other versions of pcluster

hanwen-pcluste:
As a result, a AMI created by other versions of pcluster cannot be used as a BASE_AMI in AMI creation or a custom_AMI in cluster creation.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -145,13 +145,19 @@ end
 def ami_bootstrapped?
   version = ''
   bootstrapped_file = '/opt/parallelcluster/.bootstrapped'
+  current_version = 'aws-parallelcluster-' + node['cfncluster']['cfncluster-version']
 
   if ::File.exist?(bootstrapped_file)
     version = IO.read(bootstrapped_file).chomp
     Chef::Log.info("Detected bootstrap file #{version}")
+    if version != current_version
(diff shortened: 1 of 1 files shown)


libraries/helpers.rb
```diff
@@ -145,13 +145,17 @@ def graphic_instance?
 def ami_bootstrapped?
   version = ''
   bootstrapped_file = '/opt/parallelcluster/.bootstrapped'
+  current_version = 'aws-parallelcluster-' + node['cfncluster']['cfncluster-version']
 
   if ::File.exist?(bootstrapped_file)
     version = IO.read(bootstrapped_file).chomp
     Chef::Log.info("Detected bootstrap file #{version}")
+    if version != current_version
+      raise "The AMI can only be used by " + version + " but currently it is " + current_version
```
rexcsn, line 154: Let's make sure we manually test that this logic fails quickly in both createami and cluster creation case.

rexcsn (dismissed):

libraries/helpers.rb
hanwen-pcluste: It takes about 5 minutes to detect the error when creating an ami. It is almost the very first step after chef is set up. If we want to reduce the time to 3-4 minutes, we have to use shell commands instead of recipes to detect the error

rexcsn (approved):

PR #736, opened 2020-09-14
Add a new option post_install_script to createami CLI

yuleiwan:
* Upload script folder with createami post install script to packer instance.
* Execute the script if the script is specified.
* Move the script to `/opt/parallelcluster/createami_scripts` folder.
* [CLI PR](https://github.com/aws/aws-parallelcluster/pull/2039) and [tests performance](https://jenkins.yuleiwan.ds9.nice.aws.a2z.com/job/02_integration_test/39/testReport/)

### DIFF
--- a/amis/build_ami.sh
+++ b/amis/build_ami.sh
@@ -117,8 +117,8 @@ check_options() {
     available_arm_os="ubuntu1804 alinux2"  # subset of supported OSes for which ARM AMIs are available
     available_os="centos6 centos7 alinux ubuntu1604 ${available_arm_os}"
     cwd="$(dirname $0)"
-    tmp_dir=$(mktemp -d)
-    export COOKBOOK_PATH="${tmp_dir}/cookbook"
+    export COOKBOOK_PATH="$(cd ${cwd}/..; pwd)"
+    export SCRIPT_PATH="${cwd}/../../script"
 
     if [ "${_custom}" == "true" ]; then
         only=custom-${_os}
@@ -203,8 +203,8 @@ check_options() {
 do_command() {
     RC=0
 
-    mkdir -p "${COOKBOOK_PATH}"
-    cp -R ${cwd}/../* "${COOKBOOK_PATH}"
+    mkdir -p "${SCRIPT_PATH}"
+
     if [ "x${_build_date}" == "x" ]; then
       export BUILD_DATE=$(date +%Y%m%d%H%M)
     else
--- a/amis/packer_alinux.json
+++ b/amis/packer_alinux.json
@@ -9,6 +9,7 @@
     "ami_perms" : "{{env `AMI_PERMS`}}",
     "build_date" : "{{env `BUILD_DATE`}}",
     "cookbook_path" : "{{env `COOKBOOK_PATH`}}",
+    "script_path": "{{env `SCRIPT_PATH`}}",
     "nvidia_enabled" : "{{env `NVIDIA_ENABLED`}}",
     "instance_type" : "{{env `AWS_FLAVOR_ID`}}",
     "subnet_id" : "{{env `AWS_SUBNET_ID`}}",
@@ -167,6 +168,11 @@
       "destination" : "/tmp/cookbook"
     },
     {
+      "type": "file",
+      "source": "{{user `script_path`}}",
+      "destination": "/tmp/createami_scripts"
+    },
+    {
       "type" : "shell",
       "inline" : [
         "set -x",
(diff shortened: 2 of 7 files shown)


amis/build_ami.sh
```diff
@@ -117,8 +117,9 @@ check_options() {
     available_arm_os="ubuntu1804 alinux2"  # subset of supported OSes for which ARM AMIs are available
     available_os="centos6 centos7 alinux ubuntu1604 ${available_arm_os}"
     cwd="$(dirname $0)"
-    tmp_dir=$(mktemp -d)
+    tmp_dir=${cwd}/../..
     export COOKBOOK_PATH="${tmp_dir}/cookbook"
```
lukeseawalker, line 121: if you set `COOKBOOK_PATH` to be `${cwd}/..`
there is no need to create another cookbook dir and copy the content over again, e.g. you can remove the following code in the do_command function
```
mkdir -p "${COOKBOOK_PATH}"
cp -R ${cwd}/../* "${COOKBOOK_PATH}"
```

amis/build_ami.sh
```diff
@@ -117,8 +117,9 @@ check_options() {
...
+    export SCRIPT_PATH="${tmp_dir}/script"
```
lukeseawalker, line 122: what happens if `SCRIPT_PATH` is not present? this is the case when `build_ami.sh` is called directly without passing from the `createami` command.

If we avoid to fail in case it's not present, we can omit the `"only":` attribute in the packer provisioners and always execute them

amis/packer_alinux.json
```diff
@@ -166,6 +167,12 @@
       "source" : "{{user `cookbook_path`}}",
       "destination" : "/tmp/cookbook"
     },
+    {
+      "type": "file",
+      "only": ["custom-alinux"],
+      "source": "{{user `script_path`}}",
+      "destination": "/tmp/script"
```
lukeseawalker, line 174: let's be more specific in destination name, something like `/tmp/post_install_script_dir` or `/tmp/createami_scripts`

amis/packer_alinux.json
```diff
@@ -249,6 +256,15 @@
         "aws-parallelcluster::default"
       ]
     },
+    {
+      "type": "shell",
+      "only": ["custom-alinux"],
+      "inline": [
+        "set -x",
+        "script_name=$(ls /tmp/script)",
+        "[[ ! -z $script_name ]] && (sudo chmod +x /tmp/script/${script_name} && sudo /tmp/script/${script_name} && sudo mkdir -p /opt/parallelcluster/createami_scripts && sudo mv /tmp/script/${script_name} /opt/parallelcluster/createami_scripts) || echo 'No post script'"
```
lukeseawalker, line 265: `[[ -n ${script_name} ]]`

amis/build_ami.sh
yuleiwan: COOKBOOK_PATH to be `${cwd}/..` met an error, maybe because we can't upload path /.. to tmp/cookbook, so I use `export COOKBOOK_PATH="$(cd ${cwd}/..; pwd)"`

amis/packer_alinux.json
```diff
@@ -249,6 +255,15 @@
         "aws-parallelcluster::default"
       ]
     },
+    {
+      "type": "shell",
+      "only": ["custom-alinux"],
```
lukeseawalker, line 260: this and all the others `only` can now be removed, right?

amis/packer_alinux.json
```diff
@@ -249,6 +255,15 @@
...
+      "inline": [
+        "set -x",
+        "script_name=$(ls /tmp/createami_scripts)",
+        "[[ -n $script_name ]] && (sudo chmod +x /tmp/createami_scripts/${script_name} && sudo /tmp/createami_scripts/${script_name} && sudo mkdir -p /opt/parallelcluster/createami_scripts && sudo mv /tmp/createami_scripts/${script_name} /opt/parallelcluster/createami_scripts) || echo 'No post script'"
```
lukeseawalker, line 264: `echo 'No post script'"` -> `echo 'No post install script'"`

amis/packer_alinux.json
```diff
@@ -255,6 +261,14 @@
         "aws-parallelcluster::default"
       ]
     },
+    {
+      "type": "shell",
+      "inline": [
+        "set -x",
+        "script_name=$(ls /tmp/createami_scripts)",
+        "[[ -n ${script_name} ]] && (sudo chmod +x /tmp/createami_scripts/${script_name} && sudo /tmp/createami_scripts/${script_name} && sudo mkdir -p /opt/parallelcluster/createami_scripts && sudo mv /tmp/createami_scripts/${script_name} /opt/parallelcluster/createami_scripts) || echo 'No post script'"
```
lukeseawalker, line 269: please change `'No post script'` to `'No post install script'`

hanwen-pcluste (commented):
Great work! Sorry for not adding any specific comments, because everything does look good to me.

enrico-usai (approved):
If you tested it both with and without post-install script it is ready to go for me.
If so could you add the performed tests in the commit message and/or in the PR description?

In the PR description it is also useful to link the related PR for the CLI, in this way they will be connected to each other.

PR #738, opened 2020-09-15
Fix condition to avoid failure when partition changes require a reboot

lukeseawalker:
The patch will let the script continue also when the following error is returned by the parted command: "Error: Partition(s) Y on /dev/XXX have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use. As a result, the old partition(s) will remain in use. You should reboot now before making further changes."

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/files/default/setup-ephemeral-drives.sh
+++ b/files/default/setup-ephemeral-drives.sh
@@ -24,9 +24,11 @@ function error_exit () {
 function exec_command() {
   _command_output=$($@ 2>&1)
   _exit_code=$?
+  echo "${_command_output}" | grep -vi "you should reboot now" >& /dev/null
+  _grep_exit_code=$?
 
   # Do not set RC=1 if error says that changes have been written but a reboot is required to inform the kernel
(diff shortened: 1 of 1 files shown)


files/default/setup-ephemeral-drives.sh
```diff
@@ -26,7 +26,7 @@ function exec_command() {
   _exit_code=$?
 
   # Do not set RC=1 if error says that changes have been written but a reboot is required to inform the kernel
-  [[ $_exit_code -ne 0 && $(echo "${_command_output}" | grep -i "you should reboot now") ]] && RC=1
+  [[ $_exit_code -ne 0 && $(echo "${_command_output}" | grep -vi "you should reboot now") ]] && RC=1
```
demartinofra, line 29: grep will return 1 if it does't find any match and the overall condition will result to false right?

lukeseawalker: check improved

demartinofra (approved):

PR #740, opened 2020-09-16
Do not use restart command for chrony service

lukeseawalker:
The restart command for chrony service is not correctly checking if the process is stopped before starting the new one, causing the new process to not being able to start
e.g.

```
chrony[2359]: chronyd is running and online.
systemd[1]: Started LSB: Controls chronyd NTP time daemon.
chronyd[2463]: Selected source 50.205.244.25
chronyd[2463]: Selected source 169.254.169.123
chrony[4195]: Stopped /usr/sbin/chronyd (pid 2463).
chronyd[2463]: chronyd exiting
systemd[1]: Stopped LSB: Controls chronyd NTP time daemon.
systemd[1]: Starting LSB: Controls chronyd NTP time daemon...
chrony[4203]: /usr/sbin/chronyd already running
```

Replace the restart with a stop/start sequence handled by chef

Fix also the check to see if a service is running when systemctl is used
```
# systemctl show -p SubState chrony
SubState=exited
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/recipes/chrony_config.rb
+++ b/recipes/chrony_config.rb
@@ -16,6 +16,7 @@
 # limitations under the License.
 
 service node['cfncluster']['chrony']['service'] do
-  supports restart: true
+  # chrony service supports restart but is not correctly checking if the process is stopped before starting the new one
+  supports restart: false
   action %i[enable restart]
 end
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -183,8 +183,10 @@ if node['init_package'] == 'init'
   gmond_check_command = "service #{node['cfncluster']['ganglia']['gmond_service']} status | grep -i running"
   gmetad_check_command = "service gmetad status | grep -i running"
 elsif node['init_package'] == 'systemd'
-  gmond_check_command = "systemctl status #{node['cfncluster']['ganglia']['gmond_service']} | grep -i running"
-  gmetad_check_command = "systemctl status gmetad | grep -i running"
+  # $ systemctl show -p SubState <service>
+  # SubState=Running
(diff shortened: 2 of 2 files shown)


recipes/chrony_config.rb
```diff
@@ -16,6 +16,6 @@
 # limitations under the License.
 
 service node['cfncluster']['chrony']['service'] do
-  supports restart: true
+  supports restart: false
```
enrico-usai, line 19: Minor: we could avoid to set it because it is the default. But we can leave it as is to explicitly specify it.

recipes/tests.rb
```diff
@@ -183,8 +183,8 @@
   gmond_check_command = "service #{node['cfncluster']['ganglia']['gmond_service']} status | grep -i running"
   gmetad_check_command = "service gmetad status | grep -i running"
 elsif node['init_package'] == 'systemd'
-  gmond_check_command = "systemctl status #{node['cfncluster']['ganglia']['gmond_service']} | grep -i running"
-  gmetad_check_command = "systemctl status gmetad | grep -i running"
+  gmond_check_command = "systemctl show -p SubState #{node['cfncluster']['ganglia']['gmond_service']} | grep -i running"
```
enrico-usai, line 186: let's add an example of the output we're expecting.

recipes/tests.rb
```diff
@@ -183,8 +183,8 @@
...
+  gmetad_check_command = "systemctl show -p SubState gmetad | grep -i running"
```
enrico-usai, line 187: same here

enrico-usai (approved):

enrico-usai (approved):

enrico-usai (approved):

PR #741, opened 2020-09-18
Validate OS specified by users is the same as the actual OS of a AMI

hanwen-pcluste:
The OS specified by users in config file or command parameter will be compared with the actual OS of a AMI in cluster/AMI creation process. The creation process will fail if the two OSs are not consistent.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -15,6 +15,8 @@ suites:
       - recipe[aws-parallelcluster::_default_pre]
       - recipe[aws-parallelcluster::default]
     attributes:
+      cfncluster:
+        cfn_base_os: <%= ENV['CFN_BASE_OS'] %>
 
   - name: default_gpu
     run_list:
@@ -22,6 +24,7 @@ suites:
       - recipe[aws-parallelcluster::default]
     attributes:
       cfncluster:
+        cfn_base_os: <%= ENV['CFN_BASE_OS'] %>
         nvidia:
           enabled: 'yes'
 
@@ -67,7 +70,10 @@ suites:
         cfn_ephemeral_dir: <%= ENV['CFN_EPHEMERAL_DIR'] %>
         cfn_shared_dir: <%= ENV['CFN_SHARED_DIR'] %>
         cfn_cluster_user: <%= ENV['CFN_CLUSTER_USER'] %>
+        # cfn_base_os is one of the base OSs supported by pcluster
+        cfn_base_os: <%= ENV['CFN_BASE_OS'] %>
         custom_node_package: <%= ENV['PARALLELCLUSTER_NODE_URL'] %>
+        # os attribute is used in pipeline code. i.e. centos7, centos7-custom
         os: <%= ENV['OS'] %>
 
   - name: sge_config_MasterServer
@@ -90,7 +96,10 @@ suites:
         cfn_cluster_user: <%= ENV['CFN_CLUSTER_USER'] %>
         cfn_sqs_queue: <%= ENV['CFN_SQS_QUEUE'] %>
         cfn_ddb_table: <%= ENV['CFN_DDB_TABLE'] %>
+        # cfn_base_os is one of the base OSs supported by pcluster
+        cfn_base_os: <%= ENV['CFN_BASE_OS'] %>
         custom_node_package: <%= ENV['PARALLELCLUSTER_NODE_URL'] %>
+        # os attribute is used in pipeline code. i.e. centos7, centos7-custom
         os: <%= ENV['OS'] %>
         ganglia_enabled: 'yes'
         dcv_enabled: 'master'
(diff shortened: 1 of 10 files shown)


rexcsn (commented):
General comment, `cfn_base_os` is only included in `dna.json` of the head node. The same validation will be executed on the compute nodes, so we also need to find a way to add `cfn_base_os` to compute nodes.

hanwen-pcluste:
Thank you @rexcsn for pointing out the bug. The bug will be triggered when submitting a job, because compute nodes cannot be initialized. The easiest fix is to bypass the check if `cfn_base_os` is `None`, I will think about if it is the right solution

.kitchen.yml
```diff
@@ -90,6 +94,7 @@ suites:
         cfn_cluster_user: <%= ENV['CFN_CLUSTER_USER'] %>
         cfn_sqs_queue: <%= ENV['CFN_SQS_QUEUE'] %>
         cfn_ddb_table: <%= ENV['CFN_DDB_TABLE'] %>
+        cfn_base_os: <%= ENV['CFN_BASE_OS'] %>
         custom_node_package: <%= ENV['PARALLELCLUSTER_NODE_URL'] %>
         os: <%= ENV['OS'] %>
```
rexcsn, line 99: Minor: we are also exporting a kitchen test OS here. On a first glance it seems a bit strange that we are exporting OS in 2 different places. When looking in the code, `ENV['OS']` is used to separate out different kitchen test cases, such as `centos7` and `centos7-custom`.
To make things more obvious, I would maybe rename this to `kitchen_test_os: <%= ENV['OS'] %>`? And/or add a line of comment showing the example of `centos7` vs `centos7-custom`
If we want to change this let's change it everywhere in this file.

libraries/helpers.rb
```diff
@@ -160,6 +160,39 @@ def ami_bootstrapped?
   version != '' && ( node['cfncluster']['skip_install_recipes'] == 'yes' || node['cfncluster']['skip_install_recipes'] == true )
 end
 
+#
+# Check if OS type specified by the user is the same as the OS identified by Ohai
+#
+def validate_OS_type
+  case node['platform']
+  when 'ubuntu'
+    current_os = 'ubuntu'+ node['platform_version'].tr('.','')
```
rexcsn, line 169: Can we think about having a common function to retrieve OS from ohai, and use it everywhere in the cookbook? For example we currently have scattered logic [like this](https://github.com/aws/aws-parallelcluster-cookbook/blob/v2.9.1/attributes/default.rb#L106) and in other places when we have to determine the OS in cookbook.
Is it possible to use a common logic for all of these so we don't have to keep track of these logic separately?

libraries/helpers.rb
```diff
@@ -160,6 +160,39 @@ def ami_bootstrapped?
   version != '' && ( node['cfncluster']['skip_install_recipes'] == 'yes' || node['cfncluster']['skip_install_recipes'] == true )
 end
 
+#
+# Check if OS type specified by the user is the same as the OS identified by Ohai
+#
+def validate_OS_type
```
rexcsn, line 166: Minor: Let's try to name things in `snake_case`(lowercase) according to [this guide](https://github.com/rubocop-hq/ruby-style-guide#snake-case-symbols-methods-vars)

libraries/helpers.rb
hanwen-pcluste: Can I do it in a separate PR? This change requires change in multiple files that are not relevant to the current task. And separating the PR also helps to identify the reason if one of the PR fails some tests.

.kitchen.yml
hanwen-pcluste: I added some comments. Changing the name requires code change in other files. I would prefer keep the current name.

recipes/default.rb
```diff
@@ -15,6 +15,9 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Validate OS type specified by the user is the same as the OS identified by Ohai
+validate_OS_type
```
rexcsn, line 19: This needs to change since the method is renamed

recipes/prep_env.rb
```diff
@@ -15,6 +15,9 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Validate OS type specified by the user is the same as the OS identified by Ohai
+validate_OS_type
```
rexcsn, line 19: This needs to change since the method is renamed

.kitchen.yml
rexcsn: Sounds good.

On another note, comments can be a lot shorter, we want to describe the attributes rather than listing all available values.
For `cfn_base_os` we can simply add a comment saying something like, "base_os supported by pcluster"
For `os` we can simply say something like, "os attribute used in kitchen test pipeline, i.e. centos7/centos7-custom"

rexcsn (approved):
LGTM, nice work!

PR #742, opened 2020-09-22
Add check on munge library

lukeseawalker:
Check is added to verify if munge library are installed. Munge libraries are used later in the process to compile other binaries (e.g. torque)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/recipes/munge_install.rb
+++ b/recipes/munge_install.rb
@@ -50,9 +50,7 @@ bash 'make install' do
     make -j $CORES
     make install
   MUNGE
-  # TODO: Fix, so it works for upgrade
-  creates '/usr/bin/munge'
-  not_if "/usr/sbin/munged --version | grep -q munge-#{node['cfncluster']['munge']['munge_version']}"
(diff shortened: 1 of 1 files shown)


recipes/munge_install.rb
```diff
@@ -50,9 +50,7 @@
     make -j $CORES
     make install
   MUNGE
-  # TODO: Fix, so it works for upgrade
-  creates '/usr/bin/munge'
-  not_if "/usr/sbin/munged --version | grep -q munge-#{node['cfncluster']['munge']['munge_version']}"
+  not_if "/usr/sbin/munged --version | grep -q munge-#{node['cfncluster']['munge']['munge_version']} && ls #{munge_libdir}/libmunge*"
```
tilne, line 53: what happens if an unexpected version of munge is installed in `munge_libdir`?

lukeseawalker: If unexpected means same version but not installed by us it should be ok.
If unexpected means different version, then the chef resource is triggered and the libraries are overridden

tilne (approved):

PR #749, opened 2020-09-28
Create packer file for Centos8

rexcsn:
* Create packer file for CentOS8 with similar logic to CentOS7
* Using dnf instead of yum as package manager everywhere

### DIFF
--- a/amis/packer_centos8.json
+++ b/amis/packer_centos8.json
@@ -1,0 +1,308 @@
+{
+  "variables" : {
+    "parallelcluster_version" : "",
+    "parallelcluster_cookbook_version" : "",
+    "chef_version" : "",
+    "chef_custom_install_url" : "{{env `CHEF_CUSTOM_INSTALL_URL`}}",
+    "berkshelf_version" : "",
+    "build_for" : "{{env `BUILD_FOR`}}",
+    "ami_perms" : "{{env `AMI_PERMS`}}",
+    "build_date" : "{{env `BUILD_DATE`}}",
+    "cookbook_path" : "{{env `COOKBOOK_PATH`}}",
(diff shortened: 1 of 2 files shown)


enrico-usai (approved):
LGTM.

We should test it locally before merging, we just need to verify that the packer process starts for now, then we'll fix the internal details.

amis/packer_centos8.json
```diff
@@ -0,0 +1,308 @@
...
+    },
+    {
+      "type" : "shell",
+      "inline" : [
+        "region=\"{{user `region`}}\"",
+        "bucket=\"s3.amazonaws.com\"",
+        "[[ ${region} =~ ^cn- ]] && bucket=\"s3.cn-north-1.amazonaws.com.cn/cn-north-1-aws-parallelcluster\"",
+        "curl --retry 3 -L -o /tmp/aws-cfn-bootstrap-latest.tar.gz https://${bucket}/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz",
+        "which pip2",
+        "if [ $? -eq 0 ]; then sudo pip2 install /tmp/aws-cfn-bootstrap-latest.tar.gz; else sudo pip install /tmp/aws-cfn-bootstrap-latest.tar.gz; fi"
```
enrico-usai, line 285: Probably we'll need to add symlinks for `cfn-init` and `cfn-signal` scripts:
https://computingforgeeks.com/how-to-install-aws-cloudformation-helper-scripts-on-centos/

enrico-usai (approved):

enrico-usai (commented):
We should fix the validation step, to be ready to support ARM:
`if [ "${_arch}" == "arm64" ] && [[ "${_os}" =~ ^centos[0-9]+ ]]; then`

PR #751, opened 2020-09-30
Add support for CentOS 8 in base_install and base_config logic

rexcsn:
This PR enables base_install/base_config to run on a bare CentOS 8 AMI.

The following changes were made:
* Enable PowerTools Repo so *-devel packages can be installed with DNF
* Install Python3 as system python for CentOS 8
* Remove obsolete logic of running nfs::server recipe when Ubuntu, nfs::server4 already includes nfs::server recipe
* Workaround to only run nfs::server instead of nfs::server4 for CentOS 8 due to issue: https://github.com/atomic-penguin/cookbook-nfs/issues/116
* Do not enforce kernel_devel version because kernel_devel package with same version as kernel release version cannot be found
* Modify logic to get EBS device to volume id mapping. Specifically ec2_dev_2_volid.py and parallelcluster-ebsnvme-id are modified for CentOS 8 to use nvme-cli to retrieve volume id for a device following this guide: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html#identify-nvme-ebs-device
* Install firewalld so iptables executable becomes available
* Enable EPEL repo by default

### DIFF
--- a/files/centos-8/ec2_dev_2_volid.py
+++ b/files/centos-8/ec2_dev_2_volid.py
@@ -1,0 +1,71 @@
+# Same as default/ec2_dev_2_volid.py except call to /usr/local/sbin/parallelcluster-ebsnvme-id
+import requests
+import sys
+import os
+import syslog
+import time
+import boto3
+import configparser
+from botocore.config import Config
+
+
+def main():
+    syslog.syslog("Starting ec2_dev_2_volid.py script")
+    # Get dev
+    try:
+        dev = str(sys.argv[1])
+        syslog.syslog("Input block device is %s" % dev)
+    except IndexError:
+        syslog.syslog(syslog.LOG_ERR, "Provide block device i.e. xvdf")
+
+    # Convert dev to mapping format
+    if 'nvme' in dev:
+        # For newer instances which expose EBS volumes as NVMe devices, translate the
+        # device name so boto can discover it.
+        output = os.popen('sudo /usr/local/sbin/parallelcluster-ebsnvme-id /dev/' + dev).read().split(":")[1].strip()
+        print(output)
+        sys.exit(0)
+    else:
+        dev = dev.replace('xvd', 'sd')
+        dev = '/dev/' + dev
+
+    # Get instance ID
+    instanceId = requests.get("http://169.254.169.254/latest/meta-data/instance-id").text
+
+    # Get region
+    region = requests.get("http://169.254.169.254/latest/meta-data/placement/availability-zone").text
+    region = region[:-1]
+
+    # Parse configuration file to read proxy settings
+    config = configparser.RawConfigParser()
+    config.read('/etc/boto.cfg')
+    proxy_config = Config()
+    if config.has_option('Boto', 'proxy') and config.has_option('Boto', 'proxy_port'):
+        proxy = config.get('Boto', 'proxy')
+        proxy_port = config.get('Boto', 'proxy_port')
+        proxy_config = Config(proxies={'https': "{0}:{1}".format(proxy, proxy_port)})
+
+    # Connect to AWS using boto
+    ec2 = boto3.client('ec2', region_name=region, config=proxy_config)
+
+    # Poll for blockdevicemapping
+    devices = ec2.describe_instance_attribute(InstanceId=instanceId, Attribute='blockDeviceMapping').get('BlockDeviceMappings')
+    devmap = dict((d.get('DeviceName'), d) for d in devices)
+    x = 0
+    while dev not in devmap:
+        if x == 36:
+            syslog.syslog("Dev %s did not appears in 180 seconds." % dev)
+            sys.exit(1)
+        syslog.syslog("Looking for dev %s in devmap %s" % (dev, devmap))
+        time.sleep(5)
+        devices = ec2.describe_instance_attribute(InstanceId=instanceId, Attribute='blockDeviceMapping').get('BlockDeviceMappings')
+        devmap = dict((d.get('DeviceName'), d) for d in devices)
+        x += 1
+
+    # Return volumeId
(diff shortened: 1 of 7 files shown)


recipes/base_install.rb
```diff
@@ -66,18 +70,18 @@
...
-  end
-when 'debian'
-  apt_package node['cfncluster']['kernel_generic_pkg'] do
-    retries 3
-    retry_delay 5
+package "install kernel packages" do
+  case node['platform_family']
+  when 'rhel', 'amazon'
+    package_name node['cfncluster']['kernel_devel_pkg']['name']
+    if node['platform_version'].to_i < 8
```
enrico-usai, line 77: as before, let's differentiate rhel and amazon case, because this platform_version is valid for rhel only.

recipes/base_install.rb
```diff
@@ -116,11 +120,12 @@
   mode '0644'
 end
 
-if node['platform'] == 'ubuntu' && node['platform_version'].to_f >= 16.04
-  # FIXME: https://github.com/atomic-penguin/cookbook-nfs/issues/93
```
enrico-usai, line 120: This issue is still opened: https://github.com/atomic-penguin/cookbook-nfs/issues/93

Are we sure we can this workaround?

recipes/base_install.rb
```diff
@@ -66,18 +70,18 @@
...
-  yum_package node['cfncluster']['kernel_devel_pkg']['name'] do
-    version node['cfncluster']['kernel_devel_pkg']['version']
-    retries 3
-    retry_delay 5
-  end
-when 'debian'
-  apt_package node['cfncluster']['kernel_generic_pkg'] do
-    retries 3
-    retry_delay 5
+package "install kernel packages" do
```
enrico-usai, line 73: It's an interesting change to mention in the commit message. 
Great idea to use `package` instead of the specific `yum_package` and `apt_package`.

Let's double check the entire code to see if there are other places to change.

recipes/base_install.rb
```diff
@@ -66,18 +70,18 @@
...
+  case node['platform_family']
+  when 'rhel', 'amazon'
+    package_name node['cfncluster']['kernel_devel_pkg']['name']
+    if node['platform_version'].to_i < 8
+      version node['cfncluster']['kernel_devel_pkg']['version']
```
enrico-usai, line 78: I'd also add a comment in the code to clarify this different behaviour for Centos8.

enrico-usai (changes requested):

files/centos-8/ec2_dev_2_volid.py
```diff
@@ -0,0 +1,70 @@
+import requests
```
enrico-usai, line 1: Could you add a comment describing where this file come from?
Which are the differences with the default one?

recipes/base_install.rb
```diff
@@ -217,15 +225,19 @@
 # Install NVIDIA and CUDA
 include_recipe "aws-parallelcluster::_nvidia_install"
 
-# Install FSx options
-include_recipe "aws-parallelcluster::_lustre_install"
+# FIXME:
+# # Install FSx options
+# include_recipe "aws-parallelcluster::_lustre_install"
```
enrico-usai, line 230: You can re-enable it, I merged the https://github.com/aws/aws-parallelcluster-cookbook/pull/753

recipes/base_install.rb
```diff
@@ -217,15 +225,19 @@
...
 
-# Install EFA & Intel MPI
-include_recipe "aws-parallelcluster::efa_install"
-include_recipe "aws-parallelcluster::intel_mpi"
+# FIXME:
+# # Install EFA & Intel MPI
+# include_recipe "aws-parallelcluster::efa_install"
```
enrico-usai, line 234: You can enable `efa_install` and `intel_mpi` once https://github.com/aws/aws-parallelcluster-cookbook/pull/754 is merged.

files/centos-8/ec2_dev_2_volid.py
rexcsn: Done

recipes/base_install.rb
rexcsn: Enabled.

recipes/base_install.rb
rexcsn: #754 is merged, enabled

recipes/base_install.rb
```diff
@@ -217,15 +225,19 @@
...
+# include_recipe "aws-parallelcluster::intel_mpi"
 
-# Install the AWS cloudwatch agent
-include_recipe "aws-parallelcluster::cloudwatch_agent_install"
+# FIXME:
+# # Install the AWS cloudwatch agent
+# include_recipe "aws-parallelcluster::cloudwatch_agent_install"
```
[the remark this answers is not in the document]
rexcsn: Enabled

recipes/base_install.rb
rexcsn: On a second look this still seems to be an issue. Keeping original workflow for Ubuntu and adding workaround logic to skip `nfs::server4` for CentOS 8

recipes/base_install.rb
```diff
@@ -26,12 +26,18 @@
     include_recipe "yum-epel"
   end
 
-
   unless node['platform_version'].to_i < 7
     execute 'yum-config-manager_skip_if_unavail' do
       command "yum-config-manager --setopt=\*.skip_if_unavailable=1 --save"
```
enrico-usai, line 31: should we set this flag for dnf too?

rexcsn: This command actually "just works" on CentOS 8, and should be equivalent to running `dnf config-manager`.
We could change it if we want but there should be no difference

PR #753, opened 2020-10-05
Cleanup initialization of FSx Lustre defaults and add CentOS 8 support

enrico-usai:
# Cleanup initialization of FSx Lustre default

## Changes

* Moved FSx attributes initialization out from main platform switch case.
* Removed CentOS 6 attributes: this OS is not supported.
* Renamed ubuntu specific `repository_uri` attribute to be aligned with other OSes.
* Refactored attributes initialization by using `value_for_platform` keywork.
* Differentiated attributes according to the version.
   The attributes `version`, `kmod_url` and `client_url` are only used for CentOS 7.5 and 7.6
   `public_key` and `base_url` are used for CentOS 7.7 and Ubuntu.
* Renamed `get_rhel_kernel_minor_version` given that it's CentOS 7 specific.

## References
* Value for platform: https://docs.chef.io/dsl_recipe/#value_for_platform

# Add FSx support for CentOS 8

* Use `package` in place of `yum_package` to support `dnf`.
* Added `gdisk` package, required by `update initramfs` action, called by `kernel_module 'lnet'` resource.
* Use `['platform_version'].to_f`  to compare minor version of the OS in place of `['platform_version'].split('.')[1].to_i`
  to be compatible with multiple major OS values.
* Explicitly added `x86_64` at the end of the `base_url` parameter, like we have for CentOS 7.

## Other changes
* Improved check for CentOS 7.5. and 7.6.
* Use `package` in place of `apt_package` to be aligned to other OSes.

## Tests
Verified installation:
```
$ rpm -qa | grep lustre-client
lustre-client-2.10.8-1.fsx6.el8.x86_64
kmod-lustre-client-2.10.8-1.fsx6.el8.x86_64
```

## References
* Lustre installation guide: https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
* gdisk issue: https://www.spinics.net/lists/centos-devel/msg18766.html
* package resource: https://docs.chef.io/resources/package/

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -196,12 +196,6 @@ when 'rhel', 'amazon'
                                                 blas-devel fftw-devel libffi-devel openssl-devel dkms mysql-devel libedit-devel
                                                 libical-devel postgresql-devel postgresql-server sendmail mdadm python python-pip
                                                 libgcrypt-devel glibc-static]
-
-    # Lustre Drivers for Centos 6
-    default['cfncluster']['lustre']['version'] = '2.10.6'
-    default['cfncluster']['lustre']['kmod_url'] = 'https://downloads.whamcloud.com/public/lustre/lustre-2.10.6/el6/client/RPMS/x86_64/kmod-lustre-client-2.10.6-1.el6.x86_64.rpm'
-    default['cfncluster']['lustre']['client_url'] = 'https://downloads.whamcloud.com/public/lustre/lustre-2.10.6/el6/client/RPMS/x86_64/lustre-client-2.10.6-1.el6.x86_64.rpm'
-
     if node['platform_version'].to_i >= 7
       default['cfncluster']['base_packages'] = %w[vim ksh tcsh zsh openssl-devel ncurses-devel pam-devel net-tools openmotif-devel
                                                   libXmu-devel hwloc-devel libdb-devel tcl-devel automake autoconf pyparted libtool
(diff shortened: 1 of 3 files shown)


attributes/default.rb
```diff
@@ -306,6 +288,38 @@
   end
 end
 
+# Lustre defaults (for CentOS >=7.7 and Ubuntu)
+default['cfncluster']['lustre']['public_key'] = value_for_platform(
+  'centos' => { '>=7.7' => "https://fsx-lustre-client-repo-public-keys.s3.amazonaws.com/fsx-rpm-public-key.asc" },
+  'ubuntu' => { :default => "https://fsx-lustre-client-repo-public-keys.s3.amazonaws.com/fsx-ubuntu-public-key.asc" }
```
rexcsn, line 294: Nice use of `value_for_platform `!
Just curious, is `:default` the same as `'default'`? From [docs here](https://docs.chef.io/dsl_recipe/#value_for_platform) I can't seem to find `:default` syntax

rexcsn (approved):

attributes/default.rb
enrico-usai: `:default` or `default:` are symbols. Using symbol is an [alternative way](https://launchschool.com/books/ruby/read/hashes) to define keys in hashes. See also [symbols vs strings](https://medium.com/@lcriswell/ruby-symbols-vs-strings-248842529fd9).

In this specific case I started using the symbol because Intellij was showing me a warning: `"String key in hash" --> "Convert to symbol key"`.

The bad is that `:default` is not accepted by rubocop (we're following [ruby19 syntax](https://www.rubydoc.info/gems/rubocop/0.46.0/RuboCop/Cop/Style/HashSyntax)) and mixing the two syntaxes is not a clear solution for me, so for now I'm going use the same syntax for all the keys, despite to IntelliJ warning.

See also: https://docs-archive.chef.io/release/11-18/ruby.html#symbols-or-strings

lukeseawalker (approved):

PR #757, opened 2020-10-06
Update logic to retrieve EBS device to volume id mapping

rexcsn:
* Following this guide in general: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html#identify-nvme-ebs-device
* Modify ec2_dev_2_volid.py and /usr/local/sbin/parallelcluster-ebsnvme-id under files/default/ to follow instruction outlined in the guide for "Other Linux AMIs". /usr/local/sbin/parallelcluster-ebsnvme-id now uses nvme-cli package to retrieve volume id for EBS device
* Modify ec2_dev_2_volid.py for alinux AMIs to use /sbin/ebsnvme-id to retrieve volume id for EBS device
* Keep existing logic for CentOS6 because logic to use nvme-cli only works on OS with kernel version >=4.2, which do not include CentOS6


rexcsn:
default `parallelcluster-ebsnvme-id` has a bug, closing

PR #759, opened 2020-10-07
Add CentOS 8 support for NICE DCV and attributes cleanup 

enrico-usai:
# Add CentOS 8 support for NICE DCV
* Disable Wayland, the default GNOME Display Manager for  CentOS 8, that is not supported by DCV
* Add default value for dcv_port

## Selinux notes
Default SELinux policies in RHEL8 can lead to failures in `xdm` processes using NVIDIA drivers/libraries,
thus for example the `gnome-shell` and the `dcv` system agent (being children of `gdm`) can be impacted.
By default SELinux is disabled so it doesn't impact our installation.
If SELinux is re-enabled, custom policies must be defined to grant `xdm_t` processes the permissions they need.

## Tests
* Verified gdm configuration file
* Started session and listed

## References
DCV guide: https://docs.aws.amazon.com/dcv/latest/adminguide/setting-up-installing-linux-server.html

# Refactor NICE DCV defaults attributes
* Use a global test for ARM instances, to avoid to repeat the check for each variable.
* Use value_for_platform to define defaults attributes.

# Add newlines between different features
* Move Munge default close to other Munge attributes.
* Add retries to Xorg package installation
* Rubocop fixes

## References:
* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IfUnlessModifier
* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/SelfAssignment

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -21,11 +21,13 @@ default['cfncluster']['sources_dir'] = "#{node['cfncluster']['base_dir']}/source
 default['cfncluster']['scripts_dir'] = "#{node['cfncluster']['base_dir']}/scripts"
 default['cfncluster']['license_dir'] = "#{node['cfncluster']['base_dir']}/licenses"
 default['cfncluster']['configs_dir'] = "#{node['cfncluster']['base_dir']}/configs"
+
 # Cluster config
 default['cfncluster']['cluster_s3_bucket'] = nil
 default['cfncluster']['cluster_config_s3_key'] = nil
 default['cfncluster']['cluster_config_version'] = nil
 default['cfncluster']['cluster_config_path'] = "#{node['cfncluster']['configs_dir']}/cluster_config.json"
+
 # Python Version
 default['cfncluster']['python-version'] = '3.6.9'
 default['cfncluster']['python-version-centos6'] = '2.7.17'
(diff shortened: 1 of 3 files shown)


tilne (approved):

attributes/default.rb
```diff
@@ -103,23 +103,26 @@
...
+default['cfncluster']['dcv_port'] = 8443
 default['cfncluster']['dcv']['installed'] = 'yes'
 default['cfncluster']['dcv']['version'] = '2020.1-9012'
 if arm_instance?
-  default['cfncluster']['dcv']['supported_os'] = %w[ubuntu18 amazon2]
+  default['cfncluster']['dcv']['supported_os'] = %w[centos8 ubuntu18 amazon2]
   default['cfncluster']['dcv']['url_architecture_id'] = 'aarch64'
   default['cfncluster']['dcv']['sha256sum'] = value_for_platform(
     'centos' => {
+      '>=8' => "eca626c10a6fe8b8770b1809555e6c1747a8ab7b6991e25e370d8d0c19902f4e",
```
rexcsn, line 114: Should we use "pessimistically greater than" operator `~>` here as well?
If we decide to do this we should change for all `value_for_platform ` calls

recipes/dcv_config.rb
```diff
@@ -20,10 +20,10 @@
 
 # Configure the system to enable NICE DCV to have direct access to the Linux server's GPU and enable GPU sharing.
 def allow_gpu_acceleration
-  # On CentOS 7 fix circular dependency multi-user.target -> cloud-init-> isolate multi-user.target.
+  # On CentOS >=7 fix circular dependency multi-user.target -> cloud-init-> isolate multi-user.target.
   # multi-user.target doesn't start until cloud-init run is finished. So isolate multi-user.target
   # is stuck into starting, which keep hanging chef until the 3600s timeout.
-  unless node['platform'] == 'centos' && node['platform_version'].to_i == 7
+  unless node['platform'] == 'centos' && node['platform_version'].to_f >= 7.0
```
rexcsn, line 26: Wondering if there is a reason we changed from using `to_i` to `to_f`

enrico-usai: You right, let me restore it.

rexcsn (approved):

recipes/dcv_config.rb
```diff
@@ -33,9 +33,7 @@ def allow_gpu_acceleration
   # Update the xorg.conf to set up NVIDIA drivers.
   # NOTE: --enable-all-gpus parameter is needed to support servers with more than one NVIDIA GPU.
   nvidia_xconfig_command = "nvidia-xconfig --preserve-busid --enable-all-gpus"
-  if node['ec2']['instance_type'].start_with?("g2.")
-    nvidia_xconfig_command = nvidia_xconfig_command + " --use-display-device=none"
-  end
+  nvidia_xconfig_command += " --use-display-device=none" if node['ec2']['instance_type'].start_with?("g2.")
```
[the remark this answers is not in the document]
enrico-usai: It is not part of my patch, I just converted the if to be one-liner. 

I tried to search in the [official guide](https://docs.aws.amazon.com/dcv/latest/adminguide/setting-up-installing-linux-prereq.html#linux-prereq-nvidia) with no luck. I think it is due to the specific graphic card [available on g2 instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#public-nvidia-driver).

PR #766, opened 2020-10-09
Improve function to retrieve Centos 7 kernel minor version

enrico-usai:
The function was in the `value_for_platform` method, so it was called in any case.

If the os was not rhel7 the `.match` method was returning a `nul` value so the `match(...)[1]` was failing.

With this commit we're adding a default for all the oses in the `value_for_platform` selector and then calling the function only if needed.

In any case we're also adding robustness to the function to avoid to fail if the match is failing.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -375,6 +375,9 @@ def chrony_reload_command
     chrony_reload_command = "service #{node['cfncluster']['chrony']['service']} force-reload"
   elsif node['init_package'] == 'systemd'
     chrony_reload_command = "systemctl force-reload #{node['cfncluster']['chrony']['service']}"
+  else
+    raise "Init package #{node['init_package']} not supported."
   end
+
   chrony_reload_command
 end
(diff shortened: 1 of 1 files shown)


attributes/default.rb
```diff
@@ -339,10 +339,14 @@
 default['cfncluster']['lustre']['base_url'] = value_for_platform(
   'centos' => {
     '>=8' => "https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/x86_64/",
-    'default' => "https://fsx-lustre-client-repo.s3.amazonaws.com/el/7.#{get_rhel7_kernel_minor_version}/x86_64/"
+    '7.8' => "https://fsx-lustre-client-repo.s3.amazonaws.com/el/7.8/x86_64/",
```
lukeseawalker, line 342: isn't better leave the default as it was and move the check with the `if node['platform'] == 'centos'` inside the get_rhel7_kernel_minor_version function?

enrico-usai: Done

lukeseawalker (approved):

PR #781, opened 2020-10-26
Centos 8: fix Intel HPC cache issue by setting `keepcache=True`

rexcsn:
* Set `keepcache=True` so downloaded packages are not removed after successful installation of any package
* Add retries to Intel install


recipes/intel_install.rb
```diff
@@ -17,6 +17,20 @@
 
 return unless node['conditions']['intel_hpc_platform_supported'] && node['cfncluster']['enable_intel_hpc_platform'] == 'true'
 
+# Enforce keepcache=True for dnf, else downloaded packages will be removed after any successful install
+bash "configure dnf caching" do
+  cwd node['cfncluster']['sources_dir']
+  code <<-DNF_CACHE
```
enrico-usai, line 23: Let's use [`replace_or_add`](https://github.com/sous-chefs/line/blob/master/documentation/resources/replace_or_add.md) resource:
```ruby
replace_or_add "configure dnf caching" do
  path "/etc/dnf/dnf.conf"
  pattern "keepcache.*"
  line "keepcache=True"
end
```

This resource already has `ignore_missing` set to `true` by default so it's safe for non Centos8 systems too.

recipes/intel_install.rb
```diff
@@ -68,12 +88,14 @@
...
 # Installs intel-hpc-platform rpms
+# Install from /opt/intel/rpms, with -C option
+# If -C is not specified, cached packages in /opt/intel/rpms might be removed
 # This rpm installs a file /etc/intel-hpc-platform-release that contains the INTEL_HPC_PLATFORM_VERSION
 bash "install intel hpc platform" do
   cwd node['cfncluster']['sources_dir']
   code <<-INTEL
     set -e
-    yum install -y /opt/intel/rpms/*
+    yum install -C -y /opt/intel/rpms/*
```
enrico-usai, line 98: I'd use `--cacheonly` instead of `-C` that is more explicit.

enrico-usai (approved):
👍

PR #787, opened 2020-10-28
CentOS 8: Fix bug in parallelcluster-ebsnvme-id

rexcsn:
* Fix bug introduced in 2c0de640eda12248c606bdf8428923ef866c6c5a; centos8 specific `parallelcluster-ebsnvme-id` does not accept option and was unable to print out volume device info when called from `attachVolume.py`, causing mounting of multiple EBS to fail
* `parallelcluster-ebsnvme-id` needs to accept the options `-v/-b/-u` to output volume id and block device information when called from `ec2_dev_2_volid.py` and `attachVolume.py`
* Modify centos8 specific `parallelcluster-ebsnvme-id` to output correct info based on option specified
* Centos8 specific `ec2_dev_2_volid.py` no longer needed and removed, as new `parallelcluster-ebsnvme-id` script will accept `-v` option to output volume id


enrico-usai (approved):
Great job!

PR #790, opened 2020-10-30
Deprecate CentOS6 and apply rubocop fixes

enrico-usai:
# Deprecate CentOS6

Removed:
* packer files, scripts used to build the base ami and from `build_ami.sh` script
* `platform_supports_impi`, `platform_supports_lustre_on_x86_64` and `platform_supports_pmix` helpers
* specific default attributes

Removed from:
* conditions
* comments
* kitchen tests

# Rubocop fixes
* Before: 152 offenses detected
* After: 111 offenses detected


.kitchen.yml
```diff
@@ -6,7 +6,6 @@ provisioner:
   name: chef_zero
 
 platforms:
```
lukeseawalker, line 8: I think we can completely remove the "platform" here

amis/build_ami.sh
```diff
@@ -184,18 +184,15 @@ check_options() {
     esac
 
     # Ensure the specified architecture-OS combination is valid
-    if [ "${_arch}" == "arm64" ] && [[ "${_os}" =~ ^centos[6-7]$ ]]; then
+    if [ "${_arch}" == "arm64" ] && [[ "${_os}" =~ ^(centos7|alinux)$ ]]; then
```
lukeseawalker, line 187: you are going to match also `alinux2` which I don't think is wanted

enrico-usai: Alinux2 should be excluded because there are `^` and `$` sorrounding characters, do you agree?

lukeseawalker: you are right, I have missed the `$`

lukeseawalker (approved):

PR #796, opened 2020-11-04
Cleanup and refactoring of defaults and helpers file

enrico-usai:
# Remove default AMI IDs from kitchen tests 

The values are old ones, never updated and valid only for `us-east-1`.

The `*_IMAGE_ID` parameter is now mandatory but it's not a breaking change because we're always setting a value when running kitchen tests, retrieving the AMI to use with an `ec2 describe-images` call in the specific region.

# Cleanup FSx Lustre defaults and helpers file

## FSx Lustre defaults Changes

* Moved FSx attributes initialization out from main platform switch case.
* Removed CentOS 6 attributes: this OS is not supported.
* Renamed ubuntu specific `repository_uri` attribute to be aligned with other OSes.
* Refactored attributes initialization by using `value_for_platform` keywork.
* Differentiated attributes according to the version.
   The attributes `version`, `kmod_url` and `client_url` are only used for CentOS 7.5 and 7.6 `public_key` and `base_url` are used for CentOS 7.7 and Ubuntu.
* Renamed `get_rhel_kernel_minor_version` given that it's CentOS 7 specific.

## References
* Value for platform: https://docs.chef.io/dsl_recipe/#value_for_platform
* FSx client installation: https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html

## Improve function to retrieve Centos 7 kernel minor version

The function is in the `value_for_platform` method, so it is called in any case.
If the os was not rhel7 the `.match` method was returning a `nul` value so the `match(...)[1]` was failing.
With this commit we're using '7' as a default and calling the function only if needed.
In any case we're also adding robustness to the function to raise an error if the match fails.

## Improve function to retrieve Chrony reload command

The chrony_reload_command variable could be initialized.
Now we're raising an error if the init package method is an unsupported one.

# Refactor NICE DCV defaults attributes and separate features

Use a global test for ARM instances, to avoid to repeat the check for each variable.
Use value_for_platform to define defaults attributes.

Add newlines between different features

* Move Munge default close to other Munge attributes.
* Add retries to Xorg package installation
* Rubocop fixes

## References:
* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IfUnlessModifier
* https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/SelfAssignment

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -21,11 +21,13 @@ default['cfncluster']['sources_dir'] = "#{node['cfncluster']['base_dir']}/source
 default['cfncluster']['scripts_dir'] = "#{node['cfncluster']['base_dir']}/scripts"
 default['cfncluster']['license_dir'] = "#{node['cfncluster']['base_dir']}/licenses"
 default['cfncluster']['configs_dir'] = "#{node['cfncluster']['base_dir']}/configs"
+
 # Cluster config
 default['cfncluster']['cluster_s3_bucket'] = nil
 default['cfncluster']['cluster_config_s3_key'] = nil
 default['cfncluster']['cluster_config_version'] = nil
 default['cfncluster']['cluster_config_path'] = "#{node['cfncluster']['configs_dir']}/cluster_config.json"
+
 # Python Version
 default['cfncluster']['python-version'] = '3.6.9'
 default['cfncluster']['python-version-centos6'] = '2.7.17'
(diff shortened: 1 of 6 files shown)


recipes/_lustre_install.rb
```diff
@@ -68,12 +68,15 @@
...
+
 elsif node['platform'] == 'centos'
+  # Centos 6
   Chef::Log.warn("Unsupported version of Centos, #{node['platform_version']}, supported versions are >= 7.5")
+
 elsif node['platform'] == 'ubuntu'
 
   apt_repository 'fsxlustreclientrepo' do
-    uri          node['cfncluster']['lustre']['repository_uri']
+    uri          node['cfncluster']['lustre']['base_url']
```
demartinofra, line 79: wasn't this the actual repository_uri? why changing to base_url?

attributes/default.rb
```diff
@@ -89,51 +99,67 @@
...
-  default['cfncluster']['dcv']['package'] = "nice-dcv-#{node['cfncluster']['dcv']['version']}-ubuntu1804-#{node['cfncluster']['dcv']['url_architecture_id']}"
-  default['cfncluster']['dcv']['server'] = "nice-dcv-server_2020.1.9012-1_#{node['cfncluster']['dcv']['package_architecture_id']}.ubuntu1804.deb" # NICE DCV server package
-  default['cfncluster']['dcv']['xdcv'] = "nice-xdcv_2020.1.338-1_#{node['cfncluster']['dcv']['package_architecture_id']}.ubuntu1804.deb"  # required to create virtual sessions
-  default['cfncluster']['dcv']['gl'] = "nice-dcv-gl_2020.1.840-1_#{node['cfncluster']['dcv']['package_architecture_id']}.ubuntu1804.deb"  # required to enable GPU sharing
-  default['cfncluster']['dcv']['sha256sum'] = if arm_instance?
-                                                "0cd0512d57808cee0c48d0817a515825f9c512cb9d70e5672fecbb7450b729b3"
-                                              else
-                                                "7569c95465743b512f1ab191e58ea09777353b401c1ec130ee8ea344e00f8900"
-                                              end
+  default['cfncluster']['dcv']['package_architecture_id'] = arm_instance? ? 'arm64' : 'amd64'
```
demartinofra, line 133: why don't we move also this in the if/else above by assigning only to ubuntu 18.04 in value_for_platform?

demartinofra (approved):
Thanks for cleaning things up. I left two minor optional comments.

recipes/_lustre_install.rb
enrico-usai: > Renamed ubuntu specific repository_uri attribute to be aligned with other OSes.

`apt_repository / uri` has the same meaning of `yum_repository / baseurl` and I wanted to avoid using a different variable for this os only and be able to initialize it with a single `value_for_platform` assignment, instead of having an if/else block on which the variable changes name.

attributes/default.rb
enrico-usai: I decided to keep this into an if block because this variable is very specific fo ubuntu18, I agree we can do this change. We'll do in a future patch.

PR #799, opened 2020-11-05
Explicitly install php-xml package required for Ganglia

enrico-usai:
In Centos8 Ganglia execution was failing because unable to find `xml_parser_create`.

The error in `/var/log/php-fpm/www-error.log` was:
```
PHP Fatal error:  Uncaught Error: Call to undefined function xml_parser_create()
in /usr/share/ganglia/ganglia.php:356
```

With this patch I'm also updating the gmond template with the updated info.

### DIFF
--- a/recipes/ganglia_install.rb
+++ b/recipes/ganglia_install.rb
@@ -19,7 +19,7 @@ case node['cfncluster']['cfn_node_type']
 when 'MasterServer', nil
   case node['platform']
   when "redhat", "centos", "amazon", "scientific" # ~FC024
-    package %w[ganglia ganglia-gmond ganglia-gmetad ganglia-web httpd php php-gd rrdtool] do
+    package %w[ganglia ganglia-gmond ganglia-gmetad ganglia-web httpd php php-gd php-xml rrdtool] do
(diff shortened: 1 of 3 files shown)


recipes/ganglia_install.rb
```diff
@@ -19,7 +19,7 @@
 when 'MasterServer', nil
   case node['platform']
   when "redhat", "centos", "amazon", "scientific" # ~FC024
-    package %w[ganglia ganglia-gmond ganglia-gmetad ganglia-web httpd php php-gd rrdtool] do
+    package %w[ganglia ganglia-gmond ganglia-gmetad ganglia-web httpd php php-gd php-xml rrdtool] do
```
lukeseawalker, line 22: Is this `php-xml ` package name consistent among all the supported OSs under this platform condition?

lukeseawalker (approved):

recipes/ganglia_install.rb
enrico-usai: I verified Alinux and Alinux2 and the package is available and I ran kitchen tests for all the OSes

PR #800, opened 2020-11-05
Upgrade EFA to 1.10.1 to support Centos8

enrico-usai:
Add Centos8 as supported OS for EFA.
The `efa_supported` condition is no longer needed because all the OSes are supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,23 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 
 **CHANGES**
 
+- Upgrade EFA installer to version 1.10.1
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 
 **CHANGES**
 
+- Upgrade EFA installer to version 1.10.1
```
lukeseawalker, line 16: probably it's worth to list the components even with no changes in version

enrico-usai: Done

demartinofra (approved):

PR #801, opened 2020-11-06
Configure NFS threads

rexcsn:
* Run nfs::server to in base_install to install NFS packages at AMI build time
* Run NFS cookbook in base_config to setup NFS threads to be max(8, num_cores) for performance
* Overwrite NFS server config for alinux/alinux2/centos8 because these OSs are not/incorrectly supported by NFS cookbook


recipes/_nfs_config.rb
```diff
@@ -0,0 +1,55 @@
...
+  'default' => '/etc/sysconfig/nfs'
+)
+
+# For performance, set NFS threads to max(num_cores, 8); note cores, not vcpus
+ruby_block "set nfs threads" do
+  block do
+    require 'chef/mixin/shell_out'
+    cores_per_socket = shell_out!('lscpu | grep -oP "^Core\(s\) per socket:\s*\K(.+)$"').stdout.strip.to_i
+    sockets = shell_out!('lscpu | grep -oP "^Socket\(s\):\s*\K(.+)$"').stdout.strip.to_i
+    total_cores = cores_per_socket * sockets
```
demartinofra, line 33: you might simply use `node['cpu']['cores']`

recipes/_nfs_config.rb
```diff
@@ -0,0 +1,55 @@
...
+if node['platform'] == 'ubuntu' && node['platform_version'].to_f >= 16.04
+  # FIXME: https://github.com/atomic-penguin/cookbook-nfs/issues/93
+  include_recipe "nfs::server"
+end
+include_recipe "nfs::server4"
+
+# Extend NFS cookbook logic to OS not officially supported
+# By overwriting templates for node['nfs']['config']['server_template'] used by NFS cookbook
+# When running, NFS cookbook will use nfs.conf.erb templates provided in this cookbook to generate server_template
+edit_resource(:template, node['nfs']['config']['server_template']) do
```
demartinofra, line 52: don't we want to override only the missing/broken templates?

recipes/_nfs_config.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+#
+# Cookbook Name:: aws-parallelcluster
+# Recipe:: _nfs_config
```
demartinofra, line 5: let's drop the leading underscore. We are going to remove it from all recipes soon

recipes/base_config.rb
```diff
@@ -17,6 +17,8 @@
 
 include_recipe 'aws-parallelcluster::base_install'
 
+include_recipe 'aws-parallelcluster::_nfs_config'
```
demartinofra, line 20: let's make sure nfs is installed at ami build time and only configured at runtime

templates/centos-8/nfs.conf.erb
```diff
@@ -0,0 +1,19 @@
+# Generated by Chef for <%= node['fqdn'] %>
```
demartinofra, line 1: how did we come up with this config? can you add some details?

demartinofra: I can see there are some differences with the default config for example in teh fact that lockd only has the udp port configured

recipes/base_config.rb
rexcsn: Ok, including `nfs::server` in `base_install` so packages are installed at build time

recipes/_nfs_config.rb
rexcsn: Sure, added logic to only overwrite for alinux OSs and centos8

templates/centos-8/nfs.conf.erb
rexcsn: I was trying to translate all options set in templates used by NFS cookbook into options in `nfs.conf`

demartinofra: you mean from 
```
<% when 'rhel' -%>
# Rendered RHEL template variant
STATD_PORT="<%= node['nfs']['port']['statd'] -%>"
STATD_OUTGOING_PORT="<%= node['nfs']['port']['statd_out'] -%>"
STATDARG="-p <%= node['nfs']['port']['statd'] -%> -o <%= node['nfs']['port']['statd_out'] -%>"
MOUNTD_PORT="<%= node['nfs']['port']['mountd'] -%>"
RPCMOUNTDOPTS="-p <%= node['nfs']['port']['mountd'] -%>"
LOCKD_UDPPORT="<%= node['nfs']['port']['lockd'] -%>"
LOCKD_TCPPORT="<%= node['nfs']['port']['lockd'] -%>"
RQUOTAD_PORT="<%= node['nfs']['port']['rquotad'] -%>"
  <% unless node['nfs']['v2'].nil? -%>
MOUNTD_NFS_V2="<%= node['nfs']['v2'] -%>"
  <% end -%>
  <% unless node['nfs']['v3'].nil? -%>
MOUNTD_NFS_V3="<%= node['nfs']['v3'] -%>"
  <% end -%>
RQUOTAD="<%= node['nfs']['rquotad'] -%>"
  <% unless node['nfs']['threads'] == 0 -%>
RPCNFSDCOUNT="<%= node['nfs']['threads'] -%>"
  <% end -%>
```

Can you please add some details on why you excluded some of them (such as `LOCKD_TCPPORT`  `RQUOTAD_PORT` `RQUOTAD`)?

rexcsn: Setting `[lockd] - port` should be equivalent to setting both `LOCKD_TCPPORT` and `LOCKD_UDPPORT` from old config
From [doc of `nfs.conf`](https://man7.org/linux/man-pages/man5/nfs.conf.5.html) and [doc of `rpc.statd`](https://man7.org/linux/man-pages/man8/rpc.statd.8.html):
> lockd  Recognized values: port and udp-port.

>-T, --nlm-port port
              Specifies the port number that lockd should listen on for NLM
              requests.  This sets both the TCP and UDP ports unless the UDP
              port is set separately.
...
The values recognized in the [lockd] section include port and udp-port which have the same effect as the --nlm-port and --nlm-udp-port options, respectively.

There is no option to configure `rquotad` in `nfs.conf` anymore.
From [this guide](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/exporting-nfs-shares_deploying-different-types-of-servers) it seems there is still a way to define the `rquotad` port number:
> You can override the default port number by appending -p port-number to the RPCRQUOTADOPTS variable in the /etc/sysconfig/rpc-rquotad file.

But I would argue this is not needed, since nfs recipe seems to [disable `rquotad` by default](https://github.com/atomic-penguin/cookbook-nfs/blob/master/attributes/default.rb#L27)

recipes/base_install.rb
```diff
@@ -123,15 +123,8 @@
...
 
-if node['platform'] == 'ubuntu' && node['platform_version'].to_f >= 16.04
-  # FIXME: https://github.com/atomic-penguin/cookbook-nfs/issues/93
-  include_recipe "nfs::server"
-end
-if node['platform'] == 'centos' && node['platform_version'].to_i == 8
-  # Workaround for issue: https://github.com/atomic-penguin/cookbook-nfs/issues/116
-  node.force_override['nfs']['service']['idmap'] = 'nfs-idmapd'
-end
-include_recipe "nfs::server4"
```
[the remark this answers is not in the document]
demartinofra: ok then let's also double check with chef logs that at runtime we don't perform any extra operation that can be moved in the install recipes.

demartinofra (approved):
LGTM once we confirm with chef logs that no extra steps are performed ad config time other than producing nfsd configuration

libraries/helpers.rb
```diff
@@ -421,3 +421,13 @@ def efa_gdr_enabled?
   end
   (config_value == enabling_value || config_value == "cluster") && graphic_instance?
 end
+
+# CentOS8 and alinux OSs currently not correctly supported by NFS cookbook
+# Overwriting templates for node['nfs']['config']['server_template'] used by NFS cookbook for these OSs
+# When running, NFS cookbook will use nfs.conf.erb templates provided in this cookbook to generate server_template
+def overwrite_nfs_template?
```
demartinofra, line 428: minor: theoretically we can also get rid of this and move the condition in the conditions rather than calling the function

rexcsn:
Manually tested with AL2 that the only actions at config time are producing NFS related configs and restarting NFS related daemons.

PR #808, opened 2020-11-12
Arm performance library support

yuleiwan:
**Changes**
* Add arm performance library installation and module configuration
* Add gcc-9.3 module configuration and create modulefile
* Currently support alinux2, ubuntu1804 and centos8 arm instance
* Arm performance library kitchen test

**Version Update**
we could simply set  `default['cfncluster']['armpl']['version']`(e.g. 20.2.1 or 20.3), `default['cfncluster']['armpl']['gcc']['major_minor_version']` values to new version, then the value will propagate.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -68,6 +68,23 @@ default['cfncluster']['intelmpi']['version'] = '2019.7.217'
 default['cfncluster']['intelmpi']['modulefile'] = "/opt/intel/impi/#{node['cfncluster']['intelmpi']['version']}/intel64/modulefiles/mpi"
 default['cfncluster']['intelmpi']['kitchen_test_string'] = 'Version 2019 Update 7'
 
+# Arm Performance Library
+default['cfncluster']['armpl']['version'] = '20.2.1'
+default['cfncluster']['armpl']['gcc']['major_minor_version'] = '9.3'
+default['cfncluster']['armpl']['gcc']['patch_version'] = '0'
+default['cfncluster']['armpl']['gcc']['url'] = "https://ftp.gnu.org/gnu/gcc/gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.#{node['cfncluster']['armpl']['gcc']['patch_version']}/gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.#{node['cfncluster']['armpl']['gcc']['patch_version']}.tar.gz"
+default['cfncluster']['armpl']['platform'] = value_for_platform(
+    'centos' => { '~>8' => 'RHEL-8' },
+    'amazon' => { '2' => 'RHEL-8' },
+    'ubuntu' => { '18.04' => 'Ubuntu-16.04' }
+)
+
+default['cfncluster']['armpl']['url'] = value_for_platform(
+    'centos' => { '~>8' => "https://aws-parallelcluster-dev-commercial-dev.s3.amazonaws.com/archives/armpl/RHEL-8/arm-performance-libraries_#{node['cfncluster']['armpl']['version']}_#{node['cfncluster']['armpl']['platform']}_gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.tar" },
+    'amazon' => { '2' => "https://aws-parallelcluster-dev-commercial-dev.s3.amazonaws.com/archives/armpl/RHEL-8/arm-performance-libraries_#{node['cfncluster']['armpl']['version']}_#{node['cfncluster']['armpl']['platform']}_gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.tar" },
+    'ubuntu' => { '18.04' => "https://aws-parallelcluster-dev-commercial-dev.s3.amazonaws.com/archives/armpl/Ubuntu-16.04/arm-performance-libraries_#{node['cfncluster']['armpl']['version']}_#{node['cfncluster']['armpl']['platform']}_gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.tar" }
+)
+
 # Python packages
 default['cfncluster']['cfncluster-version'] = '2.10.1'
 default['cfncluster']['cfncluster-cookbook-version'] = '2.10.1'
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -78,11 +78,10 @@ default['cfncluster']['armpl']['platform'] = value_for_platform(
     'amazon' => { '2' => 'RHEL-8' },
     'ubuntu' => { '18.04' => 'Ubuntu-16.04' }
 )
-
 default['cfncluster']['armpl']['url'] = value_for_platform(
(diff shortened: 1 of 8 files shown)


attributes/default.rb
```diff
@@ -58,6 +58,18 @@
 default['cfncluster']['intelmpi']['version'] = '2019.7.217'
 default['cfncluster']['intelmpi']['modulefile'] = "/opt/intel/impi/#{node['cfncluster']['intelmpi']['version']}/intel64/modulefiles/mpi"
 
+# Arm Performance Library
+default['cfncluster']['armpl']['version'] = '20.2.1'
+default['cfncluster']['armpl']['gcc']['version'] = '9.3'
+default['cfncluster']['armpl']['gcc']['url'] = "http://ftp.gnu.org/gnu/gcc/gcc-#{node['cfncluster']['armpl']['gcc']['version']}.0/gcc-#{node['cfncluster']['armpl']['gcc']['version']}.0.tar.gz"
```
tilne, line 64: It looks like the two version numbers defined above are only used in installation. If we ever wanted to update either of these versions we'd have to change tests.rb and all of the modulefiles. Instead can these variables be used in tests.rb, and can the modulefiles be turned into templates? That way upgrading a version is as easy as changing a single line (at least in the ideal case).

tilne:
can you add a changelog entry?

attributes/default.rb
```diff
@@ -58,6 +58,18 @@
...
+if node['platform'] == 'ubuntu'
+  default['cfncluster']['armpl']['platform'] = 'Ubuntu-16.04'
+  default['cfncluster']['armpl']['url'] = "https://developer.arm.com/-/media/Files/downloads/hpc/arm-performance-libraries/20-2-1/Ubuntu16.04/arm-performance-libraries_20.2.1_Ubuntu-16.04_gcc-#{node['cfncluster']['armpl']['gcc']['version']}.tar"
+else
+  default['cfncluster']['armpl']['platform'] = 'RHEL-8'
+  default['cfncluster']['armpl']['url'] = "https://developer.arm.com/-/media/Files/downloads/hpc/arm-performance-libraries/20-2-1/RHEL8/arm-performance-libraries_20.2.1_RHEL-8_gcc-#{node['cfncluster']['armpl']['gcc']['version']}.tar"
```
mauri-melato, line 70: Just to confirm, is the package for RHEL8 also valid for AL2?

yuleiwan: yes, this package is valid for AL2

attributes/default.rb
yuleiwan: Done

templates/default/gcc_modulefile.erb
```diff
@@ -0,0 +1,44 @@
+#%Module1.0
+
+## Required internal variables
+set name     gcc
+set version  <%= node['cfncluster']['armpl']['gcc']['version'] %>.0
```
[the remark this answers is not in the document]
tilne: I followed up with Yulei offline about this. We hardcode this patch number here and in the installer URL. Ideally we could add another parameter in the default attributes for this, because if we did ever want to use a patch update other than `.0` it seems likely that this change would be forgotten.

attributes/default.rb
yuleiwan: As we discussed offline, I keep the url until S3 distribution.

templates/default/gcc_modulefile.erb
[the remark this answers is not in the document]
yuleiwan: Done, use major_minor_version and patches_version

tilne (approved):

attributes/default.rb
```diff
@@ -58,6 +58,19 @@
 default['cfncluster']['intelmpi']['version'] = '2019.7.217'
 default['cfncluster']['intelmpi']['modulefile'] = "/opt/intel/impi/#{node['cfncluster']['intelmpi']['version']}/intel64/modulefiles/mpi"
 
+# Arm Performance Library
+default['cfncluster']['armpl']['version'] = '20.2.1'
+default['cfncluster']['armpl']['gcc']['major_minor_version'] = '9.3'
+default['cfncluster']['armpl']['gcc']['patch_version'] = '0'
+default['cfncluster']['armpl']['gcc']['url'] = "http://ftp.gnu.org/gnu/gcc/gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.#{node['cfncluster']['armpl']['gcc']['patch_version']}/gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.#{node['cfncluster']['armpl']['gcc']['patch_version']}.tar.gz"
```
demartinofra, line 65: can we use https?

attributes/default.rb
```diff
@@ -58,6 +58,19 @@
 default['cfncluster']['intelmpi']['version'] = '2019.7.217'
 default['cfncluster']['intelmpi']['modulefile'] = "/opt/intel/impi/#{node['cfncluster']['intelmpi']['version']}/intel64/modulefiles/mpi"
 
+# Arm Performance Library
+default['cfncluster']['armpl']['version'] = '20.2.1'
+default['cfncluster']['armpl']['gcc']['major_minor_version'] = '9.3'
+default['cfncluster']['armpl']['gcc']['patch_version'] = '0'
```
demartinofra, line 64: it looks like most of the times you need a concatenation between major/minor and patch version. I'd rather have a single attribute for the full version and extract the patch with a split when necessary. This way it's also easier to see the full used version in a single string.

attributes/default.rb
```diff
@@ -58,6 +58,19 @@
...
+default['cfncluster']['armpl']['gcc']['url'] = "http://ftp.gnu.org/gnu/gcc/gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.#{node['cfncluster']['armpl']['gcc']['patch_version']}/gcc-#{node['cfncluster']['armpl']['gcc']['major_minor_version']}.#{node['cfncluster']['armpl']['gcc']['patch_version']}.tar.gz"
+if node['platform'] == 'ubuntu'
```
demartinofra, line 66: optional: I think Enrico did some work to replace all these if statements with `value_for_platform`

recipes/arm_pl.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+#
+# Cookbook Name:: aws-parallelcluster
+# Recipe:: arm_pl
```
enrico-usai, line 5: I'd call the recipe: arm_pl_install.rb, in this way it's easier to distinguish recipes executed at install and config time.

attributes/default.rb
yuleiwan: done

attributes/default.rb
yuleiwan: done

recipes/arm_pl.rb
yuleiwan: done

attributes/default.rb
yuleiwan: I am thinking if we can split string in erb file or not. We need to pass 9.3 and 9.3.0 to two erb files, if we use full version like  `default['cfncluster']['armpl']['gcc']['version'] = 9.3.0`, do you know how to deal with the erb files?
Or we can use two versions in default file, like `default['cfncluster']['armpl']['gcc']['gcc_version'] = 9.3.0`, `default['cfncluster']['armpl']['gcc']['armpl_version'] = 9.3`.

tilne (dismissed):

tilne (approved):

PR #827, opened 2020-12-01
Fix routing with multiple Network Interfaces in Ubuntu 1804

ddeidda:
When private IP addresses are assigned to network interfaces, kernel creates automatic routes to the subnet. Since all multiple Network Interfaces are on the same subnet and these routes have all the same metrics (0 by default), the order by which they are declared determines which Network Interface will be used to reach other nodes on the same subnet.
In Ubuntu 18 the order of these routes is reverted from that of the devices and this makes a secondary interface being selected to connect to other nodes in the subnet. This breaks the contract by which compute nodes are expected to work in ParallelCluster and can lead to unexpected failures, for instance when MPI jobs are run with Ubuntu 18 and SGE scheduler. This in turn is due to the way SGE resolves the hostnames starting from node IPs: since compute nodes can be contacted from secondary Network Interfaces of other nodes the resolved hostname won't match with the real one asigned to the node, leading to errors like this:
```
error: commlib error: access denied (client IP resolved to host name "ip-172-31-45-139.us-west-2.compute.internal". This is not identical to clients host name "ip-172-31-34-150.us-west-2.compute.internal")
```

This commit fixes this issue by ensuring that the primary network interface is selected to contact other nodes.

### DIFF
--- a/files/ubuntu-18.04/configure_nw_interface.sh
+++ b/files/ubuntu-18.04/configure_nw_interface.sh
@@ -14,7 +14,8 @@ if
   [ -z "${GW_IP_ADDRESS}" ] ||        # gateway ip address
   [ -z "${DEVICE_IP_ADDRESS}" ] ||    # ip address to assign to the interface
   [ -z "${CIDR_PREFIX_LENGTH}" ] ||   # the prefix length of the device IP cidr block
-  [ -z "${NETMASK}" ]                 # netmask to apply to device ip address
+  [ -z "${NETMASK}" ] ||              # netmask to apply to device ip address
+  [ -z "${CIDR_BLOCK}" ]              # (full) subnet cidr block
 then
   echo 'One or more environment variables missing'
   exit 1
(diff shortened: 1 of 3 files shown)


files/ubuntu-18.04/configure_nw_interface.sh
```diff
@@ -57,4 +58,22 @@ $STATIC_IP_CONFIG
...
+# Hook to delete automatic routes to subnet automatically created from kernel.
+# This is needed in Ubuntu 18 because they are generated in the reverse order in which the network interfaces are
+# attached and this prevents the primary Network Interface from being selected to communicate in the subnet
+FILE="/etc/networkd-dispatcher/routable.d/cleanup-routes.sh"
+
+if [ ! -f "$FILE" ]; then
+/bin/cat <<EOF >${FILE}
+#!/bin/sh
+
+logger -t parallelcluster "Removing Automatic route for Interface: \${IFACE}"
```
lukeseawalker, line 72: where `IFACE` is defined?

files/ubuntu-18.04/configure_nw_interface.sh
```diff
@@ -57,4 +58,22 @@ $STATIC_IP_CONFIG
...
+ip route del ${CIDR_BLOCK} dev \${IFACE}
```
lukeseawalker, line 73: Is this change persistent to reboot?

files/ubuntu-18.04/configure_nw_interface.sh
ddeidda: It's an environment variable passed from [networkd-dispatcher](https://gitlab.com/craftyguy/networkd-dispatcher).

files/ubuntu-18.04/configure_nw_interface.sh
ddeidda: The route deletion is not persistent by itself, but the hook gets re-executed at boot.

lukeseawalker (approved):

PR #831, opened 2020-12-07
CentOS 8 PowerTools repo needs to be lowercase in newer kernel versions

enrico-usai:
We're enabling `PowerTools` or `powertools` according to the kernel version:
https://access.redhat.com/articles/3078#RHEL8

The name changed with 8.3 release:
https://bugzilla.redhat.com/show_bug.cgi?id=1900785

## Additional changes

I changed the `find_rhel_minor_version` to return just a number, without the `.` character to be easier to compare it.

With this patch I'm also removing `epel-release` installation which is already performed as a packer step.

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -32,10 +32,10 @@ when 'rhel', 'amazon'
     end
   end
   if node['platform'] == 'centos' && node['platform_version'].to_i == 8
-    # Enable PowerTools Repo so *-devel packages can be installed with DNF
-    # Enable EPEL repos
-    execute 'dnf enable powertools and EPEL repos' do
-      command "dnf config-manager --set-enabled PowerTools && dnf install -y epel-release"
+    # Enable powertools repo so *-devel packages can be installed with DNF
+    powertools_repo = find_rhel_minor_version <= '2' ? "PowerTools" : "powertools"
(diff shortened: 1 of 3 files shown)


recipes/base_install.rb
```diff
@@ -32,10 +32,10 @@
     end
   end
   if node['platform'] == 'centos' && node['platform_version'].to_i == 8
-    # Enable PowerTools Repo so *-devel packages can be installed with DNF
+    # Enable powertools repo so *-devel packages can be installed with DNF
     # Enable EPEL repos
     execute 'dnf enable powertools and EPEL repos' do
-      command "dnf config-manager --set-enabled PowerTools && dnf install -y epel-release"
+      command "dnf config-manager --set-enabled powertools && dnf install -y epel-release"
```
lukeseawalker, line 38: This will not work with older minor version.
Also `dnf install -y epel-release` is duplicated, because it's already present in https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/amis/packer_centos8.json#L162

rexcsn: Should we keep it here for the case of creating cluster from bare AMI?

enrico-usai: Fixed

lukeseawalker (approved):

PR #832, opened 2020-12-09
Fix NFS Threads kitchen test by restarting NFS before checking threads

rexcsn:
* Notifications to restart NFS are queued at the end of Chef run by default
* This causes the kitchen test to fail because NFS server does not restart and load updated NFS config before tests recipe runs
* Add logic in kitchen test to restart NFS so that new config is loaded before checking NFS threads. Actual cluster creation case always worked because services are restarted at the end of chef run
* Add notification to restart NFS when overwriting template. Originally restart was triggered anyway from nfs::_common, but adding this notification as a reassurance that NFS needs to be restarted after config is modified

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -480,13 +480,16 @@ elsif node['cfncluster']['cfn_node_type'] == 'MasterServer'
   end
 end
 
-ruby_block 'check nfs threads' do
-  block do
-    nfs_threads = shell_out!("cat /proc/net/rpc/nfsd | grep th | awk '{print$2}'").stdout.strip.to_i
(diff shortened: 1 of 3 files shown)


recipes/tests.rb
```diff
@@ -489,6 +489,7 @@ module load intelmpi && mpirun --help | grep 'Version 2019 Update 7'
       raise "Expected number of nfs threads configured to be #{expected_threads} but is actually #{nfs_threads}"
     end
   end
+  notifies :restart, "service[#{node['nfs']['service']['server']}]", :before
```
demartinofra, line 492: can't this hide potential bugs in case nfs server is not restarted by the config recipe?

yuleiwan (approved):

PR #834, opened 2020-12-10
Change source package for gridengine on Ubuntu 1804

ddeidda:
Eoan (19.10) repo is no longer available. With this commit the package will be downloaded from Focal Fossa (20.04).

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -302,7 +302,7 @@ when 'debian'
   if node['platform_version'] == '18.04'
     default['cfncluster']['base_packages'].delete('libatlas-dev')
     default['cfncluster']['base_packages'].push('libatlas-base-dev', 'libssl-dev', 'libglvnd-dev')
-    default['cfncluster']['sge']['version'] = '8.1.9+dfsg-9build1'
+    default['cfncluster']['sge']['version'] = '8.1.9+dfsg-9build2'
   end
 
   # Modulefile Directory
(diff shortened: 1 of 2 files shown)


tilne:
Was the skip-changelog-update label added to temporarily prevent checks from failing while this PR is in draft? This change seems worthy of a changelog update to me.

attributes/default.rb
```diff
@@ -302,7 +302,7 @@
   if node['platform_version'] == '18.04'
     default['cfncluster']['base_packages'].delete('libatlas-dev')
     default['cfncluster']['base_packages'].push('libatlas-base-dev', 'libssl-dev', 'libglvnd-dev')
-    default['cfncluster']['sge']['version'] = '8.1.9+dfsg-9build1'
+    default['cfncluster']['sge']['version'] = '8.1.9+dfsg-9build2'
```
lukeseawalker, line 305: can we avoid pinning the build number?

lukeseawalker (approved):

attributes/default.rb
ddeidda: As discussed, we'll evaluate this approach in a further patch.

ddeidda:
> Was the skip-changelog-update label added to temporarily prevent checks from failing while this PR is in draft? This change seems worthy of a changelog update to me.

Why do you think we should mention it? The version of SGE is still 8.1.9; the build number has never been mentioned and it's specific for Ubuntu.

PR #843, opened 2020-12-15
Set MODULEPATH config path based on CentOS version

tilne:
As part of the change from CentOS 8.2 to CentOS 8.3, the environment
modules package installed by default changed from
environment-modules-4.1.4-4 to environment-modules-4.5.2-1. According to
a changelog from rpmfind.net (linked at the end of this commit message),
one of the changes is to 'Make .modulespath a config file'. With this
change, it seems the value of the MODULEPATH environment variable (used
to tell the `module` command which directories to search) must be
configured via /etc/environment-modules/modulespath rather than
/usr/share/Modules/init/.modulespath. This commit sets this path
dynamically based on the OS in use.

Link to info for RPM providing environment-modules-4.5.2-1:
http://rpmfind.net/linux/RPM/centos/8-stream/baseos/x86_64/Packages/environment-modules-4.5.2-1.el8.x86_64.html

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -254,6 +254,14 @@ when 'rhel', 'amazon'
   default['cfncluster']['modulefile_dir'] = "/usr/share/Modules/modulefiles"
   # MODULESHOME
   default['cfncluster']['moduleshome'] = "/usr/share/Modules"
+  # Config file used to set default MODULEPATH list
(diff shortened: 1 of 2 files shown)


rexcsn (approved):

tilne:
Overriding travis checks since they're stuck.

PR #862, opened 2021-02-02
Add logging to debug sporadic chrony check failure

rexcsn:

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -208,11 +208,26 @@ end
 # Amazon Time Sync
 ###################
 if node['init_package'] == 'init'
-  chrony_check_command = "service #{node['cfncluster']['chrony']['service']} status | grep -i running"
+  get_chrony_status_command = "service #{node['cfncluster']['chrony']['service']} status"
 elsif node['init_package'] == 'systemd'
   # $ systemctl show -p SubState <service>
   # SubState=Running
-  chrony_check_command = "systemctl show -p SubState #{node['cfncluster']['chrony']['service']} | grep -i running"
+  get_chrony_status_command = "systemctl show -p SubState #{node['cfncluster']['chrony']['service']}"
+end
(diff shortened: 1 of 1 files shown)


recipes/tests.rb
```diff
@@ -208,11 +208,19 @@
...
   # $ systemctl show -p SubState <service>
   # SubState=Running
-  chrony_check_command = "systemctl show -p SubState #{node['cfncluster']['chrony']['service']} | grep -i running"
+  get_chrony_status_command = "systemctl show -p SubState #{node['cfncluster']['chrony']['service']}"
+end
+chrony_check_command = get_chrony_status_command + " | grep -i running"
+
+ruby_block 'log_chrony_status' do
+  block do
+    chrony_status = shell_out!(get_chrony_status_command).stdout
```
lukeseawalker, line 221: this will report the status of the service without giving other info. In case of failure we know from the ` `chrony_check_command` that the status isn't running. Probably would be useful to retrieve the full service log, e.g. for `journalctl -u node['cfncluster']['chrony']['service']`

recipes/tests.rb
```diff
@@ -208,11 +208,22 @@
...
   # $ systemctl show -p SubState <service>
   # SubState=Running
-  chrony_check_command = "systemctl show -p SubState #{node['cfncluster']['chrony']['service']} | grep -i running"
+  get_chrony_status_command = "systemctl show -p SubState #{node['cfncluster']['chrony']['service']}"
+end
+chrony_check_command = get_chrony_status_command + " | grep -i running"
+
+ruby_block 'log_chrony_status' do
+  block do
+    get_chrony_service_log_command = "journalctl -u #{node['cfncluster']['chrony']['service']}"
```
tilne, line 221: is journalctl systemd-specific? [the man page](https://www.freedesktop.org/software/systemd/man/journalctl.html) makes me think it is. if so will this break on initd systems?

tilne (approved):

PR #864, opened 2021-02-05
Revise ssh_target_checker.sh syntax

rexcsn:
* Current ssh_target_checker.sh syntax works with ssh but might not be compatible with package that depends on ssh, such as pdsh. Change to simpler syntax when parsing VPC CIDR into array

### DIFF
--- a/files/default/ssh_target_checker.sh
+++ b/files/default/ssh_target_checker.sh
@@ -24,7 +24,9 @@ retrieve_vpc_cidr_list() {
     fi
 
     vpc_cidr_uri="http://169.254.169.254/latest/meta-data/network/interfaces/macs/${mac}/vpc-ipv4-cidr-blocks"
-    if ! mapfile -t vpc_cidr_list < <(curl --retry 3 --retry-delay 0 --silent --fail "${vpc_cidr_uri}"); then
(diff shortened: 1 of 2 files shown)


files/default/ssh_target_checker.sh
```diff
@@ -24,7 +24,9 @@ retrieve_vpc_cidr_list() {
     fi
 
     vpc_cidr_uri="http://169.254.169.254/latest/meta-data/network/interfaces/macs/${mac}/vpc-ipv4-cidr-blocks"
-    if ! mapfile -t vpc_cidr_list < <(curl --retry 3 --retry-delay 0 --silent --fail "${vpc_cidr_uri}"); then
+    vpc_cidr_list=(curl --retry 3 --retry-delay 0 --silent --fail "${vpc_cidr_uri}")
```
lukeseawalker, line 27: fix with `vpc_cidr_list=($(curl --retry 3 --retry-delay 0 --silent --fail "${vpc_cidr_uri}"))`

lukeseawalker (commented):
it may be worth a mention in the changelog

lukeseawalker (approved):

PR #870, opened 2021-02-19
Refactoring Munge setup

rexcsn:
* Refactor process to setup munge key

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -432,3 +432,46 @@ def overwrite_nfs_template?
     node['platform'] == 'centos' && node['platform_version'].to_i == 8
   ].any?
 end
+
+def enable_munge_service
+  service "munge" do
+    supports restart: true
+    action %i[enable start]
+  end
+end
+
+def setup_munge_head_node
+  # Generate munge key
+  bash 'generate_munge_key' do
+    user 'munge'
+    group 'munge'
+    cwd '/tmp'
+    code <<-HEAD_MUNGE_KEY
+      set -e
+      # Generates munge key in /etc/munge/munge.key
+      /usr/sbin/mungekey --verbose
+      # Enforce correct permission on the key
+      chmod 0600 /etc/munge/munge.key
+      # Copy key to shared dir
+      cp -p /etc/munge/munge.key /home/munge/.munge.key
+    HEAD_MUNGE_KEY
+  end
+
+  enable_munge_service()
+end
+
+def setup_munge_compute_node
+  # Get munge key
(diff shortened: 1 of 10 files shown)


libraries/helpers.rb
```diff
@@ -432,3 +432,51 @@ def overwrite_nfs_template?
...
+def setup_munge_head_node
+  # Generate munge key
+  bash 'generate_munge_key' do
+    user 'root'
+    group 'root'
+    cwd '/tmp'
+    code <<-MUNGE_KEY
+      set -e
+      # Generates munge key in /etc/munge/munge.key
+      sudo -u munge /usr/sbin/mungekey --verbose
```
demartinofra, line 445: why `verbose`? let's make sure this does not cause the key to be printed to stdout and end up in logs.

libraries/helpers.rb
```diff
@@ -432,3 +432,51 @@ def overwrite_nfs_template?
...
+      # Enforce correct permission on the key
+      chmod 0600 /etc/munge/munge.key
+      chown munge:munge /etc/munge/munge.key
+      # Copy key to shared dir, this tmp key is owned by root
+      cp /etc/munge/munge.key /home/#{node['cfncluster']['cfn_cluster_user']}/.munge.key
```
demartinofra, line 450: shall we copy with `-p` so that we preserve the munge ownership and don't need to enforce in compute nodes again?or actually we could run this bash script directly as munge user

demartinofra: also why not using the home of the munge user to share the key?

libraries/helpers.rb
rexcsn: The key won't be printed, it just prints out some info:
```
# With verbose
$ sudo -u munge /usr/sbin/mungekey --verbose
mungekey: Info: Created "/etc/munge/munge.key" with 1024-bit key
$

# Without verbose
$ sudo -u munge /usr/sbin/mungekey
$
```

libraries/helpers.rb
rexcsn: > shall we copy with -p so that we preserve the munge ownership and don't need to enforce in compute nodes again?or actually we could run this bash script directly as munge user

Definitely yes for `-p`, I will try and see if I can run the script as munge user, if not I will keep running as root

> also why not using the home of the munge user to share the key?

I just did it because we don't have a home directory for munge user, but we can create one. Currently we have:
```
$ ls /home
dcvextauth  sgeadmin  slurm  ubuntu
```

demartinofra (approved):
👍🏻

PR #871, opened 2021-02-22
Upgrade to Slurm 20.11.4

rexcsn:
* Upgrade and install Slurm 20.11.4
* Update slurmctld and slurmd systemd unit files according to latest provided by slurm
* Add new SlurmctldParameters, power_save_min_interval=30, which means power actions will only be processed every 30 seconds, and slurm will likely to group more nodes in a single call to power_up or power_down nodes
* Add new SlurmctldParameters, cloud_reg_addrs, which will reset a node's NodeAddr automatically on power_down


files/default/slurmctld.service
```diff
@@ -4,14 +4,11 @@ After=network.target munge.service
...
-Type=forking
+Type=simple
 EnvironmentFile=-/etc/sysconfig/slurmctld
-ExecStart=/opt/slurm/sbin/slurmctld $SLURMCTLD_OPTIONS
+ExecStart=/opt/slurm/sbin/slurmctld -D $SLURMCTLD_OPTIONS
 ExecReload=/bin/kill -HUP $MAINPID
-PIDFile=/var/run/slurmctld.pid
-LimitNOFILE=562930
-LimitMEMLOCK=infinity
-LimitSTACK=infinity
```
demartinofra, line 14: please restore these cause we explicitly changed them for large size clusters

templates/default/slurm/slurm.conf.erb
```diff
@@ -24,10 +24,10 @@ ReconfigFlags=KeepPartState
 #
 # CLOUD CONFIGS OPTIONS
 <% if node['cfncluster']['use_private_hostname'] == 'true' or node['cfncluster']['cfn_dns_domain'].nil? or node['cfncluster']['cfn_dns_domain'].empty? -%>
-SlurmctldParameters=idle_on_node_suspend
+SlurmctldParameters=idle_on_node_suspend,power_save_min_interval=30
```
demartinofra, line 27: isn't 30 seconds a little bit too much? what was the default before? I would set it to something like 10 seconds maybe for now? This way we are still able to batch sequence of jobs submitted programmatically and it might not be so impactful for the scaling time e for users that submit jobs individually to leverage the current all-or-nothing approach.

demartinofra: also in the changelog you mention `cloud_reg_addrs` but I don't see it anywhere

files/default/slurm/templates/slurm_parallelcluster_queue_gres.conf
```diff
@@ -5,10 +5,10 @@
         {% set dynamic_size = compute_resource_config.max_count - compute_resource_config.min_count %}
         {% set sanitized_instance_type = compute_resource_config.instance_type | sanify_instance_type %}
         {% if static_size > 0 %}
-NodeName={{ queue_name }}-st-{{ sanitized_instance_type }}-[1-{{ static_size }}] Name=gpu Type=tesla File=/dev/nvidia[0-{{ compute_resource_config.gpus - 1 }}]
+NodeName={{ queue_name }}-st-{{ sanitized_instance_type }}-[1-{{ static_size }}] Name=gpu Type={{ compute_resource_config.gpu_type }} File=/dev/nvidia[0-{{ compute_resource_config.gpus - 1 }}]
```
demartinofra, line 8: moving from `tesla` to something like A100 can be a breaking change for users submitting with tesla as job requirement. We might be ok with that but let's review this choice once more.

demartinofra: Also if we are using the GPU name from the describe-instance-types API be aware that some of them might have spaces in it (e.g. g4ad.4xlarge)
```
                    {
                        "Name": "Radeon Pro V520",
                        "Manufacturer": "AMD",
                        "Count": 1,
                        "MemoryInfo": {
                            "SizeInMiB": 8192
                        }
                    }
```

Shall we do a lowercase conversion + removal of whitespaces?

templates/default/slurm/slurm.conf.erb
demartinofra: Thanks for the explanation. I still think a worst case scenario of 30 secs to wait after job submission is a bit too much. I would cut it in half unless we are already aware of bottlenecks that might justify the 30 secs value.

demartinofra (approved):

PR #881, opened 2021-02-26
Share munge key from dir owned by root, under the default user home f…

lukeseawalker:
…older

Share munge key from dir owned by root from head node, than setup it in compute node
This because `munge` user can be different between head and compute node (e.g. during baking at runtime), hence the shared home munge dir can have different permission.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -446,30 +446,46 @@ def setup_munge_head_node
     user 'munge'
     group 'munge'
     cwd '/tmp'
-    code <<-HEAD_MUNGE_KEY
+    code <<-HEAD_CREATE_MUNGE_KEY
       set -e
       # Generates munge key in /etc/munge/munge.key
       /usr/sbin/mungekey --verbose
       # Enforce correct permission on the key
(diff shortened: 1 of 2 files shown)


libraries/helpers.rb
```diff
@@ -446,15 +446,30 @@ def setup_munge_head_node
...
+      set -e
+      mkdir /home/#{node['cfncluster']['cfn_cluster_user']}/.munge
       # Copy key to shared dir
-      cp -p /etc/munge/munge.key /home/munge/.munge.key
-    HEAD_MUNGE_KEY
+      cp /etc/munge/munge.key /home/#{node['cfncluster']['cfn_cluster_user']}/.munge/.munge.key
+    HEAD_SHARE_MUNGE_KEY
   end
 
   enable_munge_service()
```
demartinofra, line 475: probably we need to remove this

lukeseawalker: thanks, done

demartinofra (approved):

PR #885, opened 2021-03-09
Install AWSBatch CLI in virtual env

rexcsn:
* Create separate python3 virtual env for AWSBatch CLI and install Batch CLI in this dedicated virtual env
* If using AWSBatch, add batch virtual env to profile.d so batch CLI commands are in default PATH

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   - Add new SlurmctldParameters, power_save_min_interval=30, so power actions will be processed every 30 seconds
   - Add new SlurmctldParameters, cloud_reg_addrs, which will reset a node's NodeAddr automatically on power_down
   - Specify instance GPU model as GRES GPU Type in gres.conf
-  
+- Install ParallelCluster AWSBatch CLI in dedicated python3 virtual env.
+
 **BUG FIXES**
 - Fix `encrypted_ephemeral = true` when using Alinux2 or CentOS8
 
--- a/recipes/awsbatch_config.rb
+++ b/recipes/awsbatch_config.rb
@@ -19,6 +19,14 @@
 include_recipe 'aws-parallelcluster::base_config'
 include_recipe 'aws-parallelcluster::base_install'
 
+# Add awsbatch virtualenv to default path
+template "/etc/profile.d/pcluster_awsbatchcli.sh" do
+  source "pcluster_awsbatchcli.sh.erb"
+  owner 'root'
+  group 'root'
+  mode '0644'
+end
+
 # Install aws-parallelcluster-awsbatch-cli.cfg
 awsbatch_cli_config_dir = "/home/#{node['cfncluster']['cfn_cluster_user']}/.parallelcluster/"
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   - Add new SlurmctldParameters, power_save_min_interval=30, so power actions will be processed every 30 seconds
   - Add new SlurmctldParameters, cloud_reg_addrs, which will reset a node's NodeAddr automatically on power_down
   - Specify instance GPU model as GRES GPU Type in gres.conf
+- Install AWSBatch CLI in dedicated python3 virtual env
```
tilne, line 22: this appears to be under-indented

tilne (approved):
LGTM.

Is it worth adding a simple kitchen test case that makes sure all of the Batch CLI commands are on the default head node PATH by calling each without an absolute path and passing the `-h` flag?

recipes/awsbatch_config.rb
```diff
@@ -19,6 +19,14 @@
 include_recipe 'aws-parallelcluster::base_config'
 include_recipe 'aws-parallelcluster::base_install'
 
+# Add awsbatch virtualenv to default path
+template "/etc/profile.d/awsbatchcli.sh" do
```
demartinofra, line 23: I'd call it pcluster.sh or pcluster_awsbatchcli.sh

templates/default/awsbatchcli.sh.erb
```diff
@@ -0,0 +1,8 @@
+#
+# awsbatchcli.sh:
+#   Setup AWSBatch CLI environment variables
```
demartinofra, line 3: Let's always refer to this as the ParallelCluster AWSBatch CLI just because this is not an offcial AWS Batch CLI. Same applies everywhere else

recipes/tests.rb
```diff
@@ -593,3 +593,22 @@ module load armpl/#{node['cfncluster']['armpl']['version']}
     user node['cfncluster']['cfn_cluster_user']
   end
 end
+
+###################
+# Pcluster AWSBatch CLI
+###################
+if node['cfncluster']['cfn_scheduler'] == 'awsbatch'
```
tilne, line 600: minor: ideally there would be another condition here that prevents this block from running during the compute fleet kitchen tests. in practice i don't think it will matter since we don't run the kitchen tests with batch for compute fleets.

CHANGELOG.md
tilne: ah i see now this was my mistake

CHANGELOG.md
```diff
@@ -19,7 +19,8 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   - Add new SlurmctldParameters, power_save_min_interval=30, so power actions will be processed every 30 seconds
   - Add new SlurmctldParameters, cloud_reg_addrs, which will reset a node's NodeAddr automatically on power_down
   - Specify instance GPU model as GRES GPU Type in gres.conf
-  
+- Install AWSBatch CLI in dedicated python3 virtual env.
```
tilne, line 22: probably this needs to change per Francesco's comment about differentiating between our Batch CLI versus an official one

demartinofra (approved):

PR #887, opened 2021-03-11
Support development instance_types_data configuration

ddeidda:
This commit integrates the support for the new `instance_types_data` configuration parameter, copying the value coming from configuration, if present, into ParallelCluster daemons configuration files.

### DIFF
--- a/templates/default/jobwatcher.cfg.erb
+++ b/templates/default/jobwatcher.cfg.erb
@@ -4,3 +4,4 @@ scheduler = <%= node['cfncluster']['cfn_scheduler'] %>
 stack_name = <%= node['cfncluster']['stack_name'] %>
 cfncluster_dir = <%= node['cfncluster']['base_dir'] %>
 proxy = <%= node['cfncluster']['cfn_proxy'] %>
+instance_types_data = <%= node['instance_types_data'].to_json %>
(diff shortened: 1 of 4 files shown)


templates/default/jobwatcher.cfg.erb
```diff
@@ -4,3 +4,4 @@ scheduler = <%= node['cfncluster']['cfn_scheduler'] %>
 stack_name = <%= node['cfncluster']['stack_name'] %>
 cfncluster_dir = <%= node['cfncluster']['base_dir'] %>
 proxy = <%= node['cfncluster']['cfn_proxy'] %>
+instance_types_data = <%= node['instance_types_data'].to_json %>
```
demartinofra, line 7: why do we need the to_json? the node is reading this as a string. what is the to_json for?

ddeidda: Because it is a Hash object in memory, so it would get serialized like here:
```
{:key=>"key", :value=>"value"}
```

demartinofra: Thanks

demartinofra (approved):

demartinofra (approved):

PR #893, opened 2021-03-15
Move instance types data into cfncluster section

ddeidda:


templates/default/jobwatcher.cfg.erb
```diff
@@ -4,4 +4,4 @@ scheduler = <%= node['cfncluster']['cfn_scheduler'] %>
 stack_name = <%= node['cfncluster']['stack_name'] %>
 cfncluster_dir = <%= node['cfncluster']['base_dir'] %>
 proxy = <%= node['cfncluster']['cfn_proxy'] %>
-instance_types_data = <%= node['instance_types_data'].to_json %>
\ No newline at end of file
+instance_types_data = <%= node['cfncluster']['instance_types_data'].to_json %>
```
demartinofra, line 7: don't we need to update the default as well?

ddeidda: Right! Thanks

ddeidda: Done

demartinofra (approved):

PR #898, opened 2021-03-20
Install epel package in createami and runtime baking

yuleiwan:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/recipes/base_install.rb
+++ b/recipes/base_install.rb
@@ -22,7 +22,7 @@ when 'rhel', 'amazon'
   include_recipe 'yum'
   if node['platform_family'] == 'amazon'
     alinux_extras_topic 'epel'
-  elsif node['platform_version'].to_i < 7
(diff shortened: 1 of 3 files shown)


recipes/base_install.rb
```diff
@@ -22,7 +22,7 @@
   include_recipe 'yum'
   if node['platform_family'] == 'amazon' && node['platform_version'].to_i == 2
     alinux_extras_topic 'epel'
-  elsif node['platform_version'].to_i < 7
+  elsif node['platform_family'] == 'rhel'
```
tilne, line 25: minor: is there a reason to use `node['platform_family'] == 'rhel'` instead of `node['platform'] == 'centos'`? in practice i think this is the same since we don't support RHEL and AL1 and AL2 AMIs will have a platform_family of `'amazon'`.

tilne (approved):

PR #903, opened 2021-03-25
Install Python3 by default and update ebsnvme-id script

rexcsn:
* Install Python3 in all OSs by default
* Update and move files/default/parallelcluster-ebsnvme-id to templates/default/parallelcluster-ebsnvme-id.erb, which is now python3 compatible. Run parallelcluster-ebsnvme-id with cookbook virtual env. Script is taken from `/sbin/ebsnvme-id` that is distributed with `alinux2`, see [official guide on identifying EBS device](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html#identify-nvme-ebs-device) for details
* Remove files/centos8/parallelcluster-ebsnvme-id shell script and use default parallelcluster-ebsnvme-id python script instead


yuleiwan (commented):
Should we add a kitchen test to verify python2 and python2-pip has been removed?

templates/default/parallelcluster-ebsnvme-id.erb
```diff
@@ -1,28 +1,24 @@
-#!/usr/bin/env python
+#!<%= node['cfncluster']['cookbook_virtualenv_path'] %>/bin/python
```
yuleiwan, line 1: why convert python to vitual environment? Doesn't python at system level use same version as python at virtualenv ?

yuleiwan (approved):

PR #904, opened 2021-03-25
Add linters to CI 

tilne:
* Enable dependency updates via dependabot PRs
* Add rubocop, foodcritic, semgrep, bandit, and dependency-check to CI
* Add a [pre-commit](https://pre-commit.com/) config file that can be used to automatically run linters pre-commit on modified source


.isort.cfg
```diff
@@ -1,6 +1,7 @@
 [settings]
 line_length=120
-known_third_party=boto3,botocore,awscli,tabulate,argparse,configparser,pytest,pytest,pytest-datadir,pytest-html,pytest-rerunfailures,pytest-xdist,argparse,retrying,junitparser,Jinja2
+known_future_library=future
```
rexcsn, line 3: Probably don't need this anymore because we are using python3 everywhere?

Gemfile
```diff
@@ -8,6 +8,7 @@ group :style do
   gem 'foodcritic', '~> 16.2.0'
   gem 'rake', '~> 13.0.1'
   gem 'rubocop', '~> 0.80.1'
+  gem 'rubocop-gitlab-security'
```
rexcsn, line 11: Do we need to pin to a version? I assume not because we would want to get the most updated rubocop security linter?

tilne: It's not a bad idea to pin to the latest version. If the commit from this PR for enabling dependabot PRs works as expected it will remind us to update this whenever new versions are available anyways.

.isort.cfg
tilne: Changed

rexcsn (approved):

PR #911, opened 2021-04-01
Enable DCV+Alinux2+Batch kitchen test

rexcsn:
* Move "Ensure local users can access X server" so it's only run if DCV is enabled
* Enable DCV in Batch head node kitchen test


recipes/awsbatch_config.rb
```diff
@@ -64,3 +64,8 @@
     command "#{node['cfncluster']['awsbatch_virtualenv_path']}/bin/pip install aws-parallelcluster==#{node['cfncluster']['cfncluster-version']}"
   end
 end
+
+if node['cfncluster']['dcv_enabled'] == "master"
+  # Activate DCV on head node
+  include_recipe 'aws-parallelcluster::dcv_config'
```
lukeseawalker, line 70: isn't the `head_node_base_config` recipe taking care of calling `dcv_config`?

rexcsn: Ah yes, so it was actually an issue of DCV not enabled in kitchen test. PR is updated

yuleiwan (approved):

PR #913, opened 2021-04-02
Add support for Ubuntu 20.04

ddeidda:
Packer script, recipes and configuration files needed to introduce the support for Ubuntu 20.04 on x86_64 and arm_64 architectures.


files/ubuntu-20.04/configure_nw_interface.sh
```diff
@@ -0,0 +1,83 @@
...
+         via: 0.0.0.0
+         scope: link
+         table: ${ROUTE_TABLE}
+      routing-policy:
+        - from: ${DEVICE_IP_ADDRESS}
+          table: ${ROUTE_TABLE}
+EOF
+
+# Hook to delete automatic routes to subnet automatically created from kernel.
+# This is needed in Ubuntu 20 as well as for Ubuntu 18 for a slightly different reason. Automatic routes here are
```
rexcsn, line 67: nit: I think this file is almost identical to the one used for Ubuntu18. I see that there are comments here that explains the differences and why things are done in ubuntu20. Maybe we can just have 1 file with comments for both Ubuntu18 and 20?

files/ubuntu-20.04/ec2blkdev-init
```diff
@@ -0,0 +1,46 @@
...
+# Required-Stop: $network
+# Default-Start: 2 3 4 5
+# Default-Stop: 0 1 6
+# Short-Description: ec2blkdev
+# Description: Script to call udevadm to setup any links to devices already attached at instance launch
+### END INIT INFO
+
+. /lib/lsb/init-functions
+
+prog=/bin/udevadm
```
rexcsn, line 20: I think this is the only line that is different between Ubuntu18 and Ubuntu20 version of the file? Maybe we can make this a template and substitute in the different programs, but only have 1 template for both ubuntu OSs?

files/ubuntu-20.04/sge_preinstall.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
rexcsn, line 1: This file looks identical to the Ubuntu18 version. Can we use the same file under `files/ubuntu/` for both OSs? Same comment applies for other files that might be the same for Ubuntu OSs

files/ubuntu-20.04/configure_nw_interface.sh
ddeidda: Agreed. I moved the files into a common `ubuntu` folder. 
Comments for `configure_nw_interfaces.sh` script have been merged, `ec2blkdev-init` logic unified.

files/ubuntu-20.04/ec2blkdev-init
ddeidda: Done

files/ubuntu-20.04/sge_preinstall.sh
ddeidda: Done

rexcsn (dismissed):

attributes/default.rb
```diff
@@ -84,7 +84,7 @@
 default['cfncluster']['armpl']['platform'] = value_for_platform(
   'centos' => { '~>8' => 'RHEL-8' },
   'amazon' => { '2' => 'RHEL-8' },
-  'ubuntu' => { '18.04' => 'Ubuntu-16.04' }
+  'ubuntu' => { '>=18.04' => 'Ubuntu-16.04' }
```
lukeseawalker, line 87: not related to this PR but why we are using `Ubuntu-16.04` archive for both 18.04 and 20.04?

recipes/setup_raid_on_head_node.rb
```diff
@@ -59,6 +59,7 @@
   # Create RAID device with mdadm
   raid_superblock_version = value_for_platform(
     'centos' => { '>=8' => '1.2' },
+    'ubuntu' => { '>=20.04' => '1.2' },
```
lukeseawalker, line 62: what is happening for U18 case? is it going to use the default value?

ddeidda: Yes, 0.90 is used on Ubuntu 18

attributes/default.rb
ddeidda: It was the only available Ubuntu platform. Now I see that there is a new version, released on Mar 30, which should target Ubuntu 18.04. We might want to do the upgrade as a separate patch.

PR #915, opened 2021-04-05
Transition from IMDSv1 to IMDSv2

chenwany:
Signed-off-by: chenwany <chenwany@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/files/default/attachVolume.py
+++ b/files/default/attachVolume.py
@@ -57,10 +57,17 @@ def main():
         sys.exit(1)
 
     # Get instance ID
-    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id").text
+    token = requests.put(
+        "http://169.254.169.254/latest/api/token",
+        headers={"X-aws-ec2-metadata-token-ttl-seconds": "300"}
+    )
+    headers = {}
+    if token.status_code == requests.codes.ok:
+        headers["X-aws-ec2-metadata-token"] = token.content
+    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id", headers=headers).text
 
     # Get region
-    region = requests.get("http://169.254.169.254/latest/meta-data/placement/availability-zone").text
+    region = requests.get("http://169.254.169.254/latest/meta-data/placement/availability-zone", headers=headers).text
     region = region[:-1]
 
     # Generate a list of system paths minus the root path
--- a/files/default/ec2_dev_2_volid.py
+++ b/files/default/ec2_dev_2_volid.py
@@ -39,10 +39,18 @@ def main():
         dev = "/dev/" + dev
 
     # Get instance ID
-    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id").text
+    token = requests.put(
+        "http://169.254.169.254/latest/api/token",
+        headers={"X-aws-ec2-metadata-token-ttl-seconds": "300"}
+    )
+
+    headers = {}
+    if token.status_code == requests.codes.ok:
+        headers["X-aws-ec2-metadata-token"] = token.content
+    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id", headers=headers).text
 
     # Get region
-    region = requests.get("http://169.254.169.254/latest/meta-data/placement/availability-zone").text
+    region = requests.get("http://169.254.169.254/latest/meta-data/placement/availability-zone", headers=headers).text
     region = region[:-1]
 
     # Parse configuration file to read proxy settings
--- a/files/default/slurm/pcluster_slurm_config_generator.py
+++ b/files/default/slurm/pcluster_slurm_config_generator.py
@@ -13,13 +13,19 @@ import argparse
 import json
 import logging
 import re
-import subprocess
 from os import makedirs, path
 from socket import gethostname
 
+import requests
 from jinja2 import Environment, FileSystemLoader
 
 log = logging.getLogger()
+
+
+class CriticalError(Exception):
+    """Critical error for the daemon."""
+
+    pass
 
 
 def generate_slurm_config_files(output_directory, template_directory, input_file, dryrun):
@@ -89,19 +95,7 @@ def _get_head_node_config():
 
 def _get_head_node_private_ip():
     """Get head node private ip from EC2 metadata."""
-    try:
-        private_ip = subprocess.run(  # nosec nosemgrep
-            "curl --retry 3 http://169.254.169.254/latest/meta-data/local-ipv4",
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            check=True,
-            encoding="utf-8",
-            shell=True,
-        ).stdout
-        return private_ip
-    except Exception:
-        log.error("Encountered exception when retrieving head node private IP.")
-        raise
+    return _get_metadata("local-ipv4")
 
 
 def _generate_queue_config(queue_name, queue_config, is_default_queue, file_type, jinja_env, output_dir, dryrun):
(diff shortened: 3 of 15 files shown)


util/cinc-install-1.1.0.sh
```diff
@@ -259,6 +259,26 @@ do_download() {
   unable_to_retrieve_package
 }
 
+# get_region INSTANCE_METADATA_FILE
+# get region from metadata
+get_region() {
```
rexcsn, line 264: We don't need to rename this file, we can just add a comment near to top and keep track of the version. See internal CR for how we will upload the file

util/cinc-install-1.1.0.sh
```diff
@@ -355,11 +375,12 @@ tmp_dir="$tmp/install.sh.$$"
 channel="stable"
 project="cinc"
 
-while getopts pnv:c:f:P:d:s:l:a opt
+while getopts pnv:b:c:f:P:d:s:l:a opt
 do
   case "$opt" in
 
     v)  version="$OPTARG";;
+    b)  bucket="$OPTARG";;
```
rexcsn, line 383: Can we add a short inline comment to explain this option a bit more? Something like, "Download CINC client from specified s3 bucket"

util/cinc-install-1.1.0.sh
```diff
@@ -259,6 +259,26 @@ do_download() {
...
+    echo "Trying curl to get region with IMDSv2 ..."
+    token=$(curl -s --connect-timeout 5 --max-time 5 --retry 3 -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 300")
+    region=$(curl -s -H "X-aws-ec2-metadata-token: ${token}" http://169.254.169.254/latest/dynamic/instance-identity/document|grep region|awk -F\" '{print $4}')
+  elif exists python; then
+    echo "Trying python to get region with IMDSv2 ..."
+    token=$(python -c "import sys,requests; sys.stdout.write(requests.put('http://169.254.169.254/latest/api/token', headers={'X-aws-ec2-metadata-token-ttl-seconds': '300', 'User-Agent': 'mixlib-install/3.11.27'}).content)")
+    region=$(python -c "import sys,requests,json; sys.stdout.write(json.loads(requests.get('http://169.254.169.254/latest/dynamic/instance-identity/document', headers={ 'X-aws-ec2-metadata-token': '${token}', 'User-Agent': 'mixlib-install/3.11.27' }).content.decode())['region'])")
+  fi
+
+  if test "x$region" = "x"; then
```
rexcsn, line 275: nit: what about `if [ -z "$region" ]`, to be consistent with other bash conditionals we do in the repo. Same comment for other places where we are doing this `test "x$var" = "x"` logic

.kitchen.cloud.yml
```diff
@@ -17,7 +17,7 @@ driver_config:
 
 provisioner:
   name: chef_zero
-  require_chef_omnibus: 15.11.8
+  require_chef_omnibus: 16.13.16
```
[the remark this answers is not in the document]
rexcsn: On a second thought chef should not be visible to the user, so it's ok to leave it out.

util/cinc-install-1.1.0.sh
chenwany: In `cinc-install.sh`, all bash  condition are in `test "x$var" = "x"`, so I make it the same format here to be consistence

util/cinc-install-1.1.0.sh
chenwany: Yes! Done

files/default/slurm/pcluster_slurm_config_generator.py
```diff
@@ -90,8 +90,17 @@ def _get_head_node_config():
 def _get_head_node_private_ip():
     """Get head node private ip from EC2 metadata."""
     try:
+        token = subprocess.run(  # nosec nosemgrep
```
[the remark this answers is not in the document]
rexcsn: Yes, we probably don't want to share a helper function, but we can at least put this `subprocess.run` call that uses all the options into a helper function within this script, so we don't have to write everything twice, let's at least do this.

Nice to have: We can also use the `requests` based way of retrieving metadata as well, since that is more pythonic than using a `subprocess` call

recipes/network_interfaces_config.rb
```diff
@@ -15,9 +15,11 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-def network_interface_macs
+def network_interface_macs(token)
   uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs")
-  res = Net::HTTP.get_response(uri)
+  request = Net::HTTP::Get.new(uri)
+  request["X-aws-ec2-metadata-token"] = token
+  res = Net::HTTP.new("169.254.169.254").request(request)
```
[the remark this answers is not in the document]
rexcsn: I see that we are doing similar things in `get_vpc_ipv4_cidr_blocks` in `helpers.rb`. Ideally we would just have 1 helper function in `helpers.rb` we can use to retrieve metadata from all recipes

files/default/slurm/pcluster_slurm_config_generator.py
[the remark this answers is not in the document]
chenwany: Right! Changed it used requests and add `_get_metadata` function in the file

recipes/network_interfaces_config.rb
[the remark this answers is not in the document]
chenwany: I see. Added

rexcsn (dismissed):
Well done!

rexcsn (approved):

rexcsn (approved):

files/default/attachVolume.py
```diff
@@ -57,10 +57,17 @@ def main():
...
     # Get instance ID
-    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id").text
+    token = requests.put(
+        "http://169.254.169.254/latest/api/token",
+        headers={"X-aws-ec2-metadata-token-ttl-seconds": "300"}
+    )
+    headers = {}
+    if token.status_code == requests.codes.ok:
+        headers["X-aws-ec2-metadata-token"] = token.content
+    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id", headers=headers).text
```
mauri-melato, line 67: At the minimum, this logic representing a precise functionality should be wrapped in a function, e.g. `get_instance_id()`

files/default/ec2_dev_2_volid.py
```diff
@@ -39,10 +39,18 @@ def main():
...
-    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id").text
+    token = requests.put(
+        "http://169.254.169.254/latest/api/token",
+        headers={"X-aws-ec2-metadata-token-ttl-seconds": "300"}
+    )
+
+    headers = {}
+    if token.status_code == requests.codes.ok:
+        headers["X-aws-ec2-metadata-token"] = token.content
+    instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id", headers=headers).text
```
mauri-melato, line 50: Same comment as above. Can we reuse the same function definition in a shared python file?
In any case, we have a bunch of code meant to a single specific functionality that's described in the comment `# Get instance ID`.  In these cases you should have better speaking code, here a function named `get_instance_id()`, that would let you to remove both all those lines of code and the comment as well..

files/default/attachVolume.py
chenwany: Yes, add a function get_imdsv2_token() in this PR: https://github.com/aws/aws-parallelcluster-cookbook/pull/940 to warp the part of getting token to make it more clear.
I didn't warp get_instance_id() because the token is use both in get instance id and get region, there's only one line of code for get instance id: instance_id = requests.get("http://169.254.169.254/latest/meta-data/instance-id", headers=headers).text

files/default/ec2_dev_2_volid.py
chenwany: Also add a function get_imdsv2_token() in this PR: #940 to warp the part of getting token to make it more clear as explain above.
I am not reusing the same function definition in a shared python file, the reason is that this file is under files/default directory in our code. The cookbook files will be specified to put into different paths. For example, this file will be put in path '/sbin/ec2_dev_2_volid.py'. Other files will be specified to put in other places. These python files are called separately. We may not want to import another python files when we called them.

PR #918, opened 2021-04-08
3.0: Drop deprecated features, use inclusive language and align cookbook to 3.0 CLI

enrico-usai:
## Use new configuration file to generate slurm configuration
To unblock end-to-end test, this commit will use the new configuration file to generate slurm configuration. There are still some known issues left behind, namely: enable_efa is not handled, pcluster update is not done.

## Add log_group_name to dna.json
A timestamp is appended on the name of the CloudWatch log group, which makes impossible to infer the log group name by cluster name. So the cookbook needs to use the log group name passed in from CLI.

## Remove every reference to Torque and SGE schedulers

* Remove recipes and scripts
* Remove nodewatcher, jobwatcher, sqswatcher references
* Remove from CloudWatch Logs configuration
* Remove from Kitchen tests

## Rename cookbook parameters
- Rename cfn_efs to efs_fs_id
- Rename cfn_shared_dir to ebs_shared_dirs
- Rename cfn_cluster_cw_logging_enabled to cw_logging_enabled
- Change accepted value for dcv_enabled variable from master to head_node

## Use inclusive language: from Master to HeadNode
- Rename NodeType and tags from MasterServer to HeadNode

## Other
* Change "cfncluster" to "cluster" in dna.json and chef attributes 
* Remove SQS and compute_ready script references
* Remove cfn_ prefix from cookbook attributes
* Use parallelcluster as prefix for Python packages


recipes/cloudwatch_agent_config.rb
```diff
@@ -64,19 +64,19 @@
...
+  command "#{node.default['cluster']['cookbook_virtualenv_path']}/bin/python #{validator_script_path}"
 end
 
 execute "cloudwatch-config-creation" do
   user 'root'
   environment(
-    'LOG_GROUP_NAME' => "/aws/parallelcluster/#{node['cfncluster']['stack_name'].split(/^parallelcluster-/)[1]}",
-    'SCHEDULER' => node['cfncluster']['cfn_scheduler'],
-    'NODE_ROLE' => node['cfncluster']['cfn_node_type'],
+    'LOG_GROUP_NAME' => node['cluster']['log_group_name'], # TODO: check what will happen if log_group_name==null
```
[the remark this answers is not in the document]
enrico-usai: Removed, we have integration tests to check CW enabled/disabled

demartinofra:
>A timestamp is appended on the name of the CloudWatch log group, which makes impossible to infer the log group name by cluster name. So the cookbook needs to use the log group name passed in from CLI.

about this have we checked if the script to retrieve CW logs we have in the main package still works?

attributes/default.rb
```diff
@@ -263,115 +256,95 @@
...
-                                                httpd boost-devel system-lsb mlocate atlas-devel fftw-devel glibc-static iproute
-                                                libffi-devel dkms mysql-devel libedit-devel postgresql-devel postgresql-server
-                                                sendmail cmake byacc libglvnd-devel mdadm libgcrypt-devel libevent-devel
-                                                libxml2-devel perl-devel dpkg-dev tar gzip bison flex gcc gcc-c++ patch
-                                                rpm-build rpm-sign system-rpm-config cscope ctags diffstat doxygen elfutils
-                                                gcc-gfortran git indent intltool patchutils rcs subversion swig systemtap curl
-                                                jq wget python-pip NetworkManager-config-routing-rules libibverbs-utils
-                                                librdmacm-utils python3 python3-pip]
+    default['cluster']['base_packages'] = %w[vim ksh tcsh zsh openssl-devel ncurses-devel pam-devel net-tools openmotif-devel
```
demartinofra, line 304: I'm wondering if there are packages we can remove in pcluster3, for example:
* can we only install python3?
* we should really remove dpkg on non debian systems
* are there packages that were only required by SGE/Torque or Ganglia? (probably apache2?)

templates/default/lsb.hosts.erb
```diff
@@ -27,7 +27,7 @@ HOST_NAME     MXJ JL/U   r1m    pg    ls     tmp  DISPATCH_WINDOW  # Keywords
 #host0        1    1   3.5/4.5  15/   12/15  0      ()		   # Example
 #host1       ()   2     3.5  15/18   12/    0/  (5:19:00-1:8:30 20:00-8:30)
 #host2        ()   ()   3.5/5   18    15     ()     ()		   # Example
-<% if node['cfncluster']['cfn_node_type'] == 'MasterServer' -%>
+<% if node['cluster']['node_type'] == 'HeadNode' -%>
```
demartinofra, line 30: is this file used?

enrico-usai:
> about this have we checked if the script to retrieve CW logs we have in the main package still works?

Yes, we improved it in https://github.com/aws/aws-parallelcluster/commit/d6a004e7f5a37e87f2ea9f486a8dd16ffb2ba2c4
to be able to retrieve logs from both 3.x and 2.x clusters. We're still missing the possibility to distinguish logs from stack with the same name.

templates/default/lsb.hosts.erb
enrico-usai: nope, you right, I think it's here because in past we supported other schedulers (e.g. LSF), I'll remove it.

attributes/default.rb
enrico-usai: If you agree I'll do it in a next patch.

demartinofra (approved):

PR #923, opened 2021-04-14
Restore epel repo installation that is required for jq package

enrico-usai:
epel must be installed in the packer definition file because jq is essential to the early stages of the build process.


attributes/default.rb
```diff
@@ -294,7 +294,7 @@
                                                 blas-devel fftw-devel libffi-devel openssl-devel dkms mariadb-devel libedit-devel
                                                 libical-devel postgresql-devel postgresql-server sendmail libxml2-devel libglvnd-devel
                                                 mdadm python python-pip libssh2-devel libgcrypt-devel libevent-devel glibc-static bind-utils
-                                                iproute NetworkManager-config-routing-rules python3 python3-pip]
+                                                iproute NetworkManager-config-routing-rules python3 python3-pip jq wget curl]
```
lukeseawalker, line 297: this packages are already installed in the packer definition file, because they are essential to the early stages of the build process as well

lukeseawalker (approved):

PR #928, opened 2021-04-16
Remove dpkg dependency and revise sge_install for Alinux2

rexcsn:
* Do not use Debian package and patch for Alinux2, this allows us to remove dependency on dpkg
* Create patch file for changes required to install SGE on Alinux2, and apply patch in sge_preinstall.sh
* Add kitchen test


files/amazon-2/sge-tcsh.patch
```diff
@@ -0,0 +1,20 @@
+patch e91cde95c7e32f6ff6d8bdad463189c1da296cf4
```
lukeseawalker, line 1: if this diff it's the same used for the centos-8 case, cannot we put this under the `default` folder and use it for both alinux2 and centos8?

files/amazon-2/sge_preinstall.sh
```diff
@@ -1,5 +1,6 @@
-#!/bin/bash
+#!/bin/sh
```
lukeseawalker, line 1: why `sh`?

rexcsn: This file is mostly the same as the `sge_preinstall.sh` in `files/centos-8`, and we were using `sh` in that script already. Is there a concern against using `sh` as opposed to `bash`?

lukeseawalker (approved):

PR #935, opened 2021-04-26
Enabled FSx on Centos 7.9 ARM

sean-smith:

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -395,11 +395,13 @@ default['cfncluster']['lustre']['public_key'] = value_for_platform(
 )
 # Lustre repo string is built following the official doc
 # https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
+# 'centos' is used for arm and 'el' for x86_64
+default['cfncluster']['lustre']['centos7']['base_url_prefix'] = arm_instance? ? 'centos' : 'el'
 default['cfncluster']['lustre']['base_url'] = value_for_platform(
   'centos' => {
     # node['kernel']['machine'] contains the architecture: 'x86_64' or 'aarch64'
     '>=8' => "https://fsx-lustre-client-repo.s3.amazonaws.com/el/8.#{find_rhel_minor_version}/#{node['kernel']['machine']}/",
-    'default' => "https://fsx-lustre-client-repo.s3.amazonaws.com/el/7.#{find_rhel_minor_version}/x86_64/"
+    'default' => "https://fsx-lustre-client-repo.s3.amazonaws.com/#{default['cfncluster']['lustre']['centos7']['base_url_prefix']}/7.#{find_rhel_minor_version}/#{node['kernel']['machine']}/"
   },
   'ubuntu' => { 'default' => "https://fsx-lustre-client-repo.s3.amazonaws.com/ubuntu" }
 )
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -302,7 +302,7 @@ end
 def platform_supports_lustre_on_arm?
   [node['platform'] == 'ubuntu',
    node['platform'] == 'amazon',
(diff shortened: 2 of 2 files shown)


attributes/default.rb
```diff
@@ -395,9 +395,12 @@
 )
 # Lustre repo string is built following the official doc
 # https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
+# 'centos' is used for arm and 'el' for x86_64
+fsx_arm_prefix = arm_instance? ? 'centos' : 'el'
```
rexcsn, line 399: We would want to make clear that this variable is only used for centos. To keep in line with rest of the file, how about something like `default['cfncluster']['lustre']['centos']['base_url_prefix']= arm_instance? ? 'centos' : 'el'`

We might also not need this variable depending on your approach to address the comment about the url below

attributes/default.rb
```diff
@@ -395,9 +395,12 @@
...
 default['cfncluster']['lustre']['base_url'] = value_for_platform(
   'centos' => {
     # node['kernel']['machine'] contains the architecture: 'x86_64' or 'aarch64'
+    '>=7' => "https://fsx-lustre-client-repo.s3.amazonaws.com/#{fsx_arm_prefix}/7/#{node['kernel']['machine']}",
```
rexcsn, line 403: I think we need to do the following:
* Keep `el/7.#{find_rhel_minor_version}` logic for x86 or verify with FSx team to see if we still need this logic. This logic was added in [this commit](https://github.com/aws/aws-parallelcluster-cookbook/commit/2e4932cd5fdddb60fa68ffb3de39027f300823cc) and seems necessary to support different minor versions.
* If urls for x86 and arm are very different, we might want do a conditional in `value_for_platform`, something like `'>=7' => arm_instance? ? <arm_url> : <x86_url>`
* As is, this url will be used for x86 and ARM CentOS >=7.7 anyway and we can keep the existing code by using `'default' =>` instead of `'>=7' =>`

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 **ENHANCEMENTS**
 - Add support for Ubuntu 20.04.
 - Add support for using FSx Lustre in subnet with no internet access.
-- Add support for Centos 7 ARM.
+- Add support for Centos 7 ARM, add FSx support.
```
rexcsn, line 12: Minor: we don't need to update the changelog with every commit if it's already covered by the main enhancement "Add support for Centos 7 ARM"

We can add the `skip-changelog` tag to PR to skip the changelog CI

sean-smith: Thanks!

attributes/default.rb
sean-smith: Ok I changed this to the suggest variable

attributes/default.rb
sean-smith: Lots to unpack here:

1. I removed the `>=7` option and included that logic in `default`
2. I added / kept the minor version pinning. I think this is important.
3. No need for the conditional url. 
4. Yup done

attributes/default.rb
```diff
@@ -395,11 +395,13 @@
 )
 # Lustre repo string is built following the official doc
 # https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
+# 'centos' is used for arm and 'el' for x86_64
+default['cfncluster']['lustre']['base_url_prefix'] = arm_instance? ? 'centos' : 'el'
```
rexcsn, line 399: Minor: Can we put `centos` or `centos7` in the variable name so we know this prefix is only used by Centos7? We should also point this out in the new comment above

sean-smith: ok done

rexcsn (approved):
LGTM

PR #968, opened 2021-05-20
3.0: Create SSM script to write AWS credentials for cluster admin user

gmarciani:
### Changes
1. Create SSM script to write AWS credentials for cluster admin user.
2. Create cluster admin home during node configuration.
3. Added  `scripts_dir` to `cfnconfig`. This change is required by SSM scripts to know where to find the local scripts to run.
3. (minor) Make recipe test helper check_path_permissions handle both directories and files (used to test SSM recipe)
4. (minor) Moved inclusion of recipe `cluster_admin_user_config` after home mounting through NFS. Without this change kitchen tests on compute nodes fail due to home being overwritten.

### Tests
1. Tested on a personal cluster: creation, recipe test execution and script execution.
2. Kitchen tests succeeded

### DIFF
--- a/templates/default/ssm/write_aws_credentials.sh.erb
+++ b/templates/default/ssm/write_aws_credentials.sh.erb
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# This script writes AWS credentials for the cluster admin user.
+# This script is meant to be called by SSM.
+#
+# Usage: $SCRIPT_NAME [AWS_IAM_ASSUMED_ROLE_ARN] [AWS_ACCESS_KEY_ID] [AWS_SECRET_ACCESS_KEY] [AWS_SESSION_TOKEN]
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+SCRIPT_NAME="$(basename "${BASH_SOURCE[0]}")"
+
+CLUSTER_ADMIN_USER=<%= node['cluster']['cluster_admin_user'] %>
+
+AWS_IAM_ASSUMED_ROLE_ARN=$1
+AWS_ACCESS_KEY_ID=$2
+AWS_SECRET_ACCESS_KEY=$3
+AWS_SESSION_TOKEN=$4
+
+function error () {
+  >&2 echo "ERROR: $1"
+  exit 1
+}
+
+echo "Running script: $SCRIPT_DIR/$SCRIPT_NAME"
+echo "Running as user: $(whoami)"
+echo "Running from folder: $(pwd)"
+echo "Running for cluster admin user: $CLUSTER_ADMIN_USER"
+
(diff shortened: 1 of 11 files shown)


tilne:
The script called by SSM is currently created as part of the base_config recipe. Eventually it may need to be created earlier, because the `chefPrepEnv` config set [is called before](https://github.com/aws/aws-parallelcluster/blob/wip/pcluster3/cli/src/pcluster/templates/cluster_stack.py#L1058) the `chefConfig` config set (where the `base_config` recipe is invoked). This is important because the `prep_env` recipe (invoked as part of the `chefPrepEnv` config set) includes the `prep_env_slurm` recipe, which calls the [hit_dynamodb_info helper function](https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/pcluster3/recipes/prep_env_slurm.rb#L33), which in turn [calls AWS services](https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/pcluster3/libraries/helpers.rb#L203).

I don't think this change needs to be made as part of this PR. I'm mentioning it in case there's anything to be done here that makes it easier when the change is made.

templates/default/ssm/write_aws_credentials.sh.erb
```diff
@@ -0,0 +1,60 @@
...
+echo "Running for cluster admin user: $CLUSTER_ADMIN_USER"
+
+[[ -n $AWS_IAM_ASSUMED_ROLE_ARN ]] || error "Required argument missing: AWS_IAM_ASSUMED_ROLE_ARN"
+[[ -n $AWS_ACCESS_KEY_ID ]] || error "Required argument missing: AWS_ACCESS_KEY_ID"
+[[ -n $AWS_SECRET_ACCESS_KEY ]] || error "Required argument missing: AWS_SECRET_ACCESS_KEY"
+[[ -n $AWS_SESSION_TOKEN ]] || error "Required argument missing: AWS_SESSION_TOKEN"
+
+[[ -n $(id -u $CLUSTER_ADMIN_USER 2>/dev/null) ]] || error "User does not exist: $CLUSTER_ADMIN_USER"
+
+CLUSTER_ADMIN_USER_HOME=$(sudo -u $CLUSTER_ADMIN_USER env | grep HOME | cut -d = -f 2)
```
tilne, line 35: would it make sense to `grep -E` for `^HOME=` to guard against the case where default global envs containing `HOME` have been configured on a custom AMI?

gmarciani: Sure. In case the envar HOME is not set, the script should fail rather than creating the dir tree because the AWS CLI would fail if it is not set.

gmarciani: Sure, thanks. The script should fail when the envar `HOME` is not set because the AWS CLI would fail if it is not set.

tilne (dismissed):

tilne:
Closing this due to a change in approach (😞 )

PR #989, opened 2021-06-08
3.0: Restrict access to IMDS to root and cluster admin users, only

gmarciani:
### Changes

1. Added the script `imds-access.sh` to manage access to IMDS. With this script it is possible to: 
   1. allow IMDS access for a given list of users
   2. deny IMDS access for a given list of users
   3. cleanup iptables rules for a given list of users
2. Added config recipe that uses the above script to restrict the IMDS access to root and cluster admin users, only.
3. Added kitchen tests covering the above recipe. 

### Tests
1. Tested on a personal cluster
2. Kitchen tests (ongoing)
3. Integration tests (pending)

### DIFF
--- a/files/default/imds/imds-access.sh
+++ b/files/default/imds/imds-access.sh
@@ -9,8 +9,6 @@ set -e
 # --flush                     Restore default IMDS access
 # --help                      Print this help message
 
-IPTABLES=$(whereis -b iptables | cut -d ' ' -f2)
-
 function error() {
   >&2 echo "[ERROR] $1"
   exit 1
@@ -49,7 +47,7 @@ function iptables_delete() {
     rule_args="$chain --destination $destination -j $jump -m owner --uid-owner $user"
   fi
 
-  local iptables_delete_command="$IPTABLES -D $rule_args"
+  local iptables_delete_command="iptables -D $rule_args"
 
   # Remove rules
   local should_remove=true
@@ -81,7 +79,7 @@ function iptables_add() {
     rule_args="$chain --destination $destination -j $jump -m owner --uid-owner $user"
   fi
 
-  local iptables_add_command="$IPTABLES -A $rule_args"
+  local iptables_add_command="iptables -A $rule_args"
 
   # Add rule
   eval $iptables_add_command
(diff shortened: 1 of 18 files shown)


tilne (dismissed):

recipes/base_config.rb
```diff
@@ -15,6 +15,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+include_recipe "aws-parallelcluster::setup_envars"
```
tilne, line 18: is this not needed for base_install as well?

tilne (approved):

files/default/imds/imds-access.sh
```diff
@@ -0,0 +1,164 @@
...
+
+  # Add rule: ACCEPT user, for every allowed user
+  IFS=","
+  for user in $allow_users; do
+    info "Allowing IMDS access for user: $user"
+    iptables_add $PARALLELCLUSTER_CHAIN $IMDS_IP "ACCEPT" $user
+  done
+
+  # Add rule: REJECT user, for every denied user
+  IFS=","
```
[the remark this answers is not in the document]
gmarciani: No, setting only once in the next rev

recipes/base_config.rb
gmarciani: `base_install` is always called by `base_config`, so it inherits the values of the envars.

gmarciani: Btw, I think you're right: we should add `setup_envars` because there are execution paths where only one of them is executed.

tilne (approved):

recipes/head_node_slurm_config.rb
```diff
@@ -134,15 +134,15 @@
...
+  group node['cluster']['cluster_admin_group']
   mode '0644'
 end
 
 template "#{node['cluster']['slurm_plugin_dir']}/parallelcluster_slurm_resume.conf" do
   source 'slurm/parallelcluster_slurm_resume.conf.erb'
-  owner node['cluster']['slurm']['user']
-  group node['cluster']['slurm']['group']
+  owner 'root'
+  group 'root'
```
tilne, line 145: why root for these? since suspend and resume (the only programs that read this config i think) are now run as the cluster admin user, shouldn't be owned by that user and group?

gmarciani: I agree with you.
I've preferred to leave root in this iteration because I saw the same approach in other config files, e.g. slurm config is owned by root even if slurm is run as slurm user, so I thought it was kind of standard team approach.

Fixing it.

gmarciani: I see other misalignment like this in other part of the recipes. My proposal is to fix in this PR the ones related to suspend/resume and then double-check every permissions on every dir/file in another PR.

PR #1017, opened 2021-06-29
3.0: Make PATH include a required set of directories for every user

gmarciani:
### Changes
1. Make PATH include a required set of directories for every user. In particular: /usr/local/sbin /usr/local/bin /sbin /bin /usr/sbin /usr/bin /opt/aws/bin

### Tests
1. Manual test on alinux2,centos7 and ubuntu2004

### DIFF
--- a/templates/default/profile/path.sh.erb
+++ b/templates/default/profile/path.sh.erb
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+PATH_REQUIRED_DIRECTORIES=(<%= @path_required_directories.join(' ') %>)
+
+for directory in "${PATH_REQUIRED_DIRECTORIES[@]}"; do
+    [[ ":$PATH:" == *":$directory:"* ]] || PATH="${PATH:+"$PATH:"}$directory"
+done
+
+export PATH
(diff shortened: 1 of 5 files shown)


templates/default/profile/path.sh.erb
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+DIRECTORIES=(<%= @directories.join(' ') %>)
```
enrico-usai, line 3: I think we can use a meaningful name rather than "directories", e.g. something like `path_default_folders`

tilne: i don't share the same concern about the variable name, but if we're gonna change it let's opt for something containing 'directories' rather than 'folders', since that's the term used everywhere else.

tilne (dismissed):

templates/default/profile/path.sh.erb
gmarciani: I totally agree with you that we must avoid ambiguous naming. Imho, in this case, there is no such ambiguity, but if you saw it, I'd rather name the variable `PATH_REQUIRED_DIRECTORIES` because, at the end, PATH can contains them and others.

enrico-usai (approved):

templates/default/profile/path.sh.erb
enrico-usai: I was suggesting this renaming because in the path.sh.erb we were using a variable called `directories`, coming from elsewhere and if you look at this file alone is it not clear what this `directories` variable contains. Thanks for the renaming

PR #1018, opened 2021-06-29
3.0: Fail cluster creation when IMDS lockdown is not working correctly

gmarciani:
### Changes
1. Fail cluster creation when IMDS lockdown is not working correctly

### Tests
1. (manual) Cluster creation succeeded when no bugs are introduced
2. (manual) Cluster creation fails when a bug is introduced into IMDS lockdown procedure
3. (manual) Cluster update succeeds anyway because IMDS access checks are skipped, as expected, to allow the customer to update the cluster even if she/he introduces possibly wrong customizations on IMDS access

### DIFF
--- a/recipes/imds_config.rb
+++ b/recipes/imds_config.rb
@@ -49,3 +49,5 @@ if node['cluster']['node_type'] == 'HeadNode' && node['cluster']['scheduler'] ==
     raise "head_node_imds_secured must be 'true' or 'false', but got #{node['cluster']['head_node_imds_secured']}"
   end
 end
+
+include_recipe 'aws-parallelcluster::test_imds'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Remove packer.
 - Restrict access to IMDS to root and cluster admin users, only.
 - Make PATH include required directories for every user and recipes context.
+- Fail cluster creation when IMDS lockdown is not working correctly.
 
 
 2.x.x
--- a/recipes/update_head_node.rb
+++ b/recipes/update_head_node.rb
@@ -16,5 +16,4 @@
 # limitations under the License.
 
 include_recipe 'aws-parallelcluster::setup_envars'
-include_recipe 'aws-parallelcluster::imds_config'
 include_recipe 'aws-parallelcluster::update_head_node_slurm' if node['cluster']['scheduler'] == 'slurm'


recipes/imds_config.rb
```diff
@@ -49,3 +49,7 @@
     raise "head_node_imds_secured must be 'true' or 'false', but got #{node['cluster']['head_node_imds_secured']}"
   end
 end
+
+# IMDS access must be verified on cluster creation, but skipped on cluster update
+# because customer may have introduced IMDS access customizations.
```
tilne, line 54: even if the customer introduced IMDS access customizations, the fact remains that the cluster is not configured as the parameter dictates it should be. for that reason i think we should continue to fail after update.

there are also two other alternatives that i can think:
* provide some type of warning message to the user that performed the update. (this seems hard because it would seem to require somehow passing the message through CFN)
* don't support update of the IMDS restriction parameter

recipes/imds_config.rb
```diff
@@ -49,3 +49,7 @@
...
+include_recipe 'aws-parallelcluster::test_imds' unless node['conditions']['on_update'] == 'true'
```
tilne, line 55: the test_imds recipe relies on the get_system_users library function, which is defined as:

```ruby
def get_system_users
  cmd = Mixlib::ShellOut.new("cat /etc/passwd | cut -d: -f1")
  cmd.run_command
  cmd.stdout.split(/\n+/)
end
```

will the assumption of /etc/passwd as the only source of user information be an issue for customers that store it elsewhere?

recipes/imds_config.rb
gmarciani: I think the best solution would be to return a warning message because:
1. We should support the update of something as long as it is technically feasible, as it is a useful feature for customers
2. It is informative without causing frustration to the customer: a hard fail could make the customer be stuck and it can be frustrating

A hard fail would be right in case the customer can have full fine-grained control over IMDS lockdown though the config file. Le's think about the integration with OpsWorks: in that case the customer must manually allow-list the aws user to access IMDS, so a failing update would be really frustrating because every time she/he should remove and re-add the user to make update succeed.

recipes/imds_config.rb
gmarciani: I think it is a correct assumptions at creation time because, at that time, users are managed with the standard way that stores them in `/etc/passwd`.

On update time, this could be not the case because customers could have introduced the integration with an external tool, that stores them elsewhere.

I don't know the answer for every tool that customers can use, but with OpsWorks this approach works as new users are stored in `/etc/passwd` wheres with SimpleAD it does not.

In any case, as long as the cluster configuration does not provide a way to manage users with external integrations, we cannot cover all the cases.

My proposal is to limit the check on /etc/passwd.

recipes/imds_config.rb
gmarciani: Talked with @fdm about this offline. He proposed to keep it simple and remove the support to the update. I sounded to me a bit extreme solution, but it is true that it would simplify a lot. We are going to introduce it later on if customer need it.

recipes/imds_config.rb
gmarciani: Since, as per comment above, we are going to remove the support for the update, I think it is enough to rely on /etc/passwd

tilne (approved):

PR #1028, opened 2021-07-02
3.0: Make SSH target checker depend on a static VPC CIDR list retrieved at config time

gmarciani:
### Changes
1. Make SSH target checker depend on a static VPC CIDR list retrieved at config time

Notes: 
2. I've added a kitchen test to verify the expected value of VPC CIDR list. This test does not fully protect us from errors but I think it would be better to have at least a minimum protection.

### Tests
1. Manual test
2. Kitchen tests
3. Integration tests

### DIFF
--- a/recipes/test_openssh.rb
+++ b/recipes/test_openssh.rb
@@ -2,9 +2,9 @@
 
 #
 # Cookbook Name:: aws-parallelcluster
-# Recipe:: update_head_node
+# Recipe:: test_openssh
 #
-# Copyright 2013-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
 # License. A copy of the License is located at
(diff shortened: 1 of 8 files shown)


recipes/test_openssh.rb
```diff
@@ -0,0 +1,19 @@
...
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Verifies that ssh_target_checker.sh contains the correct VPC CIDR list
+check_ssh_target_checker_vpc_cidr_list('/usr/bin/ssh_target_checker.sh', get_vpc_cidr_list)
```
tilne, line 19: this recipe doesn't appear to be included anywhere

tilne (approved):

recipes/test_openssh.rb
gmarciani: Fixed

tilne (approved):

PR #1048, opened 2021-07-16
[2.11] Use Debian patches when building SGE on AL2

tilne:


recipes/sge_install.rb
```diff
@@ -32,9 +32,14 @@
   end
 
   if node['platform'] == 'amazon'
-    # Additional patch files required for alinux2
-    cookbook_file 'sge-tcsh.patch' do
-      path '/tmp/sge-tcsh.patch'
+    al2_patches = %w[sge-armhf-java.patch sge-compiler-flags.patch sge-java-paths.patch
```
lukeseawalker, line 35: these are required for centos as well

recipes/sge_install.rb
```diff
@@ -32,9 +32,14 @@
...
+                     sge-m32_m64.patch sge-openssl-1.1.patch sge-qmake-glob-glibc227.patch
+                     sge-skip-jgdi-with-recent-java.patch sge-source-dependencies.patch
+                     sge-union-wait.patch sge-x32.patch]
+    al2_patches.each do |patch|
+      cookbook_file patch do
+        path "/tmp/#{patch}"
```
lukeseawalker, line 41: instead of importing the patches files into the codebase, what about downloading the archive that contains them and applying what's is shipped with the archive? Similar to what is done here https://github.com/aws/aws-parallelcluster-cookbook/blob/release-2.11/files/ubuntu/sge_preinstall.sh#L42, but instead of using `dpkh` command we use the `patch` command for each .patch files found in the patch folder.

PR #1050, opened 2021-07-19
Add support for mounting external file systems to an existing directory.

gmarciani:
### Changes
1. Add support for mounting external file systems to an existing directory. In particular I've added the support for mounting external file system to /home.

**THIS PR IS IN DRAFT BECAUSE IT IS UNSTABLE**

### Tests
1. Manual tests with new file system
2. 1. Manual test with existing file systems (failing when compute nodes are added to the cluster)

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -430,6 +430,7 @@ def setup_munge_head_node
       # Enforce correct permission on the key
       chmod 0600 /etc/munge/munge.key
     HEAD_CREATE_MUNGE_KEY
+    creates '/etc/munge/munge.key'
   end
 
   enable_munge_service
@@ -447,6 +448,7 @@ def share_munge_head_node
       # Copy key to shared dir
       cp /etc/munge/munge.key /home/#{node['cluster']['cluster_user']}/.munge/.munge.key
     HEAD_SHARE_MUNGE_KEY
+    creates "/home/#{node['cluster']['cluster_user']}/.munge/.munge.key"
   end
 end
 
@@ -464,6 +466,7 @@ def setup_munge_compute_node
       # Enforce correct permission on the key
       chmod 0600 /etc/munge/munge.key
     COMPUTE_MUNGE_KEY
+    creates '/etc/munge/munge.key'
   end
 
   enable_munge_service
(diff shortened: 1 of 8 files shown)


libraries/helpers.rb
```diff
@@ -430,6 +430,7 @@ def setup_munge_head_node
       # Enforce correct permission on the key
       chmod 0600 /etc/munge/munge.key
     HEAD_CREATE_MUNGE_KEY
+    creates '/etc/munge/munge.key'
```
tilne, line 433: what's the motivation for adding this? from what i understand the `creates` directive prevents the resource from running again if the file already exists. is there a reason we would expect anything under /etc to have been already generated by a customer?

libraries/helpers.rb
```diff
@@ -447,6 +448,7 @@ def share_munge_head_node
       # Copy key to shared dir
       cp /etc/munge/munge.key /home/#{node['cluster']['cluster_user']}/.munge/.munge.key
     HEAD_SHARE_MUNGE_KEY
+    creates "/home/#{node['cluster']['cluster_user']}/.munge/.munge.key"
```
tilne, line 451: same comment here. why wouldn't we re-generate it if a key already existed?

libraries/helpers.rb
```diff
@@ -750,3 +753,46 @@ def check_ssh_target_checker_vpc_cidr_list(ssh_target_checker_script, expected_c
     TEST
   end
 end
+
+def copy_to_device(directory, device_name, fs_type, mount_options)
```
tilne, line 757: i thought that during offline discussion we ruled out the idea of copying files on behalf of the user?

libraries/helpers.rb
gmarciani: We do not want to overwrite data under /home directory, in general.
Do you agree?

libraries/helpers.rb
gmarciani: Yes, the PR is in draft because at some point it did not reflect the last decisions we made. Nevertheless, I thought it would have been useful to have an initial rev of the approach.
Will fix in next rev.

libraries/helpers.rb
gmarciani: **Reason for the change**
The munge key is generated in the head node and then copied into the head node home to be shared with computed nodes. So, assuming we should not overwrite existing the content in /home, then we should not re-generate the key in /etc.

**Alternative approach (out of scope for this PR)**
Use a directory different from /home to share data between head node and compute nodes.

PR #1058, opened 2021-07-29
[nvidia-fabricmanager] lock package version to prevent upgrades

demartinofra:
nvidia-fabricmanager needs to be alwasy aligned with NVIDIA drivers version

### DIFF
--- a/recipes/tests.rb
+++ b/recipes/tests.rb
@@ -443,6 +443,32 @@ if get_nvswitches > 1
   end
 end
 
+# Verify FabricManager version is aligned with Nvidia drivers
+case node['platform_family']
+when 'rhel', 'amazon'
+  bash 'check for FabricManager' do
+    code <<-TESTFM
+      set -ex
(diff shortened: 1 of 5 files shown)


recipes/tests.rb
```diff
@@ -443,6 +443,32 @@ module load intelmpi && mpirun --help | grep '#{node['cfncluster']['intelmpi']['
   end
 end
 
+# Verify FabricManager version is aligned with Nvidia drivers
+case node['cfncluster']['os']
+when 'alinux2', 'centos7', 'centos8'
```
lukeseawalker, line 448: minor, instead of listing OS, you could have used
```
case node['platform_family']
when 'rhel', 'amazon'
   [...]
when 'debian'
   [...]
```

lukeseawalker (approved):

recipes/tests.rb
demartinofra: done

lukeseawalker (approved):

PR #1067, opened 2021-08-02
Do not install EFA if found on AMI

demartinofra:
enable_gdr setting is ignored when EFA installation is not managed by ParallelCluster

### DIFF
--- a/recipes/efa_install.rb
+++ b/recipes/efa_install.rb
@@ -16,6 +16,12 @@
 # limitations under the License.
 
 efa_tarball = "#{node['cfncluster']['sources_dir']}/aws-efa-installer.tar.gz"
+efa_installed = efa_installed?
+
+if efa_installed && !::File.exist?(efa_tarball)
+  Chef::Log.warn("Existing EFA version differs from the one shipped with ParallelCluster. Skipping ParallelCluster EFA installation and configuration. enable_gdr option will be ignored.")
+  return
+end
 
 # Get EFA Installer
 remote_file efa_tarball do
@@ -32,12 +38,12 @@ case node['platform_family']
 when 'rhel', 'amazon'
   package %w[openmpi-devel openmpi] do
     action :remove
-    not_if { ::Dir.exist?('/opt/amazon/efa') }
+    not_if { efa_installed }
   end
 when 'debian'
   package "libopenmpi-dev" do
     action :remove
-    not_if { ::Dir.exist?('/opt/amazon/efa') }
+    not_if { efa_installed }
   end
 end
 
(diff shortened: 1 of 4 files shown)


recipes/efa_install.rb
```diff
@@ -16,6 +16,12 @@
 # limitations under the License.
 
 efa_tarball = "#{node['cfncluster']['sources_dir']}/aws-efa-installer.tar.gz"
+efa_installed = efa_installed?
+
+if efa_installed && !::File.exist?(efa_tarball)
+  Chef::Log.warn("Existing EFA version differs from the one shipped with ParallelCluster. Skipping ParallelCluster EFA installation and configuration. enable_gdr option will be ignored.")
```
yuleiwan, line 22: Shall we print out existing EFA version in log?

recipes/tests.rb
```diff
@@ -341,8 +341,17 @@ module load intelmpi && mpirun --help | grep '#{node['cfncluster']['intelmpi']['
...
 ###################
-# EFA - GDR (GPUDirect RDMA)
+# EFA
 ###################
+bash 'check efa installed' do
+  cwd Chef::Config[:file_cache_path]
+  code <<-EFA
+      set -ex
+      modinfo efa
+      grep "EFA installer version: #{node['cfncluster']['efa']['installer_version']}" /opt/amazon/efa_installed_packages
```
yuleiwan, line 351: I think the test will fail if the existing EFA version mismatches `node['cfncluster']['efa']['installer_version']`, we can verify it in custom ami kitchen test.

recipes/efa_install.rb
demartinofra: I'm printing `modinfo efa` output in `efa_installed?`. Do you think this is enough or shall I add additional prints?

recipes/tests.rb
demartinofra: I believe you are right. do we have a variable in kitchen telling us whether the test is for a custom ami or not because ideally we want to do the following:
1. for official AMIs verify that EFA is installed and matches the expected version
2. for custom AMIs only verify that EFA is installed by skip version check

recipes/efa_install.rb
yuleiwan: ok, since we print out the version, it is fine.

recipes/tests.rb
yuleiwan: > do we have a variable in kitchen telling us whether the test is for a custom ami or not

yes, we have. `node['cfncluster']['is_official_ami_build']` indicates it is official ami not custom ami.

recipes/tests.rb
```diff
@@ -341,8 +341,25 @@ module load intelmpi && mpirun --help | grep '#{node['cfncluster']['intelmpi']['
 end
 
 ###################
-# EFA - GDR (GPUDirect RDMA)
+# EFA
 ###################
+if node['cfncluster']['os'].end_with?("-custom")
+  # only check EFA is installed because when found in the base AMI we skip installation
+  execute 'check efa installed' do
+    command "modinfo efa"
```
demartinofra, line 349: shall we also check that /opt/amazon/efa_installed_packages exists? without comparing the version with our installer?

hanwen-pcluste (approved):

PR #1080, opened 2021-08-17
3.0: Do not install EFA if found on AMI

enrico-usai:
`enable_gdr` setting is ignored when EFA installation is not managed by ParallelCluster

Backport of: https://github.com/aws/aws-parallelcluster-cookbook/pull/1067 + Renamed `cfncluster` to `cluster` in cookbook properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 2.11.2
 -----
 
+**CHANGES**
+When using a custom AMI with a preinstalled EFA package, no actions are taken at node bootstrap time in case GPUDirect RDMA is enabled. The original EFA package deployment is preserved as during the Image build process.
+
 **BUG FIXES**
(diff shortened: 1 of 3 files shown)


recipes/tests.rb
```diff
@@ -218,8 +218,30 @@ module load intelmpi && mpirun --help | grep '#{node['cluster']['intelmpi']['kit
 end
 
 ###################
-# EFA - GDR (GPUDirect RDMA)
+# EFA
 ###################
+if node['cluster']['os'].end_with?("-custom")
```
demartinofra, line 223: please check with @lukeseawalker if we want to keep this test here or move it to the image builder validation phase

lukeseawalker: I think it does add value if check is moved to build validation process

enrico-usai: Removed from cookbook, it will be part of another patch for the CLI.

demartinofra (approved):

PR #1090, opened 2021-08-25
[templates] Make preinstall_args and postinstall_args be arrays.

charlesg3:
*Issue #, if available:*

*Description of changes:*
- Update the `cfnconfig.erb` script and `fetch_and_run.erb` script to support multiple quoted parameters.

### DIFF
--- a/templates/default/cfnconfig.erb
+++ b/templates/default/cfnconfig.erb
@@ -1,8 +1,8 @@
 stack_name=<%= node['cluster']['stack_name'] %>
 preinstall=<%= node['cluster']['preinstall'] %>
-preinstall_args=<%= node['cluster']['preinstall_args'] %>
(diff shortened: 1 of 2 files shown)


enrico-usai:
Note: we should modify this test: https://github.com/aws/aws-parallelcluster/blob/develop/tests/integration-tests/tests/cfn-init/test_cfn_init/test_install_args_quotes/pcluster.config.yaml

1. to test single and double quotes
2. to pass multiple parameters as arrays 
3. to verify that a single parameter with spaces is considered as a single parameter

charlesg3:
This is the test that I have been working with -- integration test will be updated once this lands.
```
      Args:
        - "R curl wget"
        - two
        - 'three and still three'
```

PR #1111, opened 2021-09-24
[system-tests] Add support for testing recipes using virtual environment

charlesg3:
*Issue #, if available:*

*Description of changes:*


libraries/helpers.rb
```diff
@@ -98,11 +98,7 @@ def get_1st_partition(device)
 # Get vpc-ipv4-cidr-blocks
 #
 def get_vpc_ipv4_cidr_blocks(eth0_mac)
-  token = get_metadata_token
-  uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{eth0_mac.downcase}/vpc-ipv4-cidr-blocks")
-  vpc_ipv4_cidr_blocks = get_metadata_with_token(token, uri)
-  # Parse into array
-
+  vpc_ipv4_cidr_blocks = node['ec2']['network_interfaces_macs'][eth0_mac.downcase]['vpc_ipv4_cidr_blocks']
```
demartinofra, line 101: how is node['ec2'] populated? we need to make sure it uses IMDSv2 and not v1.

From the [code]( https://github.com/chef/ohai/blob/main/lib/ohai/mixin/ec2_metadata.rb) it seems to support IMDSv2 but we need to check what version of ohai is used by our chef-client version

libraries/helpers.rb
```diff
@@ -696,20 +699,10 @@ def get_system_users
...
-  raise('Unable to retrieve token from EC2 meta-data') if token.empty?
-
-  mac = run_command("curl http://#{imds_ip}/latest/meta-data/mac -H 'X-aws-ec2-metadata-token: #{token}' #{curl_options}")
-  raise('Unable to retrieve MAC address from EC2 meta-data') if mac.empty?
-
-  vpc_cidr_list = run_command("curl http://#{imds_ip}/latest/meta-data/network/interfaces/macs/#{mac}/vpc-ipv4-cidr-blocks -H 'X-aws-ec2-metadata-token: #{token}' #{curl_options}")
-  raise('Unable to retrieve VPC CIDR list from EC2 meta-data') if vpc_cidr_list.empty?
-
+  mac = node['ec2']['mac']
+  vpc_cidr_list = node['ec2']['network_interfaces_macs'][mac]['vpc_ipv4_cidr_blocks']
```
demartinofra, line 705: same here for IMDSv2

recipes/head_node_slurm_config.rb
```diff
@@ -61,26 +61,31 @@
...
-fetch_config_command = "#{node['cluster']['cookbook_virtualenv_path']}/bin/aws s3api get-object"\
-                       " --bucket #{node['cluster']['cluster_s3_bucket']}"\
-                       " --key #{node['cluster']['cluster_config_s3_key']}"\
-                       " --region #{node['cluster']['region']} #{node['cluster']['cluster_config_path']}"
-fetch_config_command += " --version-id #{node['cluster']['cluster_config_version']}" unless node['cluster']['cluster_config_version'].nil?
-execute "copy_cluster_config_from_s3" do
-  command fetch_config_command
-  retries 3
-  retry_delay 5
+if not virtualized?
```
demartinofra, line 64: nit: we could join all these ifs in a single one

system_tests/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# get packages that will be installed during bootstrapping
+RUN apt-get -y install vim ksh tcsh zsh libssl-dev ncurses-dev libpam-dev net-tools \
+               libhwloc-dev dkms tcl-dev automake autoconf libtool librrd-dev       \
+               libapr1-dev libconfuse-dev apache2 libboost-dev libdb-dev tcsh       \
+               libncurses5-dev libpam0g-dev libxt-dev libmotif-dev libxmu-dev       \
+               libxft-dev libhwloc-dev man-db lvm2 python r-base libblas-dev        \
+               libffi-dev libxml2-dev mdadm libgcrypt20-dev libmysqlclient-dev      \
+               libevent-dev iproute2 python3 python3-pip libatlas-base-dev          \
+               libglvnd-dev linux-headers-aws iptables libcurl4-openssl-dev         \
+               python3-parted
```
demartinofra, line 18: are you installing these packages explicitly here rather than having them install from the chef run in order to cache this layer? Trying to understand the value in keeping this duplicate list of packages

system_tests/build.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
demartinofra, line 1: where are we using this file?probably a duplicate of install_cinc

system_tests/fetch_and_run
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
demartinofra, line 1: why we need this? can't we use the file generated by the run of prep_env recipe?

system_tests/run_env.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
demartinofra, line 1: when do you use run_env?

demartinofra (commented):
Nice! I think there is a lot of value in this first effort to run our recipes in a virtualized environment. I just dropped some minor comments.

I think it could be useful to a add a readme with a few lines/commands on how to run in a virtualized env and what has to change when the cookbook changes.

libraries/helpers.rb
charlesg3: https://github.com/chef/ohai/blob/main/CHANGELOG.md -- from here it looks like they switched to using `imdsv2` in v17.0.42 ... we're using Chef version 17.2.29 -- but I'm not sure how to determine which version of ohai we're using (or if the numbers are even related?)

system_tests/Dockerfile
charlesg3: Yes -- for running repeatedly locally having these cached saves significant time.

system_tests/run_env.sh
charlesg3: Removed this.

recipes/head_node_slurm_config.rb
```diff
@@ -61,26 +61,32 @@
...
+  fetch_config_command = "#{node['cluster']['cookbook_virtualenv_path']}/bin/aws s3api get-object --bucket #{node['cluster']['cluster_s3_bucket']}"\
+                         " --key #{node['cluster']['instance_types_data_s3_key']} --region #{node['cluster']['region']} #{node['cluster']['instance_types_data_path']}"
+  execute "copy_instance_type_data_from_s3" do
+    command fetch_config_command
+    retries 3
+    retry_delay 5
+  end
+end
+
+<<<<<<< Updated upstream
```
lukeseawalker: conflict merge garbage?

cookbooks/aws-parallelcluster-config/metadata.rb
```diff
@@ -0,0 +1,26 @@
...
+issues_url 'https://github.com/aws/aws-parallelcluster-cookbook/issues'
+source_url 'https://github.com/aws/aws-parallelcluster-cookbook'
+chef_version '17.2.29'
+version '3.0.0'
+
+supports 'amazon', '>= 2'
+supports 'centos', '>= 7'
+supports 'ubuntu', '>= 18.04'
+
+depends 'apt', '~> 7.4.0'
```
lukeseawalker, line 18: minor - are all these dependencies needed or now that install and config are split this would be a subset?

charlesg3: It could be converted to a subset.

lukeseawalker (approved):

PR #1112, opened 2021-09-28
[recipes] Config install separation

charlesg3:
*Issue #, if available:*

*Description of changes:*
Separate the config and the install recipes.


recipes/base_config.rb
```diff
@@ -16,7 +16,6 @@
 # limitations under the License.
 
 include_recipe "aws-parallelcluster::setup_envars"
-include_recipe 'aws-parallelcluster::base_install'
```
demartinofra, line 19: I'm wondering if setup_envars recipe included in base_install is required in base_config too.

recipes/efa_config.rb
```diff
@@ -16,7 +16,6 @@
 # limitations under the License.
 
 # Installation recipe must be re-executed at runtime to enable GDR
-include_recipe "aws-parallelcluster::efa_install"
```
demartinofra, line 19: I believe we can't remove this because of what reported in the comment above.

recipes/base_config.rb
charlesg3: As in you think we might be able to also remove line 18 above as well?

recipes/base_install.rb
```diff
@@ -18,8 +18,6 @@
 include_recipe "aws-parallelcluster::setup_envars"
 include_recipe "aws-parallelcluster::sudoers_install"
 
-return if node['conditions']['ami_bootstrapped']
```
lukeseawalker, line 21: I think this is still used when build-image process is executed using a pcluster backed AMI as source AMI: the install recipes won't be executed again.
We can debate if this is the right place where to put this check or better move it upstream in a common entrypoint

PR #1127, opened 2021-10-07
Fix hostname configuration in case of custom DHCP domain with a space in it

demartinofra:
In case a custom DHCP option set is configured with a domain name containing a space:
```
domain-name: my.domain eu-west-1.compute.internal
domain-name-servers: 205.251.192.72, AmazonProvidedDNS
```
node['ec2']['local_hostname'] is something like `ip-192-168-112-168.my.domain eu-west-1.compute.internal`


enrico-usai:
Closed even if the patch is still valid, we need to replicate the issue and update the patch.

PR #1193, opened 2021-11-02
Upgrade slurm to 21.08

chenwany:
* `CacheGroups=0`  is removed. Remove `CacheGroups=0` from our `slurm.conf` to avoid slurm fail.
* Install `json-c-deve`l and `http-parser-devel` when using slurm
    * `apt_update` is needed for ubuntu to install  `json-c-devel` and `http-parser-devel`,  otherwise the packages can not be found.
    * In ubuntu, the package names are `libjson-c-dev` and `libhttp-parser-dev`
* add `--enable-slurmrestd`
    * `slurmrestd` is REST API interface for Slurm


cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
```diff
@@ -19,6 +19,30 @@
...
+    retries 3
+    retry_delay 5
+  end
+
+  package 'http-parser-devel' do
+    retries 3
+    retry_delay 5
+  end
+end
+
```
demartinofra, line 45: What about something like this? (you might need to double check the syntax:

```
slurm_build_deps = value_for_platformvalue_for_platform(
  'ubuntu' => {
    'default' => w%(libjson-c-dev libhttp-parser-dev),
  },
  'default' => w%(json-c-devel http-parser-devel),
)

package slurm_build_deps do
    retries 3
    retry_delay 5
end
```

chenwany: Good idea! Changed

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
```diff
@@ -19,6 +19,20 @@
   action :purge
 end
 
+apt_update if node['platform'] == 'ubuntu'
```
demartinofra: why do we need this?

chenwany: apt_update is needed for ubuntu to install `libhttp-parser-del`,  otherwise the packages can not be found.
Instruction here also mentioned `sudo apt-get update`
https://ubuntu.pkgs.org/18.04/ubuntu-main-amd64/libhttp-parser-dev_2.7.1-2_amd64.deb.html

demartinofra: don't we already run apt_update in base.rb?

chenwany: Was testing with running single recipe and it was required.  Removed it from the recipe since it was run before in other recipes

demartinofra (approved):

PR #1249, opened 2021-11-19
2.11: Update Cinc Client to 17.2.29

enrico-usai:
See same patch for 3.0.0: https://github.com/aws/aws-parallelcluster-cookbook/pull/1046

* Update Cinc Client to 17.2.29 (cherry-pick + changelog update)
* Remove third party dependency `ulimit` because resource user_ulimit is already a resource in Cinc Client. (cherry-pick)
* Execute script instead of using apt_package when installing nvidia fabricmanager for ubuntu, because for CINC17, apt_package need `version` to be the full versions (cherry-pick + using 'cfncluster' in place of 'cluster')


chenwany:
The version number also need to be update in the cli repo

chenwany (approved):

PR #1254, opened 2021-11-22
[ad_integration] Minor Changes to Directory Services Parameters

tilne:
* Assume directory service section's domain address parameter should use ldaps protocol if none is specified
* Enable the setting of arbitrary parameters in the directory service section of the SSSD configuration file
* Only write ldap_tls_ca_cert parameter if supplied
* Only enable SSH password auth on head node


hanwen-pcluste (approved):
LGTM

PR #1269, opened 2021-11-30
[ad_integration] change directory service enabled to false

hanwen-pcluste:
This commit will fix a cluster creation issue when directory service is not enabled. This bug came from the removal of `directory_service` from compute nodes' dna.json in this [PR](https://github.com/aws/aws-parallelcluster/pull/3528). With the fix, the default of `default['cluster']["directory_service"]["enabled"]` is `false`, and the directory service recipe will be skipped if `directory_service` does not present in dna.json

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -464,7 +464,7 @@ default['cluster']['head_node_private_ip'] = nil
 default['cluster']['volume'] = nil
 
 # ParallelCluster internal variables to configure active directory service
-default['cluster']["directory_service"]["enabled"] = nil
+default['cluster']["directory_service"]["enabled"] = 'false'
 default['cluster']["directory_service"]["domain_name"] = nil
 default['cluster']["directory_service"]["domain_addr"] = nil
 default['cluster']["directory_service"]["password_secret_arn"] = nil


attributes/default.rb
```diff
@@ -464,7 +464,7 @@
 default['cluster']['volume'] = nil
 
 # ParallelCluster internal variables to configure active directory service
-default['cluster']["directory_service"]["enabled"] = nil
+default['cluster']["directory_service"]["enabled"] = 'false'
```
tilne, line 467: instead of changing the default value, would it make sense to instead change the value we use to decide whether or not to skip execution of the directory service recipe ([here](https://github.com/aws/aws-parallelcluster-cookbook/blob/wip/ad_integration/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb#L20))?

hanwen-pcluste: I believe this is better. This approach is compatible with CLI passing in the `false` explicitly

tilne: okay. my comment was based on the idea that it would make sense to have only one value enable the feature as opposed to only one value disabling it. but either works.

tilne (approved):

PR #1276, opened 2021-12-01
[develop] Check whether previous cluster config exists

tilne:
`::FileUtils.identical?` throws an exception when one of the files
doesn't exist.

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
@@ -15,33 +15,34 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-unless ::FileUtils.identical?(node['cluster']['previous_cluster_config_path'], node['cluster']['cluster_config_path'])
-  # Generate pcluster specific configs
-  no_gpu = nvidia_installed? ? "" : "--no-gpu"
-  execute "generate_pcluster_slurm_configs" do
-    command "#{node['cluster']['cookbook_virtualenv_path']}/bin/python #{node['cluster']['scripts_dir']}/slurm/pcluster_slurm_config_generator.py" \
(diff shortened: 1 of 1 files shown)


charlesg3 (dismissed):

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
```diff
@@ -15,7 +15,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-unless ::FileUtils.identical?(node['cluster']['previous_cluster_config_path'], node['cluster']['cluster_config_path'])
+unless ::File.exist?(node['cluster']['previous_cluster_config_path']) && ::FileUtils.identical?(node['cluster']['previous_cluster_config_path'], node['cluster']['cluster_config_path'])
```
hanwen-pcluste: [The config is fetched](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-config/recipes/update.rb#L19) before the execution of this recipe.
If the conditional statement here is done at compile time, it will always evaluate to `false`

tilne: i was thinking that it would work when run as part of the first cluster update because fetch_config is run in a few recipes that would be executed during cluster creation. But the update flag isn't specified as part of those usages, so you're right.

hanwen-pcluste (dismissed):

hanwen-pcluste (approved):

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
tilne: fixed by adding `not_if` guards to all the resources contained in the recipe. it's an admittedly ugly solution so let me know if you know a less repetitive way to do it.

PR #1296, opened 2021-12-18
[recipes/finalize] Add run time to the output of the finalize recipe.

charlesg3:
### Description of changes
- Add the uptime in milliseconds as output at the end of the finalize recipe so that we can compare the configuration time for different settings across compute nodes.
- Enable building of EFA in system tests as the kernel version in GitHub actions is new enough
- Install the kernel modules based on the supplied kernel version (rather than the host kernel)
- Update the bootstrapped file to point to 3.1.0b1
- Add updating of the `system_tests/systemd` file to the `bump-version.sh` script to automate the above process in the future.

### Tests
* Tested locally via system tests

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/finalize.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/finalize.rb
@@ -25,3 +25,7 @@ end
 
 include_recipe 'aws-parallelcluster-scheduler-plugin::finalize' if node['cluster']['scheduler'] == 'plugin'
 include_recipe 'aws-parallelcluster-slurm::finalize' if node['cluster']['scheduler'] == 'slurm'
+
+# Log the uptime at the end of running chef scripts.
+uptime_ms = IO.read('/proc/uptime').split[0].to_f * 1000
+printf("Recipes Finish Uptime: %f\n", uptime_ms)
--- a/system_tests/systemd
+++ b/system_tests/systemd
@@ -35,7 +35,7 @@ mkdir -p /home/ubuntu/.ssh
 chown -R ubuntu:ubuntu /home/ubuntu/.ssh
 
 mkdir -p /opt/parallelcluster
-echo "aws-parallelcluster-cookbook-3.0.1" > /opt/parallelcluster/.bootstrapped
+echo "aws-parallelcluster-cookbook-3.1.0b1" > /opt/parallelcluster/.bootstrapped
 
 mkdir -p /opt/parallelcluster/scripts
 mkdir -p /etc/parallelcluster
--- a/cookbooks/aws-parallelcluster-config/recipes/base.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/base.rb
@@ -48,7 +48,7 @@ include_recipe 'aws-parallelcluster-config::chrony'
 include_recipe "aws-parallelcluster-config::nvidia"
 
 # EFA runtime configuration
-include_recipe "aws-parallelcluster-config::efa" unless virtualized?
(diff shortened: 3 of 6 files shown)


system_tests/systemd
```diff
@@ -35,7 +35,7 @@ mkdir -p /home/ubuntu/.ssh
 chown -R ubuntu:ubuntu /home/ubuntu/.ssh
 
 mkdir -p /opt/parallelcluster
-echo "aws-parallelcluster-cookbook-3.0.1" > /opt/parallelcluster/.bootstrapped
+echo "aws-parallelcluster-cookbook-3.1.0b1" > /opt/parallelcluster/.bootstrapped
```
demartinofra, line 38: what about adding this logic to the util/bump_version.sh script so that this version gets updated every time we bump the package version?

cookbooks/aws-parallelcluster-config/recipes/finalize.rb
```diff
@@ -25,3 +25,7 @@
 
 include_recipe 'aws-parallelcluster-scheduler-plugin::finalize' if node['cluster']['scheduler'] == 'plugin'
 include_recipe 'aws-parallelcluster-slurm::finalize' if node['cluster']['scheduler'] == 'slurm'
+
+# Log the uptime at the end of running chef scripts.
+uptime_ms = IO.read('/proc/uptime').split[0].to_f * 1000
+printf("Recipes Finish Uptime: %f\n", uptime_ms)
```
demartinofra, line 31: I'm wondering if this printf is evaluated at the end of the chef converge (as we probably would like it to be) or during compile phase (meaning the chef run is not over yet).

Also since the main bootstrap orchestration is managed by cloud-init an idea could be to use the timestamps present in `/var/log/cloud-init.log`

charlesg3: When I ran it in tests it was one of the last print outs -- so it seemed to capture most of the work done in chef being completed.

I'll have to check -- but we could also leverage those depending on what we're trying to capture, however the point of this print is to have the running time, not a timestamp as it's easier on the cloudwatch side of things to not have to compute the difference between timestamps (which are absolute). Unless I'm misunderstanding what is in the cloud-inIt log

system_tests/systemd
charlesg3: resolved here: https://github.com/aws/aws-parallelcluster-cookbook/pull/1296/commits/e67a98b6b61f2fad35322961ca42e8ba4752da5a

cookbooks/aws-parallelcluster-config/recipes/finalize.rb
demartinofra: I believe that print is executed in the compile phase. if so we are printing the uptime before the finalize phase runs. Consider that in case the cluster has static capacity the finalize converge phase could take minutes.

We probably can either use the log resource (https://docs.chef.io/resources/log/) or the execute one (https://docs.chef.io/resources/execute/)

lukeseawalker: Also notice that at the end of the ` cloud-init-output.log`
there is already info about execution time, both absolute timestamp and elapsed time, e.g.
```
Cloud-init v. 19.3-44.amzn2 finished at Mon, 20 Dec 2021 14:14:23 +0000. Datasource DataSourceEc2.  Up 166.94 seconds
```

charlesg3: Ah -- I will close this PR in favor of leveraging that data point from `cloud-init`.

PR #1385, opened 2022-03-07
Set SSH key generation at SSH login, switch-user and run-as-user.

gmarciani:
### Description of changes
1. Set SSH key generation at SSH login, switch-user and run-as-user.
2. Made improvements to the script that generates SSH keys.
3. Updated changelog. 

### Tests
1. Executed the following manual tests:
  1. Content of PAM config files is as expected
  3. Persistence of the above changes after `authconfig` updates
  4. Verified that home directory and SSH key got created during SSH login, switch-user (interactive and not interactive), run command as user (interactive and not interactive).
  5. Verified that the following works:
    1. SSH login as AD user, both with SSH key and with password
    2. Switch-user to root, both interactive and not interactive
    3. Switch user to AD user, both interactive and not interactive
    4. Run command as AD user, both interactive and not interactive
2. Kitchen tests
3. Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,17 @@ x.x.x
 ------
 
 **ENHANCEMENTS**
+- Execute SSH key creation alongside with the creation of HOME directory, i.e.
+  during SSH login, when switching to another user and when executing a command as another user.
 - Add support for both FQDN and LDAP Distinguished Names in property `DirectoryService.DomainName`.
-
-**BUG FIXES**
-- Fix `DirectoryService.DomainAddr` conversion to `ldap_uri` SSSD property when it contains multiples domain addresses.
+- Add script to manually update the password used to read from Active Directory,
+  according to the secret stored in AWS Secrets Manager.
 
 **CHANGES**
 - Disable deeper C-States in x86_64 official AMIs and AMIs created through `build-image` command, to guarantee high performance and low latency.
-- Add script to manually update the password used to read from Active Directory,
-  according to the secret stored in AWS Secrets Manager.
+
+**BUG FIXES**
+- Fix `DirectoryService.DomainAddr` conversion to `ldap_uri` SSSD property when it contains multiples domain addresses.
 
 3.1.2
 ------
--- a/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
@@ -113,7 +113,8 @@ if node['cluster']['node_type'] == 'HeadNode'
     sensitive true
   end
 
-  sshd_pam_config_path = '/etc/pam.d/sshd'
+  pam_services = %w(sudo su sshd)
+  pam_config_dir = "/etc/pam.d"
   generate_ssh_key_path = "#{node['cluster']['scripts_dir']}/generate_ssh_key.sh"
   ssh_key_generator_pam_config_line = "session    optional     pam_exec.so log=/var/log/parallelcluster/pam_ssh_key_generator.log #{generate_ssh_key_path}"
   if node['cluster']["directory_service"]["generate_ssh_keys_for_users"] == 'true'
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ x.x.x
 **BUG FIXES**
 - Fix `DirectoryService.DomainAddr` conversion to `ldap_uri` SSSD property when it contains multiples domain addresses.
 
+**CHANGES**
+- Execute SSH key creation alongside with the creation of HOME directory, i.e. 
+  during SSH login, when switching to another user and when executing a command as another user.
```
francesco-giordano: What about instead of an example cloud have the specific pam services: su ,ssh and sudo?

Is this a change or an enhancements?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -86,31 +87,27 @@
...
-      pattern %r{session\s+optional\s+pam_exec\.so\s+log=/var/log/parallelcluster/pam_ssh_key_generator\.log}
-      ignore_missing true
+
+    pam_services.each do |pam_service|
+      pam_config_file = "#{pam_config_dir}/#{pam_service}"
+      delete_lines "Ensure PAM service #{pam_service} is not configured to call SSH key generation script" do
+        path pam_config_file
+        pattern %r{session\s+optional\s+pam_exec\.so\s+log=/var/log/parallelcluster/pam_ssh_key_generator\.log}
+        ignore_missing true
+      end
```
francesco-giordano, line 147: Do you test this if branch?

cookbooks/aws-parallelcluster-config/templates/default/directory_service/generate_ssh_key.sh.erb
```diff
@@ -1,14 +1,23 @@
...
+
+# We should not assume that every user has its home directory in /home/$USER.
+# So, we retrieve the home directory in the most generic way.
+user_ssh_dir="$(eval echo ~${PAM_USER})/.ssh"
+
+# Skip SSH key creation if the SSH has been already configured for the user.
+# We assume that SSH has been already configured if the directory .ssh already exists in the user home.
+[ -d "${user_ssh_dir}" ] && exit 0
+
+mkdir "${user_ssh_dir}"
```
francesco-giordano: Can be useful to enforce permission in case of different umask set.

`mkdir -m 0700 "${user_ssh_dir}"` 

Here it is sure that the home dir already exist?

CHANGELOG.md
gmarciani: I see the difference a bit subtle in this case, but it could be considered more an enhancement.

gmarciani: Regarding the description. I think changelog is a customer facing document so description should be more functional rather than describing low level specifics. What do you think?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: The instruction has not been introduced by this PR, but only readapted to be executed within a each-do.
See [current implementation](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb#L110-L114).

Btw, I will re-execute a manual test on this before emerging.

cookbooks/aws-parallelcluster-config/templates/default/directory_service/generate_ssh_key.sh.erb
gmarciani: Thank for the suggestion on umask. Will include in the next rev.

Regarding the home directory, the script assumes that the user home is created because PAM executes before the system-auth triggering home creation.

If the home directory is not created for some unexpected failures, this script is not expected to take this responsibility so it is expected to fail and it would fail when trying to create the .ssh directory.

Are you suggesting to introduce an extra verification step that verifies that HOME is created and fails otherwise?
I thin k it would be good and would fail with a more clear error message

CHANGELOG.md
francesco-giordano: I agree with you

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
francesco-giordano: Thanks

cookbooks/aws-parallelcluster-config/templates/default/directory_service/generate_ssh_key.sh.erb
francesco-giordano: > If the home directory is not created for some unexpected failures, this script is not expected to take this responsibility so it is expected to fail and it would fail when trying to create the .ssh directory.

I think that it a good point since, as you explained, the home creation should be a required PAM module

> Are you suggesting to introduce an extra verification step that verifies that HOME is created and fails otherwise?
I thin k it would be good and would fail with a more clear error message

yes I think that could be useful.

gmarciani: Added

francesco-giordano (dismissed):
LGTM

francesco-giordano (dismissed):

francesco-giordano (approved):

PR #1387, opened 2022-03-09
Fix how the cluster property `DirectoryService.DomainAddr` is used to set the SSSD property `ldap_uri`

gmarciani:
### Changes
Fix the way SSSD property `ldap_uri` is configured. 
We want to fix a bug that makes the property `DomainAddr` behave differently w.r.t. what we officially documented.

According to [ParallelCluster official documentation](https://docs.aws.amazon.com/parallelcluster/latest/ug/DirectoryService-v3.html#yaml-DirectoryService-DomainAddr), the property `DomainAddr` accepts the following value:
```
The URI or URIs that point to the AD domain controller that's used as the LDAP server. The URI corresponds to the sssd-ldap parameter that's called ldap_uri. The value can be a comma separated string of URIs. To use LDAP, you must add ldap:// to the beginning of the each URI.

ldap://192.0.2.0,ldap://203.0.113.0          # LDAP
ldaps://192.0.2.0,ldaps://203.0.113.0        # LDAPS without support for certificate verification
ldaps://abcdef01234567890.corp.pcluster.com  # LDAPS with support for certificate verification
192.0.2.0,203.0.113.0                        # AWS ParallelCluster uses LDAPS by default
```

So we may think that  `192.0.2.0,203.0.113.0` would be fine, but it's not because, due to the bug we are fixing here, `ldap_uri` would be set to `ldaps://192.0.2.0,203.0.113.0`, instead of `ldaps://192.0.2.0,ldaps://203.0.113.0`, causing SSSD service start failure and the consequent cluster creation failure.

### Tests
1. Cluster creation succeeded
1. SSSD config is correct
1. SSH login as AD user works, both with password and with SSH key
1. Execution of MPI hello world as AD user succeeds
1. [PENDING] Kitchen tests succeeded
1. [PENDING] Integration tests succeeded

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ aws-parallelcluster-cookbook CHANGELOG
 
 This file is used to list changes made in each version of the AWS ParallelCluster cookbook.
 
+x.x.x
+------
+
+**BUG FIXES**
+- Fix `DirectoryService.DomainAddr` conversion to `ldap_uri` SSSD property when it contains multiples domain addresses.
+
 3.1.2
 ------
 
--- a/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
@@ -25,11 +25,15 @@ shared_directory_service_dir = "#{node['cluster']['shared_dir']}/directory_servi
 shared_sssd_conf_path = "#{shared_directory_service_dir}/sssd.conf"
 
 if node['cluster']['node_type'] == 'HeadNode'
-  # If domain_addr doesn't specify a protocol, assume it's ldaps
-  domain_addr_prefix = ''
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ aws-parallelcluster-cookbook CHANGELOG
 
 This file is used to list changes made in each version of the AWS ParallelCluster cookbook.
 
+x.x.x
+------
+
+**BUG FIXES**
+- Fix the way `DirectoryService.DomainAddr` property is interpreted to set the SSSD property `ldap_uri`.
```
enrico-usai: What about: 
- Fix `DirectoryService.DomainAddr` conversion to `ldap_uri` SSSD property when it contains multiples domain addresses.

gmarciani: I like it, updating in the next rev

enrico-usai (approved):

enrico-usai:
Please remember to backport it to release-3.1 branch too.

PR #1391, opened 2022-03-14
Add support for both FQDN and LDAP Distinguished Names in property DirectoryService.DomainName

gmarciani:
### Description of changes
Add support for both FQDN and LDAP Distinguished Names in property DirectoryService.DomainName.
In particular:
1. when the property `DirectoryService.DomainName` is a FQDN string, it will be converted into a LDAP Distinguished Name to be written into `/etc/sssd/sssd.conf` in property `ldap_search_base `.
2. otherwise, the provided value is written.

Notice that only FQDN or LDAP DN values can be received as input because the CLI only accepts those formats.

Examples:
```sh
# FQDN
corp.example.com

# LDAP DN
DC=corp,DC=example,DC=com
```

### Tests
1. Cluster creation with DomainName as FQDN + SSH login and job submission as AD user
2. Cluster update with DomainName as LDAP DN + SSH login and job submission as AD user
5. Build&Test in Jenkins

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
@@ -25,6 +25,19 @@ shared_directory_service_dir = "#{node['cluster']['shared_dir']}/directory_servi
 shared_sssd_conf_path = "#{shared_directory_service_dir}/sssd.conf"
 
 if node['cluster']['node_type'] == 'HeadNode'
+  # DomainName
+  # We can assume that DomainName can only be a FQDN or the domain section in a LDAP Distinguished Name.
+  # We can assume it because the CLI is in charge of validating it.
+  FQDN_PATTERN = /^([a-zA-Z0-9_-]+)(\.[a-zA-Z0-9_-]+)*$/.freeze
+  domain_name = node['cluster']['directory_service']['domain_name']
+  ldap_search_base =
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -25,6 +25,17 @@
 shared_sssd_conf_path = "#{shared_directory_service_dir}/sssd.conf"
 
 if node['cluster']['node_type'] == 'HeadNode'
+  # DomainName
+  FQDN_PATTERN = /^([a-zA-Z0-9_-]+)(\.[a-zA-Z0-9_-]+)*$/.freeze
```
francesco-giordano, line 31: Same as review on https://github.com/aws/aws-parallelcluster/pull/3879

gmarciani: IP addresses should not be accepted.
Will iterate on the above regexp

gmarciani: The regexp to match FQDN and LDAP DN in a strict way are so complex that I consider them hardly maintainable.
I suggest to use a more open, but maintainable regexp as it would have a better ROI.

francesco-giordano (approved):

PR #1394, opened 2022-03-16
Add kitchen test to verify Nvidia drivers and DCV GL

gmarciani:
### Description of changes
Add kitchen test to verify Nvidia drivers and DCV GL.
In particular, we want here to execute for Nvidia drivers the same tests executed during image build validation [here](https://github.com/aws/aws-parallelcluster/blob/98c53c241f5250ff4475e0b35eedbf7627322634/cli/src/pcluster/resources/imagebuilder/parallelcluster_validate.yaml#L250-L304).

### Tests
1. Manual execution of the test recipe on both an instance with GPU (g3.4xlarge) and without (t2.micro).
1. Build&Test (only on x64)

### DIFF
--- a/cookbooks/aws-parallelcluster-test/recipes/test_dcv.rb
+++ b/cookbooks/aws-parallelcluster-test/recipes/test_dcv.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-test
+# Recipe:: test_dcv
+#
+# Copyright:: 2013-2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+return unless node['cluster']['node_type'] == "HeadNode" && node['conditions']['dcv_supported'] && graphic_instance? && nvidia_installed? && dcv_gpu_accel_supported?
+
+bash "check nice-dcv-gl installation" do
+  cwd Chef::Config[:file_cache_path]
+  code <<-TEST
+    package_name="nice-dcv-gl"
+    expected_package_version="#{node['cluster']['dcv']['gl']['version']}"
+
+    echo "Testing nice-dcv-gl installation"
+    if [[ "#{node['platform_family']}" == "rhel" || "#{node['platform_family']}" == "amazon" ]]; then
+      yum list installed | grep ${package_name} | grep ${expected_package_version} || exit 1
+    else
+      apt list --installed | grep ${package_name} | grep ${expected_package_version} || exit 1
+    fi
+    echo "nice-dcv-gl test passed: the package is installed with the expected version ${expected_package_version}"
+  TEST
+end
--- a/cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
+++ b/cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-test
+# Recipe:: test_nvidia
+#
+# Copyright:: 2013-2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+bash "check Nvidia drivers" do
+  cwd Chef::Config[:file_cache_path]
+  code <<-TEST
+    expected_nvidia_driver_version="#{node['cluster']['nvidia']['driver_version']}"
+    export PATH="/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/aws/bin"
+
+    echo "Testing Nvidia driver version"
+    nvidia_driver_version=$(modinfo -F version nvidia)
+    [[ "${nvidia_driver_version}" != "${expected_nvidia_driver_ver}" ]] && "ERROR Installed Nvidia driver version ${nvidia_driver_version} but expected ${expected_nvidia_driver_version}" && exit 1
+    echo "Correctly installed Nvidia ${nvidia_driver_version}"
+
+    echo "Testing CUDA installation with nvcc"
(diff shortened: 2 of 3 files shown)


lukeseawalker (commented):
I don't think we want to execute the same test done during the build image phase.
Here instead I think we would like to check if drivers are present regardless the instance type, and in this case we cannot use the `nvidia-smi` because I expect the command to fail for non-GPU instance type.

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
```diff
@@ -0,0 +1,64 @@
...
+  code <<-TEST
+    driver_ver="#{node['cluster']['nvidia']['driver_version']}"
+    export PATH="/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/aws/bin"
+
+    echo "Testing Nvidia driver version"
+    driver_output=$(nvidia-smi | grep -E -o "Driver Version: [0-9.]+")
+    [[ "${driver_output}" != "Driver Version: ${driver_ver}" ]] && "ERROR Installed version ${driver_output} but expected ${driver_ver}" && exit 1
+    echo "Correctly installed Nvidia ${driver_output}"
+
+    if [[ "#{node['conditions']['efa_supported']}" == "true" ]]; then
```
gmarciani: Double check if we should remove this check because EFA could be installed at runtime

gmarciani: EFa is installed at installation time (install recipe not config recipe), so the check makes sense here.

lukeseawalker: but if it's installed at AMI built time, I expect the check is already done in build/test recipes, so it's a duplication here

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    echo "Testing CUDA installation with nvcc"
+    cuda_ver="#{node['cluster']['nvidia']['cuda_version']}"
+    export PATH=/usr/local/cuda-${cuda_ver}/bin:${PATH}
+    export LD_LIBRARY_PATH=/usr/local/cuda-${cuda_ver}/lib64:${LD_LIBRARY_PATH}
+    cuda_output=$(nvcc -V | grep -E -o "release [0-9]+.[0-9]+")
+    [[ "${cuda_output}" != "release ${cuda_ver}" ]] && echo "ERROR Installed version ${cuda_output} but expected ${cuda_ver}" && exit 1
+    echo "Correctly installed CUDA ${cuda_output}"
+
+    nvidia_enabled="#{node['cluster']['nvidia']['enabled'] == 'yes' || node['cluster']['nvidia']['enabled'] == true}"
```
lukeseawalker: remove this part and the below part?

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
```diff
@@ -0,0 +1,65 @@
...
+
+bash "check Nvidia drivers" do
+  cwd Chef::Config[:file_cache_path]
+  code <<-TEST
+    driver_ver="#{node['cluster']['nvidia']['driver_version']}"
+    export PATH="/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/aws/bin"
+
+    echo "Testing Nvidia driver version"
+    actual_driver_ver=$(modinfo -F version nvidia)
+    [[ "${actual_driver_ver}" != "${driver_ver}" ]] && "ERROR Installed Nvidia driver version ${driver_output} but expected ${driver_ver}" && exit 1
```
lukeseawalker: ${driver_output} not set

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
```diff
@@ -0,0 +1,65 @@
...
+    echo "Correctly installed Nvidia ${driver_output}"
+
+    if [[ "#{node['conditions']['efa_supported']}" == "true" ]]; then
+      echo "Testing Nvidia Fabric Manager version"
+      nvidia_driver_version=$(modinfo -F version nvidia)
```
lukeseawalker: reuse the same value taken before

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
gmarciani: The tests you're referring to are executed at the end of the installation phase.
Kitchen tests are executed at the end of the configuration phase, so they basically test the final result.
If this is the case I think it makes sense to check it because we are testing the final outcome of the overall process with very limited cost, so with high ROI. What do you think?

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
gmarciani: I don't get this point, sorry.
What if we introduce something bad in the config recipes that breaks the functionality?
Isn't it better to check it at the end of the config phase in a kitchen?

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
gmarciani: You're right, thanks!

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
gmarciani: You're right, thanks!

cookbooks/aws-parallelcluster-test/recipes/test_dcv.rb
```diff
@@ -0,0 +1,38 @@
...
+bash "check nice-dcv-gl installation" do
+  cwd Chef::Config[:file_cache_path]
+  code <<-TEST
+    package_name="nice-dcv-gl"
+    expected_package_version="#{node['cluster']['dcv']['gl']['version']}"
+
+    echo "Testing nice-dcv-gl installation"
+    if [[ "#{node['platform_family']}" == "rhel" || "#{node['platform_family']}" == "amazon" ]]; then
+      yum list installed | grep ${package_name} | grep ${expected_package_version} || exit 1
+      #TODO We must add version lock to nice-dcv-gl
```
lukeseawalker: can we remove the TODOs?

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
lukeseawalker: cost is not very limited, I think it does have a considerable impact on test duration.
The underlying logic is that kitchen tests are meant to test changes applied during configuration, so I don't think we should check again all the features and setup if those aren't changed during configuration phase.
Otherwise when we create a generic test, it means that we need to check everything again from scratch, while instead it should be that a test is meant to check a specific phase so we can fail fast on the process.

gmarciani: Agree on your proposal

cookbooks/aws-parallelcluster-test/recipes/test_nvidia.rb
gmarciani: Agree on your proposal.

cookbooks/aws-parallelcluster-test/recipes/test_dcv.rb
gmarciani: Agree on removing it as version lock is not required to dcvgl since we install it from a package, not from an OS repo.

lukeseawalker (approved):

PR #1410, opened 2022-04-08
Add dkms nvidia override with AL2-kernel5

francesco-giordano:
### Description of changes
* Add [override of dkms config](https://linux.die.net/man/8/dkms) for nvidia driver to compile with CC equal to `/usr/bin/gcc10-gcc` because in Amazo Linux 2 with Kernel 5 the default compiler if gcc version 7 while the OS has been compiled with gcc version 10

### Tests
* Tested: build image 

### References
* https://linux.die.net/man/8/dkms
* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start-nccl-base.html

### Checklist


cookbooks/aws-parallelcluster-install/files/amazon/dkms/nvidia.conf
```diff
@@ -0,0 +1 @@
+MAKE[0]="'make' -j2 NV_EXCLUDE_BUILD_MODULES='' KERNEL_UNAME=${kernelver} modules CC=/usr/bin/gcc10-gcc"
```
lukeseawalker, line 1: we need to be sure that `gcc10` is installed. If it's not already in the list of installed package, let's add it for this case (AL2 with Kernel 5)

lukeseawalker (approved):
Approved with a comment, I will address it in a separate PR

PR #1413, opened 2022-04-12
Make DirectoryService/AdditionalSssdConfigs be merged into default domain configuration rather than be appended. 

gmarciani:
### Description of changes
1. With this change the properties listed in `DirectoryService/AdditionalSssdConfigs` will be merged into the final SSSD configuration rather than be appended to them. Code has been refactored in a way that is easier to distinguish between mandatory properties that are not meant to be overridden, mandatory properties that are meant to be overridden by cluster properties, and optional properties that can be overridden via `DirectoryService/AdditionalSssdConfigs`.

Example

With the previous append strategy, the following configuration:

```
DirectoryService:
  DomainName: VALUE_1
  AdditionalSssdConfigs:
    ldap_search_base: VALUE_2
```

generated:

```
[domain/default]
ldap_search_base = VALUE_1
ldap_search_base = VALUE_2
```

Whereas the proposed merge strategy will generate:

```
[domain/default]
ldap_search_base = VALUE_2
```

### Tests
1. Integ test (SimpleAD/MsAD with LDAPS) locally launched
3. [ONGOING] Build&Test

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
@@ -53,6 +53,50 @@ if node['cluster']['node_type'] == 'HeadNode'
   # Then the sssd.conf file is shared through shared_sssd_conf_path to compute nodes.
   # Only contacting the secret manager from head node avoids giving permission to compute nodes to contact the secret manager.
 
+  # Configure SSSD domain properties
+  domain_properties = {
+    # Mandatory properties that must not be overridden by the user.
+    'id_provider' => 'ldap',
+    'ldap_schema' => 'AD',
+
+    # Mandatory properties that are meant to be set via dedicated cluster config properties,
+    # but that can also be overridden via DirectoryService/AdditionalSssdConfigs.
+    'ldap_uri' => ldap_uri_components.join(','),
+    'ldap_search_base' => ldap_search_base,
+    'ldap_default_bind_dn' => node['cluster']['directory_service']['domain_read_only_user'],
+    'ldap_default_authtok' => shell_out!("aws secretsmanager get-secret-value --secret-id #{node['cluster']['directory_service']['password_secret_arn']} --region #{node['cluster']['region']} --query 'SecretString' --output text").stdout.strip,
+    'ldap_tls_reqcert' => node['cluster']['directory_service']['ldap_tls_req_cert'],
+
+    # Optional properties for which we provide a default value,
+    # that are not meant to be set via dedicated cluster config properties,
+    # but that can be overridden by the user via DirectoryService/AdditionalSssdConfigs.
+    'cache_credentials' => 'True',
+    'default_shell' => '/bin/bash',
+    'fallback_homedir' => '/home/%u',
+    'ldap_id_mapping' => 'True',
+    'ldap_referrals' => 'False',
+    'use_fully_qualified_names' => 'False',
+  }
+
+  # Optional properties that are meant to be set via dedicated cluster config properties.
+  # - ldap_tls_ca_cert
+  # - ldap_access_filter
+  # - access_provider only if ldap_access_filter is specified
+
+  unless node['cluster']['directory_service']['ldap_tls_ca_cert'].eql?('NONE')
+    domain_properties['ldap_tls_cacert'] = node['cluster']['directory_service']['ldap_tls_ca_cert']
+  end
+
+  unless node['cluster']['directory_service']['ldap_access_filter'].eql?('NONE')
+    domain_properties['access_provider'] = 'ldap'
+    domain_properties['ldap_access_filter'] = node['cluster']['directory_service']['ldap_access_filter']
+  end
+
+  # Optional properties that are meant to be set via DirectoryService/AdditionalSssdConfigs
+  if node['cluster']['directory_service']['additional_sssd_configs']
+    domain_properties.merge!(node['cluster']['directory_service']['additional_sssd_configs'])
+  end
+
   # Write sssd.conf file
   template sssd_conf_path do
     source 'directory_service/sssd.conf.erb'
@@ -60,9 +104,7 @@ if node['cluster']['node_type'] == 'HeadNode'
     group 'root'
     mode '0600'
     variables(
-      ldap_default_authtok: shell_out!("aws secretsmanager get-secret-value --secret-id #{node['cluster']['directory_service']['password_secret_arn']} --region #{node['cluster']['region']} --query 'SecretString' --output text").stdout,
-      ldap_uri: ldap_uri_components.join(","),
-      ldap_search_base: ldap_search_base
+      domain_properties: domain_properties
     )
     sensitive true
   end
--- a/cookbooks/aws-parallelcluster-config/templates/default/directory_service/sssd.conf.erb
+++ b/cookbooks/aws-parallelcluster-config/templates/default/directory_service/sssd.conf.erb
@@ -1,28 +1,7 @@
 [domain/default]
-id_provider = ldap
-cache_credentials = True
-ldap_schema = AD
-ldap_uri = <%= @ldap_uri %>
-ldap_search_base = <%=  @ldap_search_base %>
-ldap_default_bind_dn = <%=  node['cluster']['directory_service']['domain_read_only_user'] %>
-ldap_default_authtok = <%= @ldap_default_authtok %>
-<% if node['cluster']['directory_service']['ldap_tls_ca_cert'] != 'NONE' %>
-ldap_tls_cacert = <%=  node['cluster']['directory_service']['ldap_tls_ca_cert'] %>
-<% end %>
-ldap_tls_reqcert = <%=  node['cluster']['directory_service']['ldap_tls_req_cert'] %>
-ldap_id_mapping = True
-fallback_homedir = /home/%u
-default_shell = /bin/bash
-use_fully_qualified_names = False
-ldap_referrals = False
-<% if node['cluster']['directory_service']['additional_sssd_configs'] %>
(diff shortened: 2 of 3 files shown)


cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -53,16 +53,57 @@
   # Then the sssd.conf file is shared through shared_sssd_conf_path to compute nodes.
   # Only contacting the secret manager from head node avoids giving permission to compute nodes to contact the secret manager.
 
+  # Configure SSSD domain properties
+  domain_properties = {
+    # Mandatory properties that must not be overridden by the customer.
```
enrico-usai: by the user.

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -53,16 +53,57 @@
...
+    # but that can also be overridden via DirectoryService/AdditionalSssdConfigs.
+    'ldap_uri' => ldap_uri_components.join(','),
+    'ldap_search_base' => ldap_search_base,
+    'ldap_default_bind_dn' => node['cluster']['directory_service']['domain_read_only_user'],
+    'ldap_default_authtok' => shell_out!("aws secretsmanager get-secret-value --secret-id #{node['cluster']['directory_service']['password_secret_arn']} --region #{node['cluster']['region']} --query 'SecretString' --output text").stdout.strip,
+    'ldap_tls_reqcert' => node['cluster']['directory_service']['ldap_tls_req_cert'],
+
+    # Optional properties for which we provide a default value,
+    # that are not meant to be set via dedicated cluster config properties,
+    # but that can be overridden by the customer via DirectoryService/AdditionalSssdConfigs.
```
enrico-usai: by the user

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -53,16 +53,57 @@
...
+  # Optional properties that are meant to be set via dedicated cluster config properties.
+  # - ldap_tls_ca_cert
+  # - ldap_access_filter
+
+  unless node['cluster']['directory_service']['ldap_tls_ca_cert'].eql?('NONE')
+    domain_properties['ldap_tls_cacert'] = node['cluster']['directory_service']['ldap_tls_ca_cert']
+  end
+
+  unless node['cluster']['directory_service']['ldap_access_filter'].eql?('NONE')
+    domain_properties['access_provider'] = 'ldap'
```
enrico-usai, line 91: what about access_provider? in which section of the domain_properties should we put it?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -53,16 +53,57 @@
...
+    domain_properties['ldap_access_filter'] = node['cluster']['directory_service']['ldap_access_filter']
+  end
+
+  # Optional properties that are meant to be set via DirectoryService/AdditionalSssdConfigs
+  if node['cluster']['directory_service']['additional_sssd_configs']
+    domain_properties.merge!(node['cluster']['directory_service']['additional_sssd_configs'])
```
enrico-usai, line 97: Can you add any link/example about the "merge" function in the PR description, to understand how it works?

cookbooks/aws-parallelcluster-config/templates/default/directory_service/sssd.conf.erb
```diff
@@ -1,28 +1,6 @@
...
-<% end %>
-ldap_tls_reqcert = <%=  node['cluster']['directory_service']['ldap_tls_req_cert'] %>
-ldap_id_mapping = True
-fallback_homedir = /home/%u
-default_shell = /bin/bash
-use_fully_qualified_names = False
-ldap_referrals = False
-<% if node['cluster']['directory_service']['additional_sssd_configs'] %>
-  <% node['cluster']['directory_service']['additional_sssd_configs'].each_pair do |param, value| %>
+<% @domain_properties.sort_by { |key| key }.to_h.each_pair do |param, value| %>
```
enrico-usai, line 3: I'd add a comment here to describe what you're doing with this line. Why are you ordering them? by which key?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -53,16 +53,57 @@
...
+    'id_provider' => 'ldap',
+    'ldap_schema' => 'AD',
+
+    # Mandatory properties that are meant to be set via dedicated cluster config properties,
+    # but that can also be overridden via DirectoryService/AdditionalSssdConfigs.
+    'ldap_uri' => ldap_uri_components.join(','),
+    'ldap_search_base' => ldap_search_base,
+    'ldap_default_bind_dn' => node['cluster']['directory_service']['domain_read_only_user'],
+    'ldap_default_authtok' => shell_out!("aws secretsmanager get-secret-value --secret-id #{node['cluster']['directory_service']['password_secret_arn']} --region #{node['cluster']['region']} --query 'SecretString' --output text").stdout.strip,
+    'ldap_tls_reqcert' => node['cluster']['directory_service']['ldap_tls_req_cert'],
```
enrico-usai, line 68: Why we don't have a correspondence between the domain properties and the cookbookn one naming? I mean `reqcert` vs `req_cert`. Can we align the cookbook var?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: Will fix in the next rev

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: Some of them have been renamed for functional purposes, e.g. `password_secret_arn` is used to set `ldap_default_authtok`.

Some other have been renamed to make them more clear to the end user, e.g. `domain_read_only_user` is used to set `ldap_default_bind_dn`.

The one that you mentioned could be fixed, but in a different PR.

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: Fixed in the next rev

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: `access_provider` is an optional SSSD configuration that must be set to `ldap` only when the user specifies also  and LDAP access filter. That's the reason why it is added here conditionally to the presence of `ldap_access_filter`.

The above comment refers also to this property, I'll specify this in the comment.

cookbooks/aws-parallelcluster-config/templates/default/directory_service/sssd.conf.erb
gmarciani: The order is alphabetical. I'm doin that to make the resulting config easier to navigate

gmarciani: Comment added in the next rev

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: Done

enrico-usai (dismissed):

enrico-usai (approved):

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -53,16 +53,58 @@
...
+    # Mandatory properties that must not be overridden by the user.
+    'id_provider' => 'ldap',
+    'ldap_schema' => 'AD',
+
+    # Mandatory properties that are meant to be set via dedicated cluster config properties,
+    # but that can also be overridden via DirectoryService/AdditionalSssdConfigs.
+    'ldap_uri' => ldap_uri_components.join(','),
+    'ldap_search_base' => ldap_search_base,
+    'ldap_default_bind_dn' => node['cluster']['directory_service']['domain_read_only_user'],
+    'ldap_default_authtok' => shell_out!("aws secretsmanager get-secret-value --secret-id #{node['cluster']['directory_service']['password_secret_arn']} --region #{node['cluster']['region']} --query 'SecretString' --output text").stdout.strip,
```
enrico-usai, line 67: Could you please add a timeout to the shell_out command?
Without the timeout it can remain stucked if for some reason you don't have access to secretsmanager endpoint.

`shell_out` [doesn't seem to support](https://docs.chef.io/infra_language/shelling_out/) the timeout option, so let's use Mixlib::ShellOut instead: https://github.com/chef/mixlib-shellout

enrico-usai (changes requested):

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: That's a very good point: thanks for the heads up!
Done in the next revision as a separate commit

gmarciani: At the end I changed my mind: it is too risky to introduce this construct in a rush.
I've tested that it does not throw an exception when the shell command fails. So, when you specify a non existing secret, cluster creation proceeds and the user ends up having a wrong sssd config file that breaks the communication with the directory.

Furthermore, I noticed that we have such a risky usage of Mixlib::ShellOut in other parts of our cookbook, thus introducing the risk of not failing when the command actually fails.

My proposal is to delay this improvement to the next iteration and to broaden it to every execution we currently have in our cookbook.

enrico-usai (approved):

PR #1446, opened 2022-05-19
Support mounting existing FSx for Ontap and OpenZFS

hanwen-pcluste:
Accept two more attributes to cookbook:
fsx_volume_junction_paths: It is a comma separated list. There are strings between the comma if it is FSx for Ontap and OpenZFS. There is empty if it is FSx for Lustre.
fsx_fs_types: It is a comma separated list. It tells the cookbook what mount method to use according to the file system (Lustre, Ontap, OpenZFS)

### Tests
This PR works together with https://github.com/aws/aws-parallelcluster/pull/4048. Tests are done in the other PR

### Checklist


attributes/default.rb
```diff
@@ -518,6 +518,8 @@
 default['cluster']['fsx_fs_ids'] = ''
 default['cluster']['fsx_dns_names'] = ''
 default['cluster']['fsx_mount_names'] = ''
+default['cluster']['fsx_fs_types'] = ''
```
gmarciani, line 521: Just for my better understanding.
If the cluster config file has the following shared file systems: `FsxLustre,FsxOntap,FsxOpenZfs,FsxLustre,FsxOntap,FsxOpenZfs,FsxLustre`, will the two new attributes have the following values?

`fsx_fs_types = 'LUSTRE,ONTAP,OPENZSF,LUSTRE,ONTAP,OPENZSF,LUSTRE'`
`fsx_volume_junction_paths = ',/a/junction/path_1,/a/junction/path_2,,/a/junction/path_3,/a/junction/path_4,'`

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -16,15 +16,21 @@
...
 # Check to see if FSx is created
 fsx_fs_id_array.each_with_index do |fsx_fs_id, index|
   fsx_shared_dir = fsx_shared_dir_array[index]
   fsx_dns_name = fsx_dns_name_array[index]
+  fsx_fs_type = fsx_fs_type_array[index]
+  fsx_volume_junction_path = fsx_volume_junction_path_array[index]
+
   # Path needs to be fully qualified, for example "shared/temp" becomes "/shared/temp"
   fsx_shared_dir = "/#{fsx_shared_dir}" unless fsx_shared_dir.start_with?('/')
+  fsx_volume_junction_path = "/#{fsx_volume_junction_path}" unless fsx_volume_junction_path.nil? || fsx_volume_junction_path.start_with?('/')
```
gmarciani, line 33: Here you are assuming that `fsx_volume_junction_path` is *nil* for FsxLustre. 
Isn't it *empty* instead of *nil*?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
   dns_name = if fsx_dns_name && !fsx_dns_name.empty?
                fsx_dns_name
              else
-               # Hardcoded DNSname only valid for filesystem created after Mar-1 2021
-               # For older filesystems, DNSname needs to be retrieved from FSx API
+               # Hardcoded DNS name only valid for lustre file systems created after Mar-1 2021
+               # For older file systems, DNS names need to be retrieved from FSx API
+               # For Ontap and OpenZFS file systems, the DNS name is always passed from CLI.
```
gmarciani: What do you mean with "passed *from* CLI"? Do you mean that you pass it *to* the CLI?
Furthermore, the comment is a bit misleading, at least to me, because it makes me think that I should expect a call to an API to retrieve the dns name, whereas the code actually hard code the DNS name given the FS id and region. May you please clarify?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
                "#{fsx_fs_id}.fsx.#{node['cluster']['region']}.amazonaws.com"
```
gmarciani, line 50: We don't do any check on `fsx_fs_id`. Shouldn't we be sure that is is not empty or malformed?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
-    group 'root'
-    mode '1777'
+    # Make sure permission is correct
+    directory fsx_shared_dir do
+      owner 'root'
+      group 'root'
+      mode '1777'
+    end
+  when 'OPENZFS'
+    # Mount FSx over NFS
```
gmarciani: This comment is kind of redundant because it does not add more information than the actual resource.

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
+      fstype 'nfs'
+      dump 0
+      pass 0
+      options options 'nfsvers=4.2'
+      action %i(mount enable)
+      retries 10
+      retry_delay 6
+    end
+  when 'ONTAP'
+    # Mount FSx over NFS
```
gmarciani: This comment is kind of redundant because it does not add more information than the actual resource.

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
-    device "#{dns_name}@tcp:/#{mount_name}"
-    fstype 'lustre'
-    dump 0
-    pass 0
-    options mount_options
-    action %i(mount enable)
-    retries 10
-    retry_delay 6
-  end
+    # Mount FSx over NFS
```
gmarciani: This comment is kind of redundant because it does not add more information than the actual resource.
By the way, we say in the comment that we are mounting FSx over NFS, but in the resource we are actually specifying *lustre* instead of *nfs* as *fstype*. Is the comment incorrect?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
+  case fsx_fs_type
+  when 'LUSTRE'
+    mount_name = fsx_mount_name_array[index]
+    mount_options = %w(defaults _netdev flock user_xattr noatime)
 
-  mount_name = fsx_mount_name_array[index]
-  mount_options = %w(defaults _netdev flock user_xattr noatime)
-
-  mount_options.concat(%w(noauto x-systemd.automount))
+    mount_options.concat(%w(noauto x-systemd.automount))
```
gmarciani: Why should we concatenate instead of including `noauto` and `x-systemd.automount` directly in the first assignment?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
+    mount fsx_shared_dir do
+      device "#{dns_name}:#{fsx_volume_junction_path}"
+      fstype 'nfs'
+      dump 0
+      pass 0
+      options options 'nfsvers=4.2'
+      action %i(mount enable)
+      retries 10
+      retry_delay 6
+    end
```
gmarciani, line 77: Why aren't we checking directory permissions?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
 
-  # Make sure permission is correct
-  directory fsx_shared_dir do
-    owner 'root'
-    group 'root'
-    mode '1777'
+    # Make sure permission is correct
+    directory fsx_shared_dir do
```
gmarciani: The check we make on directory permissions is the same for every type of FSx. What about putting the check once in common for all after the mounting operations instead of repeating it?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
gmarciani: Why should we concatenate instead of including `noauto` and `x-systemd.automount` directly in the first assignment?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -38,32 +44,64 @@
...
+      mode '1777'
+    end
+  when 'OPENZFS'
+    # Mount FSx over NFS
+    mount fsx_shared_dir do
+      device "#{dns_name}:#{fsx_volume_junction_path}"
+      fstype 'nfs'
+      dump 0
+      pass 0
+      options options 'nfsvers=4.2'
```
gmarciani: Why `options` is repeated twice? I expected to see here a string or an array of strings as the value for `options` parameters. Am I missing something?

gmarciani (changes requested):

attributes/default.rb
hanwen-pcluste: You are right!

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Fixed. This was from old code

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: The comment was incorrect. Fixed

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Fixed. Now I am wondering why it was working

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Removed

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Removed

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: OpenZFS does not allow changing the permission of the root directory. It sets the directory permission to 1777 automatically. If I change the permission again, there is exception and the recipe execution fails

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: The check on permission is only done for Ontap and Lustre

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: If `fsx_volume_junction_paths` is empty string, the result of the `split` is nil. Therefore, the check here is necessary to avoid type mismatch exception.
If it is empty string, we will prepend `/` too.

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Thanks, I improved the comment

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: We enforce the check in CLI

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Fixed

attributes/default.rb
```diff
@@ -518,6 +518,8 @@
...
+default['cluster']['fsx_volume_junction_paths'] = ''
```
gmarciani, line 522: If I understood correctly, `fsx_volume_junction_paths` can contain volume ids or junction paths.
I think we should avoid to mix different datas within the same variable.
What about splitting this into two separate fields, i.e. `fsx_volume_ids` and `fsx_junction_paths`?

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
gmarciani: I think it's worth mentioning this in a comment to avoid some else be tempted to add such check

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
gmarciani: Two code sections with the same content are already a code duplication. What about writing the permission check block once within an unless block excluding OpenZFS and adding a comment beside it explaining why the check is skipped for OpenZFS?

attributes/default.rb
hanwen-pcluste: `fsx_volume_junction_paths` only contains the path of the voluem

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Done

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
hanwen-pcluste: Done

cookbooks/aws-parallelcluster-config/recipes/fsx_mount.rb
```diff
@@ -36,31 +42,55 @@
   end
 
   dns_name = if fsx_dns_name && !fsx_dns_name.empty?
+               # DNS names of existing Lustre, Ontap, OpenZFS file systems are passed in from CLI
                fsx_dns_name
              else
-               # Hardcoded DNSname only valid for filesystem created after Mar-1 2021
-               # For older filesystems, DNSname needs to be retrieved from FSx API
+               # DNS names of newly created Lustre file systems are hardcoded here.
+               # Note the Hardcoding format is only valid for lustre file systems created after Mar-1 2021
```
gmarciani, line 49: Minor: "Hardcoding" should be "hardcoding", "lustre" should be "Lustre"

gmarciani (approved):

PR #1450, opened 2022-06-08
Add support for Slurm memory-based scheduling

jdeamicis:
### Description of changes
* First implementation of Slurm memory-based scheduling configuration via ParallelCluster:
    * Expose a new `EnableMemoryBasedScheduling` boolean parameter to users under `Scheduling/SlurmSettings`;
    * Add default `RealMemory` value for all compute nodes and `node_reg_mem_percent` by default;
    * Change `SelectTypeParameters` and `ConstrainRAMSpace` if `EnableMemoryBasedScheduling` is `true` (these parameters have been moved to slurm include conf files controlled by ParallelCluster);
* Link to impacted open issues:
    * [#2198 ](https://github.com/aws/aws-parallelcluster/issues/2198)

### Tests
* Performed tests:
    * Manual and automated tests for cluster creation with new configuration option
    * Manual and automated tests for cluster update by changing the new option
    * Basic manual tests for the Slurm memory-scheduling functionality
* Describe the added/modified tests:
    * Modify and add unit tests that check the configuration files generations
    * Add basic integration tests for cluster creation and updatability with the new configuration parameter `EnableMemoryBasedScheduling`.

### References
* Link to related PRs in other packages (i.e. cookbook, node):
    * https://github.com/aws/aws-parallelcluster/pull/4091
    * https://github.com/aws/aws-parallelcluster-cookbook/pull/1450

### Checklist
- [x] Check if documentation is impacted by this change. -> TODO


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ x.x.x
 **CHANGES**
 - Slurm: Restart `clustermgtd` and `slurmctld` daemons at cluster update time only when `Scheduling` parameters are updated in the cluster configuration.
 - Update slurmctld and slurmd systemd service files.
+- Add new configuration parameter `Scheduling/SlurmSettings/EnableMemoryBasedScheduling` to configure memory-based
```
gmarciani: Minor: missing full stop at the end of the sentence.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -74,8 +75,12 @@ def generate_slurm_config_files(
             )
         is_default_queue = False
 
-    # Generate slurm_parallelcluster.conf and slurm_parallelcluster_gres.conf
-    for template_name in ["slurm_parallelcluster.conf", "slurm_parallelcluster_gres.conf"]:
+    # Generate slurm_parallelcluster.conf, slurm_parallelcluster_gres.conf and slurm_parallelcluster_cgroup.conf
```
gmarciani, line 78: Just out of my curiosity: what is the difference between the three configuration files?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -212,6 +218,16 @@ def _vcpus(compute_resource) -> int:
     return vcpus_count if not disable_simultaneous_multithreading else (vcpus_count // threads_per_core)
 
 
+def _realmemory(compute_resource) -> int:
+    """Get the RealMemory parameter to be added to the Slurm compute node configuration."""
+    instance_type = compute_resource["InstanceType"]
+    instance_type_info = instance_types_data[instance_type]
+    memory_info = instance_type_info.get("MemoryInfo", {})
```
gmarciani, line 225: Here we set an empty dictionary as default value, but in next lines we do not behave accordingly. 
If `memory_info` is actually an empty dictionary then the next line would return `None` and the final math.floor would raise an uncaught error. I think in such cases we have two choices:
1. Assume that `instance_type` info always contains the expected keys => in this case there is no need for a default empty dictionary
2. Assume that `instance_type` may not have the expected keys => Print an error message if this is the case because using unexpected values in later instructions.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -212,6 +218,16 @@ def _vcpus(compute_resource) -> int:
...
+    ec2_memory = memory_info.get("SizeInMiB")
+    realmemory = math.floor(ec2_memory * 0.95)
```
gmarciani, line 227: I suggest to avoid magic numbers. Let's define a constant with value `0.95`, having a self-explanatory name and if necessary a comment. What about defining a cookbook attribute for it? It would also add the advantage of making it easily configurable via DevSettings extra Chef attributes

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_cgroup.conf
```diff
@@ -0,0 +1,7 @@
+# slurm_parallelcluster.conf is managed by the pcluster processes.
+# Do not modify.
+# Please add user-specific slurm configuration options in cgroup.conf
+
+{% if scaling_config.EnableMemoryBasedScheduling %}
+ConstrainRAMSpace=yes
```
gmarciani, line 6: Why not setting explicitly the opposite case: `ConstraintRAMSpace=no`?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/cgroup.conf.erb
```diff
@@ -4,3 +4,8 @@
 CgroupAutomount=yes
 TaskAffinity=no
 ConstrainCores=yes
+
+# WARNING!!! The slurm_parallelcluster_cgroup.conf file included
+# gets updated by the pcluster process, be careful
```
gmarciani: What do you mean with "be careful"? I suggest to be more specific here, e.g. do not manually change the file, do not manually change some specific properties, else....

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/slurm.conf.erb
```diff
@@ -26,10 +26,10 @@ LaunchParameters=enable_nss_slurm
 #
 # CLOUD CONFIGS OPTIONS
 <% if node['cluster']['use_private_hostname'] == 'true' or node['cluster']['dns_domain'].nil? or node['cluster']['dns_domain'].empty? -%>
-SlurmctldParameters=idle_on_node_suspend,power_save_min_interval=30
+SlurmctldParameters=idle_on_node_suspend,power_save_min_interval=30,node_reg_mem_percent=75
```
gmarciani, line 29: What about defining the reg memory percentage as a cookbook attribute? It would also add the advantage of making it easily configurable via DevSettings extra Chef attributes

CHANGELOG.md
jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: Slurm has multiple configuration files: a "base" one (`slurm.conf`), plus additional ones for additional features (`gres.conf` for the GRES management, `cgroup.conf` for managing options related to cgroup usage in Slurm, etc.).

In these files, the Slurm administrator can add include files to include additional parameters from external files. Currently we use these external include files to configure the Slurm parameters that we want to control directly via ParallelCluster. Also, we keep the include separate for each separate original Slurm config file (base one, `gres.conf`, etc.).

In this PR we introduce a new include file for the `cgroup.conf`.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: Good catch!

I would consider it quite unlikely not to have the `MemoryInfo` for an instance type.

Though, we would be on the safe side if we set Slurm's default value for `RealMemory` (which is 1) in case we don't know what to set. I will add this in my next commit.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_cgroup.conf
jdeamicis: Because `no` is already the default value for that option. But at least setting it explicitly would protect us from a (quite unlikely) change in default value coming from SchedMD. Addressed.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/cgroup.conf.erb
jdeamicis: I took the exact same message from the [template for `slurm.conf`](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-slurm/templates/default/slurm/slurm.conf.erb#L87) (with slight improvements).

Should we edit all of them (including the one for `gres.conf`) and make them consistent?

cookbooks/aws-parallelcluster-test/recipes/tests.rb
```diff
@@ -99,6 +99,9 @@
       command "ls #{node['cluster']['slurm']['install_dir']}"
       user node['cluster']['cluster_user']
     end
+    execute 'check memory cgroup' do
+      command "[[ $(grep memory /proc/cgroups | awk '{print $4}') == 1 ]]"
```
gmarciani, line 107: Why `1` as the expected value?
May you please set a more specific description instead of `check memory cgroup`?

cookbooks/aws-parallelcluster-test/recipes/tests.rb
```diff
@@ -99,6 +99,9 @@
       command "ls #{node['cluster']['slurm']['install_dir']}"
       user node['cluster']['cluster_user']
     end
+    execute 'check memory cgroup' do
```
gmarciani: Out of scope for this PR, but I think it's worth mentioning that we should avoid to enlarge this file any further and start having more specific test files. Otherwise, we will end up with a very huge file containing everything.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
gmarciani: Discussed about it with @demartinofra. We agreed on the following:
1. Assume here that EC2 always have the `MemoryInfo`, so avoid the empty dictionary fallback. Instead use `instance_type_info["MemoryInfo"]` so that it would fail with the self-explanatory error message for KeyError.
2. Add a validator in the CLI that, if the memory based scheduling is enabled, describes the instance types and checks that the required field `MemoryInfo` is actually there (may be not the case for experimental instances).

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
gmarciani: Discussed about it with @demartinofra. We agreed on having the 0.95 threshold defined in the attributes for the reasons above.

cookbooks/aws-parallelcluster-test/recipes/tests.rb
jdeamicis: The memory constraining mechanism in Slurm (used when `ConstrainRAMSpace=yes`) is based on cgroups and requires that the memory cgroup controller be enabled. This can be checked via the `/proc/cgroups` file, which contains a table formatted in this way:
```
$ column -t /proc/cgroups 
#subsys_name  hierarchy  num_cgroups  enabled
cpuset        4          1            1
cpu           9          1            1
...
memory        2          200          1
```

We need to check that the line with `memory` is there, and that the fourth field is a `1`.

enrico-usai: I'd add the expected output we're checking in a comment in the code itself

cookbooks/aws-parallelcluster-test/recipes/tests.rb
jdeamicis: Agreed. I created an internal issue for this topic.

cookbooks/aws-parallelcluster-test/recipes/tests.rb
jdeamicis: I will add it to my next commit.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: I will add it in a future PR.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/slurm.conf.erb
jdeamicis: Agreed. I will add it in a future PR.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: Agreed. I will add it in a future PR.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/cgroup.conf.erb
jdeamicis: Discussed with @gmarciani yesterday. We decided to leave the messages as they are.

gmarciani (dismissed):

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_cgroup.conf
```diff
@@ -0,0 +1,9 @@
...
+{% else %}
+ConstrainRAMSpace=no
```
demartinofra: shall we leave the implicit default as we do today so that we are backward compatible and don't risk to break users that are already adding ConstrainRAMSpace in cgroup.conf as part of their customizations?

jdeamicis: OK, after some discussion it was decided to remove the `ConstrainRAMSpace=no` in case the feature is not enabled.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ x.x.x
 - Slurm: Set `AuthInfo=cred_expire=70` to reduce the time requeued jobs must wait before starting again when nodes are not available.
 - Make `DirectoryService/AdditionalSssdConfigs` be merged into final SSSD configuration rather than be appended.
 - During cluster update set Slurm nodes state accordingly to strategy set through QueueUpdateStrategy parameter.
+- Add new configuration parameter `Scheduling/SlurmSettings/EnableMemoryBasedScheduling` to configure memory-based
+  scheduling in Slurm.
+  - moved `SelectTypeParameters` and `ConstrainRAMSpace` to the `parallelcluster_slurm*.conf` include files.
```
gmarciani: Should be "Move" (uppercase and infinitive form as above)

gmarciani (dismissed):

gmarciani (approved):

PR #1457, opened 2022-06-17
Expose internal memory-based scheduling parameters as Chef parameters

jdeamicis:
### Description of changes
* Allow overriding default values for internal parameters for memory-based scheduling via ExtraChefAttributes in DevSettings
* Add unit tests for custom value of realmemory_to_ec2memory_ratio
* Add validation for realmemory_to_ec2memory_ratio
* Add integration test for node_reg_mem_percent and realmemory_to_ec2memory_ratio

### Tests
* Added and ran integration test for memory-based scheduling in Slurm.

### References
* https://github.com/aws/aws-parallelcluster/pull/4117

### Checklist


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -285,6 +306,27 @@ def _get_metadata(metadata_path):
...
+        except ValueError:
+            raise argparse.ArgumentTypeError("The argument must be a floating point number")
+        if f < min_value or f > max_value:
+            raise argparse.ArgumentTypeError(
+                "The argument must be"
+                + " greater or equal than "
+                + str(min_value)
+                + " and"
+                + " less or equal than "
+                + str(max_value)
```
demartinofra: nitpick: a single f-string would have simplified this

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -285,6 +306,27 @@ def _get_metadata(metadata_path):
 
 
 def main():
+    def memory_ratio_float(arg):
```
demartinofra, line 310: shall we move this outside of the main()?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_queue_partition.conf
```diff
@@ -6,11 +6,11 @@
     {% set dynamic_size = compute_resource.MaxCount - compute_resource.MinCount %}
     {% set instance_type = compute_resource.InstanceType%}
     {% if static_size > 0 %}
-NodeName={{ queue_name }}-st-{{ compute_resource.Name }}-[1-{{ static_size }}] CPUs={{ compute_resource | vcpus }} RealMemory={{ compute_resource | realmemory }} State=CLOUD Feature=static,{{ instance_type }},{{ compute_resource.Name }}{% if compute_resource.Efa.Enabled %},efa{% endif %}{% if instance_type | gpus > 0 %},gpu{% if not no_gpu %} Gres=gpu:{{ instance_type | gpu_type }}:{{ instance_type | gpus }}{% endif %}{% endif %}
+NodeName={{ queue_name }}-st-{{ compute_resource.Name }}-[1-{{ static_size }}] CPUs={{ compute_resource | vcpus }} RealMemory={{ compute_resource | realmemory(realmemory_to_ec2memory_ratio) }} State=CLOUD Feature=static,{{ instance_type }},{{ compute_resource.Name }}{% if compute_resource.Efa.Enabled %},efa{% endif %}{% if instance_type | gpus > 0 %},gpu{% if not no_gpu %} Gres=gpu:{{ instance_type | gpu_type }}:{{ instance_type | gpus }}{% endif %}{% endif %}
```
demartinofra: why do we need jinja template to be aware of realmemory_to_ec2memory_ratio? we can pass a partial version of realmemory that already has the argument assigned. See https://docs.python.org/3/library/functools.html#functools.partial

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
```diff
@@ -144,7 +145,7 @@ def update_nodes_in_queue(strategy, queues)
 
 service 'slurmctld' do
   action :restart
-  not_if { ::File.exist?(node['cluster']['previous_cluster_config_path']) && !are_queues_updated? }
+  not_if { ::File.exist?(node['cluster']['previous_cluster_config_path']) && !are_queues_updated? && !is_realmemory_to_ec2memory_ratio_updated? }
```
demartinofra: tbh I'd have avoided this added complexity and code and not supported updates for something that does not even appear in our config otherwise we create misalignment with how we manage the rest of the parameters/attributes. Feel free to leave it if you want.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: I would keep this definition wherever we define everything related to the argument parsing. In this case this is done in `main()`, so I would leave this definition there.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: I'm going to do this in the next commit.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_queue_partition.conf
jdeamicis: It looks better indeed. Implementing it in the next commit.

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
jdeamicis: I will leave it for the time being. Later on we could remove the checks on `is_realmemory_to_ec2memory_ratio_updated?` if we want to drop support for updating that parameter.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
demartinofra: I'm fine, it's just that this makes the main() function (which is very generic) much longer and complex. Ideally we want arg parsing to be in separate function but we can leave as is for now

libraries/helpers.rb
```diff
@@ -570,6 +570,19 @@ def is_compute_node_bootstrap_timeout_updated?(previous_config, config)
   evaluate_compute_bootstrap_timeout(previous_config) != evaluate_compute_bootstrap_timeout(config)
 end
 
+def evaluate_realmemory_to_ec2memory_ratio(config)
+  require 'json'
+  extra_chef_attrs = config.dig("DevSettings", "Cookbook", "ExtraChefAttributes")
+  JSON.parse(extra_chef_attrs).dig("cluster", "realmemory_to_ec2memory_ratio")
```
demartinofra: I believe that is the config does not have DevSettings/Cookbook, JSON.parse(nil) will produce an error

libraries/helpers.rb
```diff
@@ -570,6 +570,19 @@ def is_compute_node_bootstrap_timeout_updated?(previous_config, config)
...
+end
+
+def is_realmemory_to_ec2memory_ratio_updated?
+  require 'yaml'
+  config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
+  previous_config = YAML.safe_load(File.read(node['cluster']['previous_cluster_config_path']))
```
demartinofra: if you want to keep the check for the update then we should avoid reading the configs in too many places and centralize this logic. fetch_config resource already takes care of loading the current cluster config in the `node['cluster']['config']`. We could extend that logic to also load the previous config and then everywhere we need these objects we access them with `node['cluster']['config']` and `node['cluster']['previous_config']`

demartinofra (approved):

PR #1479, opened 2022-07-06
[Networking] Preserve iptables for IPv6 on head node reboot

gmarciani:
### Description of changes
Preserve iptables for IPv6 on head node reboot

### Tests
1. Built custom AMI with the changes.
2. Created a cluster with the above custom AMI.
3. Verified that iptables for IPv4 are restored after head node reboot.
4. Verified that iptables for IPv6 are restored after head node reboot.


francesco-giordano (approved):
LGTM!

PR #1480, opened 2022-07-08
Add possibility to override node package at cluster creation and update

jdeamicis:
### Description of changes
* Add possibility to override node package at cluster creation and update
* Move installation of custom node package to ad-hoc recipe

### Tests
* Manually created a cluster with custom node package
* Manually updated a cluster with custom node package

### Checklist
- ~[ ] Make sure **to have added unit tests or integration tests** to cover the new/modified code.~
- ~[ ] Check if documentation is impacted by this change.~


cookbooks/aws-parallelcluster-install/recipes/base.rb
```diff
@@ -146,27 +146,7 @@
...
-      fi
-      curl --retry 3 -L -o aws-parallelcluster-node.tgz ${custom_package_url}
-      mkdir aws-parallelcluster-custom-node
-      tar -xzf aws-parallelcluster-node.tgz --directory aws-parallelcluster-custom-node
-      cd aws-parallelcluster-custom-node/*aws-parallelcluster-node-*
-      pip install .
-      deactivate
-    NODE
-  end
+  include_recipe 'aws-parallelcluster-install::parallelcluster_custom_node'
```
lukeseawalker: I'd remove `_custom_`, we are installing `parallelcluster-node`

lukeseawalker: nevermind, I see that only the portion to install custom node was moved.
Can we move everything in the new cookbook so we have only one place where the node is installed? This would be cleaner and more maintainable

cookbooks/aws-parallelcluster-config/recipes/update.rb
```diff
@@ -24,3 +24,18 @@
...
 include_recipe 'aws-parallelcluster-scheduler-plugin::update' if node['cluster']['scheduler'] == 'plugin'
+
+# Update node package - useful for development purposes only
+if !node['cluster']['custom_node_package'].nil? && !node['cluster']['custom_node_package'].empty?
+
+  service "supervisord" do
+    action :stop
+  end
+
+  include_recipe 'aws-parallelcluster-install::parallelcluster_custom_node'
```
lukeseawalker: isn't better to postpone the node installation before the slurm::update? This because during the slurm update the clustermgtd will be restarted

cookbooks/aws-parallelcluster-config/recipes/update.rb
```diff
@@ -24,3 +24,18 @@
 include_recipe 'aws-parallelcluster-config::directory_service'
 include_recipe 'aws-parallelcluster-slurm::update' if node['cluster']['scheduler'] == 'slurm'
 include_recipe 'aws-parallelcluster-scheduler-plugin::update' if node['cluster']['scheduler'] == 'plugin'
+
+# Update node package - useful for development purposes only
+if !node['cluster']['custom_node_package'].nil? && !node['cluster']['custom_node_package'].empty?
+
+  service "supervisord" do
+    action :stop
```
lukeseawalker: why stopping the whole supervisord and not just the demon provided by the node package.
Furthermore, are you sure that you are targeting the right supervisord? AFAIK it should be the one present in the cookbook virtualenv, not the one installed at OS level

jdeamicis: > why stopping the whole supervisord and not just the demon provided by the node package.

Doing it.

> Furthermore, are you sure that you are targeting the right supervisord? AFAIK it should be the one present in the cookbook virtualenv, not the one installed at OS level

By switching to restarting only clustermgtd, this point becomes obsolete.

cookbooks/aws-parallelcluster-config/recipes/update.rb
jdeamicis: That happens only if the queues are updated, right? This PR instead is for testing the node logic, which is independent of the cluster configuration.

cookbooks/aws-parallelcluster-config/recipes/update.rb
jdeamicis: Done.

PR #1497, opened 2022-08-10
[develop] Remove manual disable hyperthreading recipe

lukeseawalker:

### Description of changes
The manual disable hyperthreading logic has been moved into cloud-init, through bootcmd directive, so that the disabling is persistent against instance reboot

Please notice that the following comment on the removed code wasn't valid anymore because of the changes introduced by [this PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/784), where number of cores were taken directly from node info and not calculated from the total:
```
# NOTE: this recipe must be included after instance_slot because it may alter the values of
#       node['cpu']['total'], which would break the expected behavior when setting scheduler_slots
#       to one of the constants looked for in the above conditionals
```

### Tests
* see Tests in https://github.com/aws/aws-parallelcluster/pull/4260

### References
* https://github.com/aws/aws-parallelcluster/pull/4260

### Checklist


enrico-usai (commented):
I'd add an entry in the changelog, it might be an impacting change for the users.

francesco-giordano (approved):
LGTM!

PR #1498, opened 2022-08-12
[DFSM] Add DFSM support for ebs, efs and fsx

chenwany:
### Description of changes
* Apply QueueUpdateStrategy to the whole fleet when detect changeset contains changes with SHARED_STORAGE_UPDATE_POLICY 
* Add unmount, unexport option to manage_ebs resource
* Add unmount option to manage_efs, manage_fsx resources
* Create `shared_storages_mapping.erb` yaml template to store shared storages information before update and after update, this is required because shared_storages IDs are not contained in the changeset if the volumes are created by ParallelCluster
* Add functionality to retrieve volumes to mount and unmount by analyzing the differences between hared_storages_mapping before and after update. 
* Add conditional checker to check if mount and unmount options are required during cluster update 

### Tests
* Manually test
1. Submit a job, wait until job running
```
[ec2-user@ip-192-168-62-81 ~]$ sbatch --wrap “sleep 500”
Submitted batch job 1
[ec2-user@ip-192-168-62-81 ~]$ sinfo
PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
queue-0*     up   infinite     29  idle~ queue-0-dy-compute-resource-0-[1-29]
queue-0*     up   infinite      1    mix queue-0-st-compute-resource-0-1
````
2. update cluster with DRAIN strategy with adding the following config:
```
SharedStorage:
  - MountDir: /ebs1
    Name: /ebs1
    StorageType: Ebs
    EbsSettings:
      VolumeType: gp3
      DeletionPolicy: Delete
  - MountDir: /efs
    Name: efs
    StorageType: Efs
    EfsSettings:
      DeletionPolicy: Retain
  - FsxLustreSettings:
      DeploymentType: PERSISTENT_1
      ExportPath: s3://integ-tests-kavb046aq386bcmi/export_dir
      ImportPath: s3://integ-tests-kavb046aq386bcmi
      PerUnitStorageThroughput: 200
      StorageCapacity: 1200
    MountDir: /shared
    Name: fsx1
    StorageType: FsxLustre
  - FsxLustreSettings:
      FileSystemId: fs-0f51c84315a0dc5fc
    MountDir: /fsx_lustre_mount_dir0
    Name: existingfsx0
    StorageType: FsxLustre
  - FsxOpenZfsSettings:
      VolumeId: fsvol-0f2d533ba1670f1e9
    MountDir: /fsx_open_zfs_mount_dir0
    Name: existingopenzfs0
    StorageType: FsxOpenZfs
  - FsxOpenZfsSettings:
      VolumeId: fsvol-07896c36a757bc25e
    MountDir: /fsx_open_zfs_mount_dir1
    Name: existingopenzfs1
    StorageType: FsxOpenZfs
  - FsxOntapSettings:
      VolumeId: fsvol-05d803cb271f2ab74
    MountDir: /fsx_ontap_mount_dir0
    Name: existingontap0
    StorageType: FsxOntap
  - FsxOntapSettings:
      VolumeId: fsvol-065ba0efcaf9ba1a5
    MountDir: /fsx_ontap_mount_dir1
    Name: existingontap1
    StorageType: FsxOntap
```
3. storage info before performing update
```
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/exports
/home 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/parallelcluster/shared 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/intel 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/slurm 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/fstab
#
UUID=44322bfa-376f-498c-9617-cd51e46a67d5     /           xfs    defaults,noatime  1   1
[ec2-user@ip-192-168-62-81 ~]$ df -h
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        3.8G     0  3.8G   0% /dev
tmpfs           3.8G     0  3.8G   0% /dev/shm
tmpfs           3.8G  764K  3.8G   1% /run
tmpfs           3.8G     0  3.8G   0% /sys/fs/cgroup
/dev/nvme0n1p1   35G   16G   20G  44% /
tmpfs           766M     0  766M   0% /run/user/1000
[ec2-user@ip-192-168-62-81 ~]$ lsblk
NAME          MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
nvme0n1       259:0    0  35G  0 disk
├─nvme0n1p1   259:1    0  35G  0 part /
└─nvme0n1p128 259:2    0   1M  0 part
```
4. after a while, nodes are set to drain, job is still running
```
[ec2-user@ip-192-168-62-81 ~]$ sinfo
PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
queue-0*     up   infinite     29  idle% queue-0-dy-compute-resource-0-[1-29]
queue-0*     up   infinite      1   drng queue-0-st-compute-resource-0-1
```
5. shared storages info after update
```
[ec2-user@ip-192-168-62-81 ~]$ df -h
Filesystem                                                                    Size  Used Avail Use% Mounted on
devtmpfs                                                                      3.8G     0  3.8G   0% /dev
tmpfs                                                                         3.8G     0  3.8G   0% /dev/shm
tmpfs                                                                         3.8G  776K  3.8G   1% /run
tmpfs                                                                         3.8G     0  3.8G   0% /sys/fs/cgroup
/dev/nvme0n1p1                                                                 35G   16G   20G  44% /
tmpfs                                                                         766M     0  766M   0% /run/user/1000
/dev/nvme1n1                                                                   35G   24K   33G   1% /ebs1
fs-0f1359b9c1448bbdb.efs.us-east-1.amazonaws.com:/                            8.0E     0  8.0E   0% /efs
192.168.112.203@tcp:/pv3gdbev                                                 1.1T  7.8M  1.1T   1% /shared
192.168.34.149@tcp:/6amgdbev                                                  1.1T  4.5M  1.1T   1% /fsx_lustre_mount_dir0
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol0                     64G  256K   64G   1% /fsx_open_zfs_mount_dir0
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol1                     64G  256K   64G   1% /fsx_open_zfs_mount_dir1
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol0  9.5G  448K  9.5G   1% /fsx_ontap_mount_dir0
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol1  9.5G  448K  9.5G   1% /fsx_ontap_mount_dir1
[ec2-user@ip-192-168-62-81 ~]$ lsblk
NAME          MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
nvme0n1       259:0    0  35G  0 disk
├─nvme0n1p1   259:1    0  35G  0 part /
└─nvme0n1p128 259:2    0   1M  0 part
nvme1n1       259:3    0  35G  0 disk /ebs1
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/fstab
#
UUID=44322bfa-376f-498c-9617-cd51e46a67d5     /           xfs    defaults,noatime  1   1
UUID=5c807000-bc15-49dc-be9d-322175ffc2c7 /ebs1 ext4 _netdev 0 0
fs-0f1359b9c1448bbdb.efs.us-east-1.amazonaws.com:/ /efs nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=30,retrans=2,noresvport,_netdev 0 0
fs-051200f3d23c564b3.fsx.us-east-1.amazonaws.com@tcp:/pv3gdbev /shared lustre defaults,_netdev,flock,user_xattr,noatime,noauto,x-systemd.automount 0 0
fs-0f51c84315a0dc5fc.fsx.us-east-1.amazonaws.com@tcp:/6amgdbev /fsx_lustre_mount_dir0 lustre defaults,_netdev,flock,user_xattr,noatime,noauto,x-systemd.automount 0 0
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol0 /fsx_open_zfs_mount_dir0 nfs nfsvers=4.2 0 0
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol1 /fsx_open_zfs_mount_dir1 nfs nfsvers=4.2 0 0
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol0 /fsx_ontap_mount_dir0 nfs defaults 0 0
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol1 /fsx_ontap_mount_dir1 nfs defaults 0 0
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/exports
/home 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/parallelcluster/shared 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/intel 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/slurm 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/ebs1 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
```
6. Update cluster with changing shared_dir of EBS, EFS and FSX, which require compute fleet stop with the following the configs:
```
SharedStorage:
  - MountDir: /ebs_change
    Name: /ebs1
    StorageType: Ebs
    EbsSettings:
      VolumeType: gp3
      DeletionPolicy: Delete
  - MountDir: /efs_change
    Name: efs
    StorageType: Efs
    EfsSettings:
      DeletionPolicy: Retain
  - FsxLustreSettings:
      DeploymentType: PERSISTENT_1
      ExportPath: s3://integ-tests-kavb046aq386bcmi/export_dir
      ImportPath: s3://integ-tests-kavb046aq386bcmi
      PerUnitStorageThroughput: 200
      StorageCapacity: 1200
    MountDir: /shared
    Name: fsx1
    StorageType: FsxLustre
  - FsxLustreSettings:
      FileSystemId: fs-0f51c84315a0dc5fc
    MountDir: /fsx_lustre_mount_dir0_change
    Name: existingfsx0
    StorageType: FsxLustre
  - FsxOpenZfsSettings:
      VolumeId: fsvol-0f2d533ba1670f1e9
    MountDir: /fsx_open_zfs_mount_dir0_change
    Name: existingopenzfs0
    StorageType: FsxOpenZfs
  - FsxOpenZfsSettings:
      VolumeId: fsvol-07896c36a757bc25e
    MountDir: /fsx_open_zfs_mount_dir1_change
    Name: existingopenzfs1
    StorageType: FsxOpenZfs
  - FsxOntapSettings:
      VolumeId: fsvol-05d803cb271f2ab74
    MountDir: /fsx_ontap_mount_dir0_change
    Name: existingontap0
    StorageType: FsxOntap
  - FsxOntapSettings:
      VolumeId: fsvol-065ba0efcaf9ba1a5
    MountDir: /fsx_ontap_mount_dir1_change
    Name: existingontap1
    StorageType: FsxOntap
```
shared storages after update:
https://aws.amazon.com/amazon-linux-2/
```
[ec2-user@ip-192-168-62-81 ~]$ df -h
Filesystem                                                                    Size  Used Avail Use% Mounted on
devtmpfs                                                                      3.8G     0  3.8G   0% /dev
tmpfs                                                                         3.8G     0  3.8G   0% /dev/shm
tmpfs                                                                         3.8G  744K  3.8G   1% /run
tmpfs                                                                         3.8G     0  3.8G   0% /sys/fs/cgroup
/dev/nvme0n1p1                                                                 35G   16G   20G  44% /
tmpfs                                                                         766M     0  766M   0% /run/user/1000
192.168.112.203@tcp:/pv3gdbev                                                 1.1T  7.8M  1.1T   1% /shared
/dev/nvme1n1                                                                   35G   24K   33G   1% /ebs_change
fs-0f1359b9c1448bbdb.efs.us-east-1.amazonaws.com:/                            8.0E     0  8.0E   0% /efs_change
192.168.34.149@tcp:/6amgdbev                                                  1.1T  4.5M  1.1T   1% /fsx_lustre_mount_dir0_change
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol0                     64G  256K   64G   1% /fsx_open_zfs_mount_dir0_change
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol1                     64G  256K   64G   1% /fsx_open_zfs_mount_dir1_change
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol0  9.5G  448K  9.5G   1% /fsx_ontap_mount_dir0_change
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol1  9.5G  448K  9.5G   1% /fsx_ontap_mount_dir1_change
[ec2-user@ip-192-168-62-81 ~]$ lsblk
NAME          MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
nvme0n1       259:0    0  35G  0 disk
├─nvme0n1p1   259:1    0  35G  0 part /
└─nvme0n1p128 259:2    0   1M  0 part
nvme1n1       259:3    0  35G  0 disk /ebs_change
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/exports
/home 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/parallelcluster/shared 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/intel 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/slurm 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/ebs_change 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/fstab
#
UUID=44322bfa-376f-498c-9617-cd51e46a67d5     /           xfs    defaults,noatime  1   1
fs-051200f3d23c564b3.fsx.us-east-1.amazonaws.com@tcp:/pv3gdbev /shared lustre defaults,_netdev,flock,user_xattr,noatime,noauto,x-systemd.automount 0 0
UUID=5c807000-bc15-49dc-be9d-322175ffc2c7 /ebs_change ext4 _netdev 0 0
fs-0f1359b9c1448bbdb.efs.us-east-1.amazonaws.com:/ /efs_change nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=30,retrans=2,noresvport,_netdev 0 0
fs-0f51c84315a0dc5fc.fsx.us-east-1.amazonaws.com@tcp:/6amgdbev /fsx_lustre_mount_dir0_change lustre defaults,_netdev,flock,user_xattr,noatime,noauto,x-systemd.automount 0 0
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol0 /fsx_open_zfs_mount_dir0_change nfs nfsvers=4.2 0 0
fs-04959b907e40e1cf7.fsx.us-east-1.amazonaws.com:/fsx/vol1 /fsx_open_zfs_mount_dir1_change nfs nfsvers=4.2 0 0
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol0 /fsx_ontap_mount_dir0_change nfs defaults 0 0
svm-009350abbac782fb7.fs-0af5fcc94874d48ca.fsx.us-east-1.amazonaws.com:/vol1 /fsx_ontap_mount_dir1_change nfs defaults 0 0
```
9. Update to remove all shared storages
Shared storages after update:
```
[ec2-user@ip-192-168-62-81 ~]$ df -h
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        3.8G     0  3.8G   0% /dev
tmpfs           3.8G     0  3.8G   0% /dev/shm
tmpfs           3.8G  700K  3.8G   1% /run
tmpfs           3.8G     0  3.8G   0% /sys/fs/cgroup
/dev/nvme0n1p1   35G   16G   20G  44% /
tmpfs           766M     0  766M   0% /run/user/1000
[ec2-user@ip-192-168-62-81 ~]$ lsblk
NAME          MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
nvme0n1       259:0    0  35G  0 disk
├─nvme0n1p1   259:1    0  35G  0 part /
└─nvme0n1p128 259:2    0   1M  0 part
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/exports
/home 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/parallelcluster/shared 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/intel 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
/opt/slurm 192.168.0.0/17(rw,sync,no_root_squash) 192.168.128.0/17(rw,sync,no_root_squash)
[ec2-user@ip-192-168-62-81 ~]$ cat /etc/fstab
#
UUID=44322bfa-376f-498c-9617-cd51e46a67d5     /           xfs    defaults,noatime  1   1
```
### References

### Checklist

### CODE REVIEW

File: cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb [hunk 1]
```diff
@@ -75,12 +75,29 @@ def update_nodes(strategy, nodelist)
     end
   end
 
+  def get_all_queues
+    # Get all queue names from the cluster config
+    config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
```
lukeseawalker: config read is already done in line 124, would be possible to do it once?

gmarciani:
May you please give a brief description of the manual tests that validated the changes?

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
chenwany: Done, doing that by passing config to get_all_queues(config) and get_queues_with_changes(config)

attributes/default.rb
```diff
@@ -35,6 +35,9 @@
 default['cluster']['launch_templates_config_path'] = "#{node['cluster']['shared_dir']}/launch-templates-config.json"
 default['cluster']['instance_types_data_path'] = "#{node['cluster']['shared_dir']}/instance-types-data.json"
 default['cluster']['computefleet_status_path'] = "#{node['cluster']['shared_dir']}/computefleet-status.json"
+default['cluster']['shared_storages_mapping_path'] = "/etc/parallelcluster/shared_storages_mapping.yaml"
+default['cluster']['update_shared_storages_mapping_path'] = "/etc/parallelcluster/update_shared_storages_mapping.yaml"
```
gmarciani: Please correct me if I'm wrong.
The difference between these two files is:
1. `shared_storages_mapping.yaml`: contains info about shared storages before the update
2. `update_shared_storages_mapping.yaml`: contains info about shared storage after the update.

Is it right?

libraries/helpers.rb
```diff
@@ -619,3 +630,18 @@ def execute_command(command, user = "root", timeout = 300, raise_on_error = true
...
 end
+
+# load cluster configuration file into node object
+def load_shared_storages_mapping
+  ruby_block "load shared storages mapping during cluster update" do
+    block do
+      require 'yaml'
+      # regenerate the share storages mapping file after update
+      shared_storages_mapping = YAML.safe_load(File.read(node['cluster']['shared_storages_mapping_path']))
+      update_shared_storages_mapping = YAML.safe_load(File.read(node['cluster']['update_shared_storages_mapping_path']))
```
gmarciani: What about assigning the parsed mapping directly instead of declaring additional variables?

I mean, going from this:

```
shared_storages_mapping = YAML.safe_load(File.read(node['cluster']['shared_storages_mapping_path']))
node.default['cluster']['shared_storages_mapping'] = shared_storages_mapping
```

to this:

```
node.default['cluster']['shared_storages_mapping'] = YAML.safe_load(File.read(node['cluster']['shared_storages_mapping_path']))
```

libraries/helpers.rb
```diff
@@ -600,6 +600,17 @@ def are_queues_updated?
   config["Scheduling"] != previous_config["Scheduling"] or is_compute_node_bootstrap_timeout_updated?(previous_config, config)
 end
 
+def are_mount_or_unmount_require?
```
gmarciani: `are_mount_or_unmount_require` => `are_mount_or_unmount_required`

libraries/helpers.rb
```diff
@@ -600,6 +600,17 @@ def are_queues_updated?
...
+  require 'json'
+  change_set = JSON.load_file("#{node['cluster']['shared_dir']}/change-set.json")
+  change_set["changeSet"].each do |change|
+    next unless change["updatePolicy"] == "SHARED_STORAGE_UPDATE_POLICY"
+    return true
+  end
+  Chef::Log.info("No shared storages operation require.")
```
gmarciani: `require` => `required`

libraries/helpers.rb
```diff
@@ -600,6 +600,17 @@ def are_queues_updated?
   config["Scheduling"] != previous_config["Scheduling"] or is_compute_node_bootstrap_timeout_updated?(previous_config, config)
 end
 
+def are_mount_or_unmount_require?
+  require 'json'
+  change_set = JSON.load_file("#{node['cluster']['shared_dir']}/change-set.json")
+  change_set["changeSet"].each do |change|
+    next unless change["updatePolicy"] == "SHARED_STORAGE_UPDATE_POLICY"
```
gmarciani, line 598: So, we are assuming that every time a mount/unmount is requested then the changeset entry **always** contains `SHARED_STORAGE_UPDATE_POLICY`. Is this actually the case?

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
```diff
@@ -0,0 +1,181 @@
...
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+load_shared_storages_mapping
+
+ruby_block "get volumes to mount and unmount" do
```
gmarciani: The block title refers to volumes but the logic is not limited to EBS volumes, as it also covers EFS and FSx.
Let's use a more general title like "get storage to mount and unmount"

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
```diff
@@ -0,0 +1,181 @@
...
+      end
+    end
+    node.default['cluster']['mount_shared_dir_array'] = mount_shared_dir_array
+    node.default['cluster']['mount_vol_array'] = mount_vol_array
+
+    # get efs to unmount
+    unmount_efs_fs_id_array = []
+    unmount_efs_shared_dir_array = []
+    unless shared_storages_mapping["efs"].nil?
+      shared_storages_mapping["efs"].each do |volume|
```
gmarciani: wrong naming: EFS does not have volumes, but file systems.
If we want to use a general term that comprehends both EBS and the others, then let's use "storage"

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
```diff
@@ -0,0 +1,181 @@
...
+    node.default['cluster']['unmount_efs_fs_id_array'] = unmount_efs_fs_id_array
+
+    # get efs to mount
+    mount_efs_fs_id_array = []
+    mount_efs_shared_dir_array = []
+    unless update_shared_storages_mapping["efs"].nil?
+      update_shared_storages_mapping["efs"].each do |volume|
+        next unless shared_storages_mapping["efs"].nil? || !shared_storages_mapping["efs"].include?(volume)
+        mount_efs_shared_dir_array.push(volume["mount_dir"])
+        mount_efs_fs_id_array.push(volume["volume_id"])
```
gmarciani: The logic to detect added and removed entries is duplicated for EBS and EFS. What about reducing this code duplication?

attributes/default.rb
gmarciani: Clarified from the PR description, thanks.

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -118,3 +118,67 @@
...
+  dev_path = [] # device labels
+  dev_uuids = [] # device uuids
+
+  vol_array.each_with_index do |volumeid, index|
+    dev_path[index] = "/dev/disk/by-ebs-volumeid/#{volumeid}"
+    dev_uuids[index] = get_uuid(dev_path[index])
+
+    # Unmount and remove volume from /etc/fstab
+    mount shared_dir_array[index] do
+      device(DelayedEvaluator.new { dev_uuids[index] })
```
gmarciani, line 142: What is the purpose of using a DelayedEvaluator?

cookbooks/aws-parallelcluster-config/resources/mount_efs.rb
```diff
@@ -53,3 +53,32 @@
...
+    # Unmount EFS
+    mount efs_shared_dir do
+      device "#{efs_fs_id}.efs.#{node['cluster']['region']}.#{node['cluster']['aws_domain']}:/"
+      fstype 'nfs4'
+      options 'nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=30,retrans=2,noresvport,_netdev'
+      dump 0
+      pass 0
+      action %i(unmount disable)
+      retries 10
+      retry_delay 6
```
gmarciani, line 72: Is there a specific reason to set 10 retries with 6 seconds delay?

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
```diff
@@ -23,12 +23,12 @@
...
-  fsx_shared_dir_array = new_resource.fsx_shared_dir_array
-  fsx_dns_name_array = new_resource.fsx_dns_name_array
-  fsx_mount_name_array = new_resource.fsx_mount_name_array
-  fsx_volume_junction_path_array = new_resource.fsx_volume_junction_path_array
+  fsx_fs_id_array = new_resource.fsx_fs_id_array.dup
+  fsx_fs_type_array = new_resource.fsx_fs_type_array.dup
+  fsx_shared_dir_array = new_resource.fsx_shared_dir_array.dup
+  fsx_dns_name_array = new_resource.fsx_dns_name_array.dup
+  fsx_mount_name_array = new_resource.fsx_mount_name_array.dup
+  fsx_volume_junction_path_array = new_resource.fsx_volume_junction_path_array.dup
```
gmarciani, line 31: Why do we need to dup?

libraries/helpers.rb
chenwany: Changed

libraries/helpers.rb
chenwany: Reworded

libraries/helpers.rb
chenwany: Yes, this is the case. When changeset entry contains `SHARED_STORAGE_UPDATE_POLICY`, it means it will be compute_fleet_stop/queue_update_strategy. (Unsupported case for awsbatch is blocked by validation.) In all these cases unmount/mount is required

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: Changed

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: Thanks for the suggestion, changed all `volume` to `storage`

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: Created different functions to handle EBS, EFS, FSX mount and unmount to reduce code duplication

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
chenwany: To be aligned with the mount option from our original code. The reason may be that `dev_uuids[index]` is not available during chef compile time

cookbooks/aws-parallelcluster-config/resources/mount_efs.rb
chenwany: To be aligned with the mount option from our original code

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
chenwany: Node attributes are read-only when you do not specify which precedence level to set. We have something like `shared_dir_array[index] = "/#{shared_dir_array[index]}" unless shared_dir_array[index].start_with?('/')` in our code, which modify the value. By using dup, we are modifying the duplication of the value instead of node attribute

libraries/helpers.rb
chenwany: Done

attributes/default.rb
gmarciani: High level question on these two files: you mentioned that we we must introduce them because shared_storages IDs are not contained in the changeset if the volumes are created by ParallelCluster.

Why do we prefer to add these two new files instead of making the CLI include storage ids into the changeset?

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
gmarciani: I got your point, thanks

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
```diff
@@ -148,5 +121,79 @@ def main():
...
 
+def detach_volume(volume_id, ec2):
+    response = ec2.detach_volume(VolumeId=volume_id)
+
+    # Poll for volume to attach
+    state = response.get("State")
+    x = 0
+    while state != "available":
+        if x == 60:
+            print("Volume %s failed to detach in 300 seconds." % volume_id)
```
gmarciani: What about adding log level prefix to these messages in order to make them more discoverable when scrolling logs? Eg: `ERROR: Volume %s failed to detach in 300 seconds.`

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
```diff
@@ -148,5 +121,79 @@ def main():
...
 
 
+def detach_volume(volume_id, ec2):
+    response = ec2.detach_volume(VolumeId=volume_id)
+
+    # Poll for volume to attach
+    state = response.get("State")
+    x = 0
+    while state != "available":
+        if x == 60:
```
gmarciani: I know that this part is equivalent to the attachment one, but what about replacing magic numbers with self.explanatory variables? For instance, something like:

```
timeout = 300 # seconds
delay = 5 # seconds
elapsed = 0

...

elapsed += delay
```

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
```diff
@@ -148,5 +121,79 @@ def main():
...
+    region = region[:-1]
+
+    # Parse configuration file to read proxy settings
+    config = configparser.RawConfigParser()
+    config.read("/etc/boto.cfg")
+    proxy_config = Config()
+    if config.has_option("Boto", "proxy") and config.has_option("Boto", "proxy_port"):
+        proxy = config.get("Boto", "proxy")
+        proxy_port = config.get("Boto", "proxy_port")
+        proxy_config = Config(proxies={"https": "{0}:{1}".format(proxy, proxy_port)})
```
gmarciani: What about putting the retrieval of proxy configs in a dedicated function?

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
```diff
@@ -20,6 +20,13 @@
   not_if { ::File.exist?(node['cluster']['previous_cluster_config_path']) && !are_queues_updated? }
 end
 
+ruby_block "update_shared_storages" do
+  block do
+    run_context.include_recipe 'aws-parallelcluster-slurm::update_shared_storages'
+  end
+  only_if { are_mount_or_unmount_required? }
```
gmarciani, line 27: I do not have a strong concern, but I'd like to share with you my thoughts.
I think that `update_head_node_base` should always include `update_shared_storage` because it is a responsibility of the latter to be idempotent and know when it is appropriate to execute and update or not on the storage.
What do you think?

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
```diff
@@ -75,16 +82,31 @@ def update_nodes(strategy, nodelist)
...
     change_set = JSON.load_file("#{node['cluster']['shared_dir']}/change-set.json")
     Chef::Log.debug("Loaded change set (#{change_set})")
-    change_set["changeSet"].each do |change|
-      next unless change["updatePolicy"] == "QUEUE_UPDATE_STRATEGY"
-      queue = change["parameter"][/Scheduling\.SlurmQueues\[([^\]]*)\]/, 1]
-      Chef::Log.info("Adding queue (#{queue}) to list of queue to be updated")
-      queues.add(queue)
+    if are_mount_or_unmount_required? # Changes with SHARED_STORAGE_UPDATE_POLICY require all queues to update
+      queues = get_all_queues(config)
+      Chef::Log.info("All queues will be updated in order to add new shared storage")
```
gmarciani: In the log we mention storage addition only ("add new shared storage"), but we should refer also to the removal or more generally to an update, right?

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
```diff
@@ -0,0 +1,165 @@
...
+        shared_dir_array.push(storage["mount_dir"])
+        vol_array.push(storage["volume_id"])
+      end
+    end
+    [shared_dir_array, vol_array]
+  end
+
+  def get_efs(action)
+    # get efs to mount or unmount
+    if action == "unmount"
```
gmarciani: We use the action strings "mount" and "unmount" in more places. What about having constants for them?

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
```diff
@@ -0,0 +1,165 @@
...
+      in_shared_storages_mapping["ebs"].each do |storage|
+        next unless not_in_shared_storages_mapping["ebs"].nil? || !not_in_shared_storages_mapping["ebs"].include?(storage)
+        shared_dir_array.push(storage["mount_dir"])
+        vol_array.push(storage["volume_id"])
+      end
+    end
+    [shared_dir_array, vol_array]
+  end
+
+  def get_efs(action)
```
gmarciani, line 58: At least the first part fo this function body is equal to the one in `get_ebs` and `get_fsx`. what about putting this common logic into a dedicated function and reuse it?

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
```diff
@@ -0,0 +1,165 @@
...
+        fsx_dns_name_array.push(storage["fsx_dns_name"])
+        fsx_mount_name_array.push(storage["fsx_mount_name"])
+        fsx_volume_junction_path_array.push(storage["fsx_volume_junction_path"])
+      end
+    end
+    [fsx_fs_id_array, fsx_fs_type_array, fsx_shared_dir_array, fsx_dns_name_array, fsx_mount_name_array, fsx_volume_junction_path_array]
+  end
+end
+# unmount ebs
+manage_ebs "ebs" do
```
gmarciani: Here and below: what about having a more explanatory block descriptions, e.g. "Unmount EBS volume #{volume_id}" instead of "ebs"?

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
```diff
@@ -148,5 +121,79 @@ def main():
...
+    # Connect to AWS using boto
+    ec2 = boto3.client("ec2", region_name=region, config=proxy_config)
+
+    if attach:
+        attach_volume(volume_id, instance_id, ec2)
+    elif detach:
+        detach_volume(volume_id, ec2)
+
+
+def main():
```
gmarciani, line 184: Minor: we should consider to rename this script into `manageEbsVolume.py` because it covers both attachment and detachment now and also because it is focused on EBS, not other types of volumes (FSx Ontap volumes)

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -118,3 +118,67 @@
...
   end
 end
+
+action :unmount do
+  shared_dir_array = new_resource.shared_dir_array.dup
+  vol_array = new_resource.vol_array.dup
+
+  shared_dir_array.each_with_index do |dir, index|
+    shared_dir_array[index] = dir.strip
+    shared_dir_array[index] = "/#{shared_dir_array[index]}" unless shared_dir_array[index].start_with?('/')
```
gmarciani: This sanitization step for strings containing paths is executed in many places in our cookbook. 
What about using a dedicated function for it (in the part of code touched by this PR)?

cookbooks/aws-parallelcluster-config/resources/mount_efs.rb
```diff
@@ -13,13 +13,13 @@
 unified_mode true
```
gmarciani, line 13: What about naming the resource "manage_efs" instead of "mount_efs" (same for fsx), since its responsibility is not only mounting but also unmounting?

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
```diff
@@ -105,3 +105,59 @@
...
+               else
+                 # DNS names of newly created Lustre file systems are hardcoded here.
+                 # Note the Hardcoding format is only valid for lustre file systems created after Mar-1 2021
+                 "#{fsx_fs_id}.fsx.#{node['cluster']['region']}.amazonaws.com"
+               end
+    fsx_shared_dir = "/#{fsx_shared_dir}" unless fsx_shared_dir.start_with?('/')
+    execute 'unmount fsx' do
+      command "umount #{fsx_shared_dir}"
+      retries 3
+      retry_delay 5
```
gmarciani: In other mounting/unmounting operations we used retries=10, retry_delay=6. Unless we have a specific reasoning for the current values, what about aligning these kind of operations to the same values?

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
```diff
@@ -105,3 +105,59 @@
...
+  fsx_shared_dir_array = new_resource.fsx_shared_dir_array.dup
+  fsx_dns_name_array = new_resource.fsx_dns_name_array.dup
+  fsx_mount_name_array = new_resource.fsx_mount_name_array.dup
+  fsx_volume_junction_path_array = new_resource.fsx_volume_junction_path_array.dup
+
+  fsx_fs_id_array.each_with_index do |fsx_fs_id, index|
+    fsx_shared_dir = fsx_shared_dir_array[index]
+    fsx_dns_name = fsx_dns_name_array[index]
+    fsx_fs_type = fsx_fs_type_array[index]
+    fsx_volume_junction_path = fsx_volume_junction_path_array[index]
```
gmarciani, line 121: Not for this PR, but still an improvement to start the discussion about.
I think that we should go from having arrays of properties for the same object to having simple arrays of objects. 

So moving from something like this:

```
items_id_array.each_with_index do |item_id, index|
    attribute_1 = attribute_1_array[index]
    ...
    attribute_n = attribute_n_array[index]
end
```

to something like this:

```
items.each do |item|
    attribute_1 = item.attribute_1
    ...
    attribute_n = item.attribute_n
end
```
What do you think?

attributes/default.rb
chenwany: Changeset is a files includes the changes in configuration files. It is displayed on the console to the user what they have changed during cluster update. I don't think we want to add them to changeset.
Also, volumes id are not available when `generate_json_change_set`

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: I think it will be more confusing to put then in a dedicated function and reuse it

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: Change the description to `Remove EBS`, `Add EBS`, also do it for efs and fsx

cookbooks/aws-parallelcluster-config/resources/mount_efs.rb
chenwany: I am not in favor of renaming them because they only take care of mount and unmount option without other manage options. The chef `mount` resources also take care of `mount` and `unmount` option.
We use `manage_raid` and `manage_ebs` for ebs and raid because it also include the export and unexport nfs option.

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
chenwany: good idea, changed

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: Currently we don't have somewhere to store constant in cookbook.

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
chenwany: This logic can only be touched when using queue update strategy "DRAIN" or "TERMINATE". Removal of storages will require compute fleet stop, the logs will not show up. I changed it to `All queues will be updated in order to update shared storage`.  Does it work for you?

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
chenwany: I was try to be aligned with what is already there for Volume failed to attach.
Now I added ERROR: to everywhere in this file to be aligned

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
chenwany: Changed also for attachment one

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
chenwany: Added a dedicate function to retrieve proxy config

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
chenwany: I don't understand why we should always execute update_shared_storages?
If mount or unmount are not required, the process doesn't need to take extra time to load the two yaml files

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
chenwany: Good idea! Changed

cookbooks/aws-parallelcluster-config/resources/mount_fsx.rb
chenwany: fsx_shared_dir_array, fsx_dns_name  comes from dns.json from the user_data. I think this problem comes from that dna.json are passing attribute array as string to save some space, instead of having a map to store each of them. Maybe we are doing that because the user data has limited size. Yes in the future we may be able to improve it.

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
chenwany: renamed to `manageVolume.py`, it also cover volumes in raid

attributes/default.rb
gmarciani: I got your general point. Let's proceed with the approach you proposed (I don't see any one way door here in any case).

cookbooks/aws-parallelcluster-config/templates/default/shared_storages/shared_storages_mapping.erb
```diff
@@ -0,0 +1,32 @@
+<%# EBS %>
```
gmarciani: This template is expected to be a YAML file containing data about shared storages.
What about naming it `shared_storages_data.yaml.erb` (we already have a file on S3 with name instance-types-data on S3)?

cookbooks/aws-parallelcluster-config/resources/mount_efs.rb
gmarciani: I understand your point, but still I think it can be preferable to name them `manage_XYZ`. 
Of course I do not have a strong concern on this so feel free to proceed the way you think it's better.
Nevertheless I'd like to share with you my opinion.

First, I think the fact that Chef has a resource named mount that actually takes care of both mounting and unmounting does not mean that we could do differently if we think it's ok.

In general, I think that we have the following alternatives:
1. apply the *single responsibility principle* and reduce the scope of every resource to a specific one. In this case, we have a resource `mount_XYZ` with the only responsibility of mounting and another one `unmount_XYZ` with the only responsibility of unmounting.
2. be more general (not a strict interpretation of open-close principle since we do not use inheritance here) and make every resource ready to be extended. In this case, we have a resource `manage_XYZ` that has the responsibility to mount, unmount and possibly other actions (export unexport or whatever we will need in the future)

In this specific case, I think it would be better to apply option 2.

But still, feel totally free to proceed your way, it's just for sharing opinions :-)

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
```diff
@@ -113,40 +97,112 @@ def main():
...
     # Poll for volume to attach
     state = response.get("State")
-    x = 0
+    delay = 5  # seconds
+    elapsed = 0
+    timeout = 300  # seconds
     while state != "attached":
-        if x == 60:
-            print("Volume %s failed to mount in 300 seconds." % volume_id)
+        if elapsed == 60:
```
gmarciani: I think the logic should be as the following pseudocode, since the meaning of elapsed is still something related to the seconds that passed:

```
delay = 5
elapsed = 0
timeout = 300

while state != "my_desired_state":
    if elapsed >= timeout:
          exit
    do_stuff
    sleep(delay)
    elapsed += delay
```

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
gmarciani: My reasoning is as follows, but please consider that I do not have a strong concern, just want to share a design reasoning.

The responsibility of `update_head_node_base` is to know *which are* the components of the system whose state must converge to the one desired by the update. It should not know anything about how such updates are internally executed by each component, but simply the components to involve. In this sense, it knows that there is an update and tells every component to do their job to converge to the desired state and it trust that they do that in an idempotent way.

The responsibility of `update_component_X` (eg update_shared_storage) is to know *what* to do when a change is requested and execute what's needed to converge to the desired state. If the component understands that there is nothing to do, then it returns immediately otherwise it does its job.

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
gmarciani: Feel free to proceed the way you prefer, but in general I think it's always better to enclose a notable logic into a dedicated function and reduce the functions body.

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
gmarciani: I'm not getting this point. Why can't we do as follows in this file?

```
MOUNT_ACTION = "mount"
UNMOUNT_ACTION = "unmount"

....


if action == MOUNT_ACTION:
    do stuff
if action == UNMOUNT_ACTION:
    do stuff
```

cookbooks/aws-parallelcluster-config/templates/default/shared_storages/shared_storages_mapping.erb
chenwany: Renamed

cookbooks/aws-parallelcluster-config/resources/mount_efs.rb
chenwany: Sounds good, I am convinced. Renamed them to `manage_efs`, `manage_fsx`

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
chenwany: Sounds good, changed

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
chenwany: I think it is valid to do it either ways: return in the recipe or include the condition.
In our code, we also have places that conditionally execute recipes:
```
include_recipe 'aws-parallelcluster-config::update_computefleet_status' unless node['cluster']['scheduler'] == 'awsbatch'
```
Here it is not easy to return in the recipe because `are_mount_or_unmount_required` include  loading the change-set.json, which doesn’t exist during compile time, it need to be in the ruby bloc,  that's why I put the only_if condition here
I will discuss with Luca about it tomorrow

cookbooks/aws-parallelcluster-slurm/recipes/update_shared_storages.rb
chenwany: Oh you mean put the constant at the beginning of this file instead of a dedicated file. Yes we can do that, changed

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
gmarciani: It works, thanks for the explanation.

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
gmarciani: I got your point. Let me know what you both think about it

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
gmarciani: Thank you!

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
gmarciani: Thanks a lot!

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/attachVolume.py
gmarciani: Thanks a lot!

gmarciani (approved):
Thanks a lot for the iteration on this PR. 
I do not have any other blocking comment.

PR #1504, opened 2022-08-22
3.x Reduce the boot time DynamoDB invocation timeout from 10 minutes to 5 minutes

davprat:
### Description of changes
* Changed the shell timeout from the default of 10 minutes to 5 minutes and set connection and read timeouts on the AWS CLI call. This will use the AWS CLI standard retry logic when the connection times out. The default maximum number of retries is currently 2 which will result in 3 attempts total (ref [doc](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-modes-standard.title)). This should result in a maximum of 130sec on connection failure (30s)(3attempts)(~20sec jitter x 2).
* Removed anachronistic 'hit' prefix from method names.

### Tests
* Deployed change into a subnet where public IP's were not automatically allocated (thus preventing DynamoDB from responding to requests) and observed the total retry time to be ~5 minutes before the node logged an error and shutdown.

### Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Upgrade NVIDIA Fabric Manager to version 470.141.03.
 - Disable cron job tasks man-db and mlocate, which may have a negative impact on node performance.
 - Add support for generating Slurm Configuration files for Compute Resources with Multiple Instance Types.
+- Change DynamoDB timeout from 10 minutes to 1 minute for Slurm compute nodes.
```
enrico-usai: From the point of view of the users the change should be something like:
`Reduce timeout from 50 to 5min in case of DynamoDB connection issues at compute node bootstrap.`

cookbooks/aws-parallelcluster-slurm/recipes/init.rb
```diff
@@ -30,7 +30,7 @@
 
   ruby_block "retrieve compute node info" do
     block do
-      slurm_nodename = hit_dynamodb_info
+      slurm_nodename = dynamodb_info
       node.force_default['cluster']['slurm_nodename'] = slurm_nodename
     end
     retries 5
```
demartinofra: optional: I'm wondering if it makes sense to drop these retries or reduce it to 2 since AWS CLI should be already retrying in case of failure. Might be worth double checking the default retry for the AWS CLI command we run in dynamodb_info function

davprat: I've removed the retries here and setup the timeout parameters for the call to the AWS CLI. The AWS CLI will make 3 attempts on timeout errors.

CHANGELOG.md
davprat: Changed.

enrico-usai (approved):

PR #1508, opened 2022-08-25
[FIT] Generate API field to discriminate create-fleet API usage and save fleet config file into local folder

enrico-usai:
#### 1. Generate new "API" field to discriminate compute resources with single and multiple instance types
Note: currently run-instances compute resources don't have all the information required for fleet
so it will be possible to force usage of run-instances API for "fleet" compute resources but
not the opposite.

To be able to use "create-fleet" API for all of them we need to propagate all the information
required for create-fleet API but this requires a change in the CLI.

Cluster config with default values don't contain for example AllocationStrategy info
for queues without InstanceTypeList.

#### 2. Move fleet config file into a local folder

The file doesn't need to be shared.

### References
* https://github.com/aws/aws-parallelcluster-node/pull/436

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_fleet_config_generator.py
+++ b/cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_fleet_config_generator.py
@@ -31,8 +31,9 @@ def generate_fleet_config_file(output_file, input_file):
 
     Generate fleet-config.json
     {
-        "my-create-fleet-queue": {
-            "my-compute-resource": {
+        "my-queue": {
+            "fleet-compute-resource": {
+                "Api": "create-fleet",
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_fleet_config_generator.py
```diff
@@ -67,6 +67,7 @@ def generate_fleet_config_file(output_file, input_file):
 
                 if compute_resource_config.get("InstanceTypeList"):
                     fleet_config[queue][compute_resource] = {
+                        "API": "create-fleet",
```
demartinofra: Probably in other places we have used `Api` and not `API`. Also need to update the method docs

demartinofra (approved):

PR #1510, opened 2022-08-28
Refactor mount share storages recipes and create mount share storage resources

chenwany:
### Description of changes
Create mount share storage resources so they can be reused to support update share storages with mount and unmount actions

### Tests
* Manually test with cluster creation with EBS, EFS and FSX
* 
### References

### Checklist


demartinofra (commented):
I like the usage of resources! well done.

A couple of minor high level optional comments:
1. I'd still keep one or more separate recipes to confine the storage logic so that we don't grow the base recipe. You could have something like fs_mount if you want to keep a single one.
2. from a logical point of view it probably makes more sense to have mount resources that handle single FS and iterate over the list of FS to mount outside of the resource. What I mean is having the loop in the recipe and then for each item add a new mount_* resource.

chenwany:
> ouple of minor high level optional comments:
> 
> 1. I'd still keep one or more separate recipes to confine the storage logic so that we don't grow the base recipe. You could have something like fs_mount if you want to keep a single one.
> 2. from a logical point of view it probably makes more sense to have mount resources that handle single FS and iterate over the list of FS to mount outside of the resource. What I mean is having the loop in the recipe and then for each item add a new mount_* resource.

Thanks fdm. first comment is addressed.
For the second comment, yeah it also make sense to have mount resources that handle single FS. But for EBS, unmount action will require stop nfs before unmount all the volumes, and then start nfs after finish the process. If we want to make mount resources that handle single FS, then we need to take nfs stop and start steps outside mount resource.

demartinofra (dismissed):
nice job 🚀

cookbooks/aws-parallelcluster-config/recipes/base.rb
```diff
@@ -90,11 +90,8 @@
   mode '0644'
 end
 
-# Mount EFS directory with efs_mount recipe
-include_recipe 'aws-parallelcluster-config::efs_mount'
-
-# Mount FSx directory with fsx_mount recipe
-include_recipe 'aws-parallelcluster-config::fsx_mount'
+# Mount EFs, FSx
```
gmarciani: Minor: EFs => EFS

cookbooks/aws-parallelcluster-config/recipes/fs_mount.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+# Recipe:: fs_mount
+#
+# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: Minor: copyright year should be 2022

gmarciani (dismissed):

cookbooks/aws-parallelcluster-config/recipes/fs_mount.rb
chenwany: Done

cookbooks/aws-parallelcluster-config/recipes/base.rb
chenwany: Done

gmarciani (approved):

PR #1518, opened 2022-09-09
[develop] Set min value for resources

lukeseawalker:
### Description of changes
* Set the min value of vCPUs

  For InstanceType or single instance type in InstanceTypeList it doesn't make a difference.
  For multiple instance type in InstanceTypeList this is going to set the min value.

* Set the min value of RealMemory

  For InstanceType or single instance type in InstanceTypeList it doesn't make a difference.
  For multiple instance type in InstanceTypeList this is going to set the min value.
  
  This will allow to always use Slurm `--mem` flag during job submission.

* Set the min value of GPU and associated type for instance in the InstanceTypeList

  For InstanceType or single instance type in InstanceTypeList it doesn't make a difference.
  For multiple instance type in InstanceTypeList this is going to set the min value.

* Remove method to get first instance type
  
  List is always returned and the caller will decide to use only first element

### Tests
* unit tests added

### References
n/a

### Checklist


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -198,70 +199,104 @@ def _get_jinja_env(template_directory, realmemory_to_ec2memory_ratio):
...
-                    "Please make sure to use a custom AMI with the appropriate drivers in order to leverage "
-                    "GPUs functionalities"
-                )
-
+        return [compute_resource["InstanceType"]]
+
+
+def _get_min_gpu_count(instance_types) -> int:
+    """Return min value for GPU in the instance type list."""
+    min_gpu_count = None
```
EddyMM, line 215: Since we're getting the min value for various properties of a list of InstanceTypeInfo objects,
Can we leverage the built-in [min()](https://docs.python.org/3/library/functions.html#min) and for the key we pass the function that extracts the property per instance-type (e.g. the GPU count or a default)?

Such that we can reuse the existing `_gpu_count(instance_type)` function.

```
def _get_min_gpu_count(instance_types) -> int:
    return min(instance_types, _gpu_count)

def _gpu_count(instance_type):
    ...
```

test/unit/slurm/test_slurm_config_generator/test_generating_slurm_config_flexible_instance_types/expected_outputs/pcluster/slurm_parallelcluster_queue6_partition.conf
```diff
@@ -0,0 +1,7 @@
+# This file is automatically generated by pcluster
+
+NodeName=queue6-st-cr1-[1-2] CPUs=16 RealMemory=237363 State=CLOUD Feature=static,cr1,gpu Gres=gpu:a10g:4
```
EddyMM: The smallest GPU count was selected but the GPU Name used for the Gres property is different from the actual instance type.

p3.8xlarge (cheaper):
    - GPU Name: V100
    - GPU Count: 4

g5.48xlarge:
    - GPU Name: A10G
    - GPU Count: 8

The min GPU count is 4 and the GPU name should be `V100`.
This seems to be a result of using the first instance type in the list when determining the GPU Type.

One way could be to get the instance type with the min GPU count first then extract the type. 

```
def _get_min_gpu_instance_type(instance_types):
    return min(instance_types, _gpu_count)

def _gpu_count(instance_type):
    ...


... 
instance_types = [...]
min_gpu_instance_type = _get_min_gpu_instance_type(instance_types)
min_gpu_count = _gpu_count(instance_type):
gpu_type = _gpu_type(min_gpu_instance_type)
```

lukeseawalker: done

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -198,70 +198,108 @@ def _get_jinja_env(template_directory, realmemory_to_ec2memory_ratio):
...
-
-def _gpu_type(instance_type):
-    """Return name or type of the GPU for the instance."""
-    gpu_info = instance_types_data[instance_type].get("GpuInfo", None)
-    # Remove space and change to all lowercase for name
-    return "no_gpu_type" if not gpu_info else gpu_info.get("Gpus")[0].get("Name").replace(" ", "").lower()
+        return [compute_resource["InstanceType"]]
+
+
+def _get_min_gpu_count_and_type(instance_types) -> Tuple[int, str]:
```
EddyMM, line 212: For readability, could we return a dictionary instead?
```
{"count": gpu_count, "type": gpu_type}
```

I've noticed the configuration is using the `first` and `last` jinja filters which makes it hard to follow without looking at the function return values again.

lukeseawalker: done

EddyMM (approved):

PR #1530, opened 2022-09-19
Add support for Slurm Accounting

jdeamicis:
### Description of changes
* Basic configuration of Slurm Accounting at cookbook level with:
    * image-builder recipes to create slurmdbd user;
    * configuration of slurmdbd service;
    * configuration of Slurm config files to set up accounting (with unit tests);
    * basic idempotent bootstrapping of Slurm database;
    * updatability via `pcluster update-cluster`;
    * basic kitchen tests for slurmdbd user and group.

### Tests
* Ran unit tests with the patched code.
* Manual creation of cluster with Slurm accounting.
* Manual update of cluster of cluster to enable/disable the functionality.
* Integration tests to be added at a later stage.

### References
* https://github.com/aws/aws-parallelcluster/pull/4356

### Checklist
- [x] Make sure **to have added unit tests ~~or integration tests~~** to cover the new/modified code.


cookbooks/aws-parallelcluster-slurm/recipes/config_head_node.rb
```diff
@@ -86,6 +86,13 @@
   mode '0644'
 end
 
+template "#{node['cluster']['slurm']['install_dir']}/etc/slurmdbd.conf" do
+  source 'slurm/slurmdbd.conf.erb'
+  owner "#{node['cluster']['slurm']['dbduser']}"
+  group "#{node['cluster']['slurm']['dbdgroup']}"
+  mode '0600'
+end
+
```
davprat: [Question] If I'm reading this correctly, we are creating slurmdb configuration files, whether or not slurm accounting is enabled?

davprat (dismissed):

cookbooks/aws-parallelcluster-slurm/recipes/config_head_node.rb
jdeamicis: Yes, but only with generic parameters - with this configuration slurmdbd wouldn't be working. And we don't enable the slurmdbd service.

demartinofra: Can this be problematic in case the user is already enabling accounting by themselves and want to continue to do so?Wouldn't it be better to not produce any file in case the user does not select accounting settings so that we are less breaking with respect to existing solutions?

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -12,8 +12,8 @@
 provides :manage_ebs
 unified_mode true
 
-property :shared_dir_array, required: true
-property :vol_array, required: true
+property :shared_dir_array, Array, required: true
+property :vol_array, Array, required: true
```
demartinofra: why this change in this PR?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/slurm.conf.erb
```diff
@@ -9,10 +9,12 @@
 # See the slurm.conf man page for more information.
 #
 # CLUSTER SETTINGS
-ClusterName=parallelcluster
+ClusterName=<%= node['cluster']['stack_name'] %>
 SlurmUser=<%= node['cluster']['slurm']['user'] %>
 SlurmctldPort=6820-6829
 SlurmdPort=6818
+AccountingStoragePort=6819
+AccountingStorageUser=<%= node['cluster']['slurm']['dbduser'] %>
```
demartinofra: I would skip this part if accounting is not enabled to avoid conflicts with existing user customiazations

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -65,6 +68,8 @@ def generate_slurm_config_files(
     cluster_config = _load_cluster_config(input_file)
     head_node_config = _get_head_node_config()
     queues = cluster_config["Scheduling"]["SlurmQueues"]
+    tags_dict = {a["Key"]: a["Value"] for a in cluster_config["Tags"]}
+    cluster_name = tags_dict["parallelcluster:cluster-name"]
```
demartinofra: [nitpick] you could avoid the extra dict if you want `next(tag["Value"] for tag in tags if tag["Key"] == "parallelcluster:cluster-name")`

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -307,6 +319,82 @@ def _realmemory(compute_resource, realmemory_to_ec2memory_ratio) -> int:
...
+            ret = "3306"
+    return ret
+
+
+def _parse_uri(uri, attr) -> str:
+    """Get a host from a URI/URL using urlparse."""
+    # First, throw error if the URI starts with a "/" (to prevent issues with the
+    # manipulation below
+    if not _check_leading_slash(uri):
+        return None
```
demartinofra: This and the if below with _check_scheme are a bit confusing because of the following:
* by reading this function it gives the impression you are returning None in case the check fails. And returning None  would produce an invalid template.
* However what it actually happens is that the function raises an exception and the `return None` line is just dead code.

I'd suggest to make this more consistent/readable. Either have the checker function always return True/False and have the caller react on that or transform this in assertions and simply call them like `_check_leading_slash(uri)` without and if/return logic.

demartinofra: Also, more of a general question, can't we assume that this checks are done by the CLI that validates the config and do not need to be repeated here?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/slurmdbd.service.erb
```diff
@@ -0,0 +1,16 @@
...
+Wants=network-online.target
+ConditionPathExists=<%= node['cluster']['slurm']['install_dir'] %>/etc/slurmdbd.conf
+
+[Service]
+Type=simple
+EnvironmentFile=-/etc/sysconfig/slurmdbd
+ExecStart=<%= node['cluster']['slurm']['install_dir'] %>/sbin/slurmdbd -D -s $SLURMDBD_OPTIONS
+ExecReload=/bin/kill -HUP $MAINPID
+LimitNOFILE=65536
+TasksMax=infinity
```
demartinofra: I see this is not included in the service file generated at compile time. Can you leave a comment on why this is necessary for future reference?

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
```diff
@@ -0,0 +1,100 @@
...
+end
+
+execute "update Slurm database password" do
+  user 'root'
+  group 'root'
+  command "#{node['cluster']['scripts_dir']}/slurm/update_slurm_database_password.sh"
+end
+
+service "slurmdbd" do
+  supports restart: false
```
demartinofra, line 52: why this is false?

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
```diff
@@ -0,0 +1,100 @@
...
+  action %i(enable start)
+end
+
+# After starting slurmdbd the database may not be fully responsive yet and
+# its bootstrapping may fail. We need to wait for sacctmgr to successfully
+# query the database before proceeding.
+bash "wait for slurm database" do
```
demartinofra: 1. are there cases when this can take much longer than 30seconds? e.g. using a db that had been already used with an older Slurm version? is this something we need to account for?
2. you can look at how a similar check is don in head_node_finalize, the useful part about that approach is that it leaves a trace in the chef logs about what it's doing so that it's easier to debug issues

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
```diff
@@ -50,6 +50,25 @@
...
+end
+
+# Setup slurmdbd user
+user node['cluster']['slurm']['dbduser'] do
+  comment 'slurmdbd user'
+  uid node['cluster']['slurm']['dbduser_id']
+  gid node['cluster']['slurm']['dbdgroup_id']
+  # home is mounted from the head node
+  manage_home ['HeadNode', nil].include?(node['cluster']['node_type'])
+  home "/home/#{node['cluster']['slurm']['dbduser']}"
```
demartinofra: can we avoid creating the home directory for this user? So that we do no add new deps on the /home if not necessary

cookbooks/aws-parallelcluster-slurm/recipes/config_head_node.rb
jdeamicis: This happens only at cluster creation time. The customers are free to manipulate the base `slurmdbd.conf`and they are good to go.

Generating the whole `slurmdbd.conf` in a dynamic manner would be inconsistent with respect to how we generate the other Slurm configuration files, and would also mean that at cluster update time we would erase any customer customization for the slurmdbd that is not generated via `pcluster_slurm_config_generator.py`.

I'll think about this, but up to now all the alternatives that come to my mind are either worse, or require that we re-think the way users can customize the Slurm configuration.

attributes/default.rb
```diff
@@ -131,9 +131,17 @@
 default['cluster']['slurm']['user_id'] = node['cluster']['reserved_base_uid'] + 1
 default['cluster']['slurm']['group'] = node['cluster']['slurm']['user']
 default['cluster']['slurm']['group_id'] = node['cluster']['slurm']['user_id']
+default['cluster']['slurm']['dbduser'] = 'slurmdbd'
+default['cluster']['slurm']['dbduser_id'] = node['cluster']['reserved_base_uid'] + 5
```
davprat, line 138: [Question] Would it make sense to have the uid offsets centralized somewhere to ensure we don't accidentally create a collision or is this not something to worry about?

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
jdeamicis: Nothing related to this PR. It's a minor improvement (the type of the property wasn't declared and I was getting warnings). Should I move this to a separate commit out of this PR?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/slurm.conf.erb
jdeamicis: Let's consider that if parameters are redefined at the end of our configuration files, our values get overridden. Still I guess I can move this to the include file.

jdeamicis: Done, although it required adding an argument to the `pcluster_slurm_config_generator.py` script, which is not ideal.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: About your first comment, I had implemented the same structure from the CLI side and left the same `if not ... return` structure, which here I probably didn't need. Thanks.

For your second comment, we have a "backdoor" when users decide to disable the validation on the CLI side. In this case should we allow the cookbook to blindly accept whatever the users provide as input?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/slurmdbd.service.erb
jdeamicis: Do you mean the `TasksMax` parameter? Or the whole block?

If you are talking about `TasksMax`, this is actually coming from [this commit](https://github.com/SchedMD/slurm/commit/02aa5a1a20cd97a238801e99ecb7af1df0022e39) in Slurm. It looks like it is being added automatically if the installer (Makefile) sees a version of systemd >= 227. Actually this is done for all the Slurm services.

I'll remove this and add it for all services once we check that we have a suitable version of systemd on all our AMIs.

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
jdeamicis: I copied what is being done for the slurmctld in the `config_head_node.rb` recipe, and it's been here since the beginning.

I assume it's to trick Chef into full stopping and then starting the service without attempting a restart. I'm not sure why we would need this, but I guess here it doesn't make too much of a difference since it's the first time we start the service (and maybe that's why we have `false` in the first place?).

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
jdeamicis: I had originally put a sleep of 10 seconds, and sometimes it worked, sometimes it didn't. So I put a more robust wait loop, assuming 30 seconds would be enough.

> e.g. using a db that had been already used with an older Slurm version? is this something we need to account for? 

Good catch. Indeed from the [Slurm documentation](https://slurm.schedmd.com/quickstart_admin.html#upgrade):

> Database table changes may be required for the upgrade, for example adding new fields to existing tables. If the database contains a large number of entries, the SlurmDBD daemon may require tens of minutes to update the database and be unresponsive during this time interval.

Should we consider this scenario here? For such an operation we should probably recommend to use a non-production `slurmdbd` to update the database (similar to the scenario of having a backup of old data in a non-production `slurmdbd`). Otherwise we should install a mysql client, try to connect to the database to see if it exists, and if so increment the wait time? 

In any case I can increase the total wait time to something like 5 minutes (30 retries with 10 sleep time) for now.

> you can look at how a similar check is don in head_node_finalize

Yes, that's not bad. I'll implement it this way.

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
jdeamicis: I would leave this to a separate PR where we remove the homes for system users (`slurm` has a home directory too).

attributes/default.rb
jdeamicis: Absolutely legit question. It looks like the centralized place is the `default.rb` file itself: the offsets are just scattered around the file because they are defined where they are used.

I guess it makes more sense to keep them as they are because this way the users are defined in parts of the file that contain all the details related to why that user has been defined. And we can always just grep `reserved_base_uid` if we need all the offsets.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/slurmdbd.service.erb
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
jdeamicis: Reimplemented the wait with a Chef `execute` resource.

cookbooks/aws-parallelcluster-slurm/recipes/config_head_node.rb
jdeamicis: I discussed with @demartinofra and the idea is:
- at create time, write `slurmdbd.conf` only if our accounting is enabled;
- at update time when feature gets enabled, overwrite any possible slurmdbd config files.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: OK to remove it from the cookbook side. To be done.

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
jdeamicis: Fdm and I decided to remove the timeout for the time being and rely on the CFN default timeout.

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
jdeamicis: We will remove the home directory for `slurmdbd`.

attributes/default.rb
jdeamicis: We wiil add it to the extra tasks for this feature.

cookbooks/aws-parallelcluster-slurm/recipes/config_head_node.rb
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/recipes/config_slurm_accounting.rb
jdeamicis: The most straightforward way to implement this with the `execute` Chef resource is to set a very large number for the `retries parameter. Setting 100000 retries of 10 seconds each.

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
jdeamicis: Done.

davprat (dismissed):

davprat (approved):

PR #1554, opened 2022-10-07
Change default NFS thread number to have better performances

enrico-usai:
### Description of changes
I tested to launch a job with 1000 nodes on a c5.12xlarge (24 NFS threads) and
I noticed the `nfsd: too many open connections, consider increasing the number of threads` message.

We should use at the very least one daemon per processor, but four to eight per processor is considered a better rule of thumb. We're capping the value to 256 to avoid having too many threads.

With this patch I'm also exposing the `node['cluster']['nfs']['threads']` parameter in the `default.rb`
file to permit to overwrite it from `DevSettings` if needed.

### Tests
* Executed build of AMIs with modified recipes
* Executed the `test_scaling.py::test_multiple_jobs_submission` integ tests for both x86 and ARM architecture.

### References
* According to https://nfs.sourceforge.net/nfs-howto/ar01s05.html in the early days of NFS, Sun decided on 8 as a rule of thumb, and everyone else copied. We should use at the very least one daemon per processor, but 4 to 8 per processor may be a better rule of thumb.
* According to Suse support, boosting the number of threads (even dramatically) is not likely to cause any negative side effects. Therefore, it is safe to be generous with this setting. https://www.suse.com/support/kb/doc/?id=000017927

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/nfs.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/nfs.rb
@@ -4,7 +4,7 @@
 # Cookbook:: aws-parallelcluster
 # Recipe:: nfs
 #
-# Copyright:: 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright:: 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
 # License. A copy of the License is located at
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-config/recipes/nfs.rb
```diff
@@ -16,11 +16,7 @@
 # limitations under the License.
 
 # For performance, set NFS threads to max(num_cores, 8)
```
lukeseawalker: align comment with the new value logic

enrico-usai: Comment removed, it is in the default.rb now. Thanks!

lukeseawalker (approved):

PR #1556, opened 2022-10-12
Handle filesystems rollback if case of cluster update failure

chenwany:
### Description of changes
* Make shared storages resources idempotent to make it handle the external filesystem rollback.
* Backup shared storage mapping file during update in case we need to use it during rollback.
### Tests
* Manually test with Update cluster with problematic EBS, EFS and FSX

### References

### Checklist

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -36,7 +36,7 @@ default['cluster']['launch_templates_config_path'] = "#{node['cluster']['shared_
 default['cluster']['instance_types_data_path'] = "#{node['cluster']['shared_dir']}/instance-types-data.json"
 default['cluster']['computefleet_status_path'] = "#{node['cluster']['shared_dir']}/computefleet-status.json"
 default['cluster']['shared_storages_mapping_path'] = "/etc/parallelcluster/shared_storages_data.yaml"
-default['cluster']['update_shared_storages_mapping_path'] = "/etc/parallelcluster/update_shared_storages_data.yaml"
+default['cluster']['previous_shared_storages_mapping_path'] = "/etc/parallelcluster/previous_shared_storages_data.yaml"
 
 default['cluster']['reserved_base_uid'] = 400
 
--- a/cookbooks/aws-parallelcluster-config/recipes/update.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/update.rb
@@ -21,8 +21,14 @@ unless node['cluster']['scheduler'] == 'awsbatch'
   end
 end
 
+ruby_block "backup storages mappings" do # backup shared storages mapping file to used for update and rollback
+  block do
+    ::FileUtils.cp_r(node['cluster']['shared_storages_mapping_path'], node['cluster']['previous_shared_storages_mapping_path'], remove_destination: true)
+  end
+end
+
 # generate the update shared storages mapping file
-template node['cluster']['update_shared_storages_mapping_path'] do
+template node['cluster']['shared_storages_mapping_path'] do
   source 'shared_storages/shared_storages_data.erb'
   mode '0644'
 end
--- a/cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
+++ b/cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
@@ -89,9 +89,22 @@ action :mount do
       device_type :uuid
       options "_netdev"
       pass 0
-      action %i(mount enable)
+      action :mount
       retries 10
       retry_delay 6
+      not_if "mount | grep ' #{shared_dir_array[index]} '"
+    end
+
+    mount shared_dir_array[index] do
+      device(DelayedEvaluator.new { dev_uuids[index] })
+      fstype(DelayedEvaluator.new { node['cluster']['volume_fs_type'] })
+      device_type :uuid
+      options "_netdev"
+      pass 0
+      action :enable
+      retries 10
+      retry_delay 6
+      only_if "mount | grep ' #{shared_dir_array[index]} '"
     end
 
     # Make sure shared directory permissions are correct
@@ -131,22 +144,23 @@ action :unmount do
 
   # Mount each volume
   dev_path = [] # device labels
-  dev_uuids = [] # device uuids
 
   vol_array.each_with_index do |volumeid, index|
     dev_path[index] = "/dev/disk/by-ebs-volumeid/#{volumeid}"
-    dev_uuids[index] = get_uuid_for_unmount(shared_dir_array[index])
 
     # Unmount and remove volume from /etc/fstab
-    mount shared_dir_array[index] do
-      device(DelayedEvaluator.new { dev_uuids[index] })
-      fstype(DelayedEvaluator.new { node['cluster']['volume_fs_type'] })
-      device_type :uuid
-      options "_netdev"
-      pass 0
-      action %i(unmount disable)
+    execute 'unmount ebs' do
+      command "umount -fl #{shared_dir_array[index]}"
       retries 10
       retry_delay 6
+      timeout 60
+      only_if "mount | grep ' #{shared_dir_array[index]} '"
+    end
+
+    # remove volume from fstab
+    delete_lines "remove volume from /etc/fstab" do
+      path "/etc/fstab"
+      pattern " #{shared_dir_array[index]} "
     end
 
     # Detach EBS volume
--- a/cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
+++ b/cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
@@ -67,9 +67,22 @@ action :mount do
         dump 0
         pass 0
         options mount_options
-        action %i(mount enable)
+        action :mount
         retries 10
         retry_delay 6
+        not_if "mount | grep ' #{fsx_shared_dir} '"
+      end
+
+      mount fsx_shared_dir do
+        device "#{dns_name}@tcp:/#{mount_name}"
+        fstype 'lustre'
+        dump 0
+        pass 0
+        options mount_options
+        action :enable
+        retries 10
+        retry_delay 6
+        only_if "mount | grep ' #{fsx_shared_dir} '"
       end
     when 'OPENZFS'
       mount fsx_shared_dir do
@@ -78,9 +91,21 @@ action :mount do
         dump 0
         pass 0
         options 'nfsvers=4.2'
-        action %i(mount enable)
+        action :mount
         retries 10
         retry_delay 6
+        not_if "mount | grep ' #{fsx_shared_dir} '"
+      end
+      mount fsx_shared_dir do
+        device "#{dns_name}:#{fsx_volume_junction_path}"
+        fstype 'nfs'
+        dump 0
+        pass 0
+        options 'nfsvers=4.2'
+        action :enable
+        retries 10
+        retry_delay 6
+        only_if "mount | grep ' #{fsx_shared_dir} '"
       end
     when 'ONTAP'
       mount fsx_shared_dir do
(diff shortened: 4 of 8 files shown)


gmarciani:
May you please add a description of the change and the way you tested it?
Thanks a lot.

attributes/default.rb
```diff
@@ -36,7 +36,7 @@
 default['cluster']['instance_types_data_path'] = "#{node['cluster']['shared_dir']}/instance-types-data.json"
 default['cluster']['computefleet_status_path'] = "#{node['cluster']['shared_dir']}/computefleet-status.json"
 default['cluster']['shared_storages_mapping_path'] = "/etc/parallelcluster/shared_storages_data.yaml"
-default['cluster']['update_shared_storages_mapping_path'] = "/etc/parallelcluster/update_shared_storages_data.yaml"
+default['cluster']['prvious_shared_storages_mapping_path'] = "/etc/parallelcluster/previous_shared_storages_data.yaml"
```
gmarciani: Typo: prvious => previous

In other places within this PR, we reference with the correct key (not typo), so it's not clear to me how did it work.

Furthermore, I suggest to name this file as `pre_update_shared_storage_mapping` rather than `previous_shared_storage_mapping` as it would be more self-explanatory.

cookbooks/aws-parallelcluster-config/recipes/update.rb
```diff
@@ -21,8 +21,14 @@
   end
 end
 
+ruby_block "backup storages mappings" do
```
gmarciani: I'd say explicitly in a comment on in the resource description that we are backing up these settings becase they need to be used for a potential rollback

cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
```diff
@@ -133,6 +133,7 @@
       command "umount #{fsx_shared_dir}"
       retries 10
       retry_delay 6
+      only_if "mountpoint -q #{fsx_shared_dir}"
```
gmarciani: The role of this `only_if` is to make the resource idempotent.
Afaik we do not have this in `manage_efs` and `manage_ebs` resource.
Why should we put it only in FSx?

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -90,8 +90,9 @@
       options "_netdev"
       pass 0
       action %i(mount enable)
-      retries 10
+      retries 1
```
gmarciani: Here and elsewhere.
Why should we decrease the number of retries?

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -90,8 +90,9 @@
...
       retry_delay 6
+      not_if "mountpoint -q #{shared_dir_array[index]}"
```
gmarciani: Minor (let's consider it outside of this PR): we reference `shared_dir_array[index]` in many places; I think we should use a more explicit variable like `mount_point = shared_dir_array[index]`

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -135,18 +136,20 @@
...
-      options "_netdev"
-      pass 0
-      action %i(unmount disable)
-      retries 10
+    execute 'unmount ebs' do
+      command "umount -l #{shared_dir_array[index]}"
+      retries 1
       retry_delay 6
+      timeout 30
+      only_if "mountpoint -q #{shared_dir_array[index]}"
```
gmarciani: Here and elsewhere.
The `only_if` attribute expects the command to return True or False.
This is not the case for the command `mountpoint`.

cookbooks/aws-parallelcluster-config/recipes/update.rb
chenwany: Added ac comment, it is also used for update

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
chenwany: I decrease the number for faster testing the change, the PR was not ready for review at that time and it was a draft change

cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
chenwany: Added also for other resources

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
chenwany: the only_if and not_if command works also for linux command

attributes/default.rb
gmarciani: Thanks for fixing the typo

cookbooks/aws-parallelcluster-config/recipes/update.rb
```diff
@@ -21,8 +21,14 @@
   end
 end
 
+ruby_block "backup storages mappings" do # backup shared storages mapping file to used for update and rollback
+  block do
+    ::FileUtils.cp_r(node['cluster']['shared_storages_mapping_path'], node['cluster']['previous_shared_storages_mapping_path'], remove_destination: true)
```
gmarciani, line 26: Minor: I think `previous_*` does not explicitly clarify the goal of that file.
If the goal is to store the configuration before the actual execution of an update/rollback, what about naming it `pre_update_*`

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -131,22 +144,23 @@
...
 
     # Unmount and remove volume from /etc/fstab
-    mount shared_dir_array[index] do
-      device(DelayedEvaluator.new { dev_uuids[index] })
-      fstype(DelayedEvaluator.new { node['cluster']['volume_fs_type'] })
-      device_type :uuid
-      options "_netdev"
-      pass 0
-      action %i(unmount disable)
+    execute 'unmount ebs' do
```
gmarciani, line 152: Here for EBs and everywhere else for EFS/FSx: what about putting the unmounting steps before the mounting steps?
I say so because:
1. lazy unmounting actions operations may have better chances to succeed w.r.t. mounting operations
2. since mounting operations may require external resources to be ready to be mounted (FS, networking,...) better to give some extra time to complete the setup if we can

cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
```diff
@@ -78,19 +91,43 @@
         dump 0
         pass 0
         options 'nfsvers=4.2'
-        action %i(mount enable)
```
gmarciani, line 81: We are replicating the mount+enable steps for every file system type. What about avoiding this code duplication by putting the mount+enable logic within a function?

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -131,22 +144,23 @@
...
+      command "umount -fl #{shared_dir_array[index]}"
```
gmarciani, line 153: The following are shared among EBS/EFS/FSx*:
1. the forced-lazy unmounting command;
2. the `not_if` condition on mount
3. the `only_if` condition on enable action
4. retries, retry_delay, timeout

What about putting this command in common?

cookbooks/aws-parallelcluster-config/recipes/update.rb
chenwany: For cluster config, we name it previous-cluster-config.yaml, I think it is better to be aligned with that

cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
chenwany: They are not completely duplicate, Different filesystems are mounted with different mounting options. The code was like that before we support DFSM. Can we handle it outside this PR?

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
chenwany: We may want to have different retry_delay here as we discussed for the EFS case, and different execution resources have different name. But I agree with put them in the function will make them looks more clean

cookbooks/aws-parallelcluster-config/recipes/update.rb
gmarciani: Got it, let's keep them aligned.

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
gmarciani: Ok, I'll keep track of the action item for clean coding

cookbooks/aws-parallelcluster-config/resources/manage_fsx.rb
gmarciani: Sure, let' address this in another Pr. I'll keep track of this

gmarciani (approved):

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
gmarciani: since this PR is high priority, let's address all the clean coding comments in another PR. I'll take care of it

cookbooks/aws-parallelcluster-config/recipes/update.rb
```diff
@@ -21,8 +21,14 @@
   end
 end
 
+ruby_block "backup storages mappings" do # backup shared storages mapping file to used for update and rollback
+  block do
```
lukeseawalker, line 25: please add a log here, similar to the one we have when backing up cluster configuration

cookbooks/aws-parallelcluster-config/recipes/update.rb
lukeseawalker, line 26: I would move this under the fetch_config with update true, few lines above

cookbooks/aws-parallelcluster-config/resources/manage_ebs.rb
```diff
@@ -89,9 +89,22 @@
...
+    mount shared_dir_array[index] do
+      device(DelayedEvaluator.new { dev_uuids[index] })
+      fstype(DelayedEvaluator.new { node['cluster']['volume_fs_type'] })
+      device_type :uuid
+      options "_netdev"
+      pass 0
+      action :enable
+      retries 10
+      retry_delay 6
+      only_if "mount | grep ' #{shared_dir_array[index]} '"
```
lukeseawalker, line 107: shouldn't this check if FS is enabled or not? This because the mount is done in the previous action, so here the mount check is passing while even if the FS is not enabled.

~~Furthermore, can the grep on mount be improved, e.g. directly checking the mount point and not the whole mount output?~~

chenwany: In the chef code, the enable action already check if it is enabled before enable it. So I was checking if it is mounted before enable it. 
https://github.com/chef/chef/blob/e85fcb8c0ad1ea4ed6f830e4b585f27c2ada4994/lib/chef/provider/mount/mount.rb#L251
For checking mountpoint, if the FS dones't have correct SG, the mountpoint command will hang. So I switch to mount | grep
```
        def enabled?
          # Check to see if there is a entry in /etc/fstab. Last entry for a volume wins.
          enabled = false
          unless ::File.exist?("/etc/fstab")
            logger.debug "/etc/fstab not found, treating mount as not-enabled"
            return
          end
          ::File.foreach("/etc/fstab") do |line|
            case line
            when /^[#\s]/
              next
            when /^(#{device_fstab_regex})\s+#{Regexp.escape(@new_resource.mount_point)}\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)/
              enabled = true
              @current_resource.device($1)
              @current_resource.fstype($2)
              @current_resource.options($3)
              @current_resource.dump($4.to_i)
              @current_resource.pass($5.to_i)
              logger.trace("Found mount #{device_fstab} to #{@new_resource.mount_point} in /etc/fstab")
            end
          end
          @current_resource.enabled(enabled)
        end
```

cookbooks/aws-parallelcluster-config/recipes/update.rb
chenwany: Good idea! Address here: https://github.com/aws/aws-parallelcluster-cookbook/pull/1563

cookbooks/aws-parallelcluster-config/recipes/update.rb
chenwany: Good idea! Added here: https://github.com/aws/aws-parallelcluster-cookbook/pull/1563

PR #1570, opened 2022-10-19
3.x Install MySQL packages from public ParalllelCluster S3 buckets.

davprat:
### Description of changes
* This change is to install MySQL packages needed for Slurm accounting from public S3 buckets controlled by the ParallelCluster team rather than from the public MySQL repositories.
* This change cannot be merged until the MySQL packages are published to the. public S3 buckets.

### Tests
* Built and validated AMI's using commercial-dev public bucket.
* Ran Slurm accounting integration tests against built AMIs.
* Checks are not passing due to MySQL packages not being available on the public S3 buckets yet.

### Checklist


attributes/default.rb
```diff
@@ -238,26 +241,55 @@
   'default' => "659400f9842fffb8d64ae0b650f081b9",
   'ubuntu' => { 'default' => "c2b410031867dc7c966ca5b1aa0c72aa" }
 )
+
+# MySQL Packages
+default['cluster']['mysql']['package']['root'] = "https://#{node['cluster']['region']}-aws-parallelcluster.s3.#{node['cluster']['region']}.#{node['cluster']['aws_domain']}/archives/mysql"
```
gmarciani: What about having the base URL for artifacts in a dedicated variable, so that it is ready to be used for other artifacts?

```
default['cluster']['artifacts_s3_url'] = ""https://#{node['cluster']['region']}-aws-parallelcluster.s3.#{node['cluster']['region']}.#{node['cluster']['aws_domain']}/archives"

...

# My SQL Packages
default['cluster']['mysql']['package']['root'] = "#{node['cluster']['artifacts_s3_url']/mysql}"
```

attributes/default.rb
```diff
@@ -238,26 +241,55 @@
...
+default['cluster']['mysql']['package']['archive'] = "#{node['cluster']['mysql']['package']['root']}/#{node['cluster']['mysql']['package']['platform']}/#{node['cluster']['mysql']['package']['file-name']}"
+default['cluster']['mysql']['package']['source'] = "#{node['cluster']['mysql']['package']['root']}/source/mysql-#{node['cluster']['mysql']['package']['source-version']}.tar.gz"
+
+# MySQL Validation
 if arm_instance?
   default['cluster']['mysql']['repository']['packages'] = value_for_platform(
-    'default' => %w(mysql-community-devel mysql-community-libs mysql-community-common mysql-community-client-plugins),
+    'default' => %w(mysql-community-devel mysql-community-libs mysql-community-common mysql-community-client-plugins mysql-community-libs-compat),
     'ubuntu' => {
-      'default' => %w(libmysqlclient-dev libmysqlclient21 mysql-common),
```
gmarciani, line 245: why are we removing `mysql-common`?

attributes/default.rb
```diff
@@ -238,26 +241,55 @@
...
-      '18.04' =>  %w(libmysqlclient-dev libmysqlclient20 mysql-common),
+      'default' => %w(libmysqlclient-dev libmysqlclient21),
+      '18.04' =>  %w(libmysqlclient-dev libmysqlclient20),
     }
   )
-  default['cluster']['mysql']['repository']['expected']['source'] = value_for_platform(
-    'default' => "mysql80-community",
-    'ubuntu' => { 'default' => "http://#{node['cluster']['region']}.ec2.ports.ubuntu.com/ubuntu-ports" }
+  default['cluster']['mysql']['repository']['expected']['version'] = value_for_platform(
```
gmarciani, line 279: [Minor] Why `expected`? it this an attribute used for testing?
What about 

```
default['cluster']['mysql']['repository']['version']
```

cookbooks/aws-parallelcluster-install/templates/default/mysql/mysql_source_code.erb
```diff
@@ -0,0 +1,3 @@
+You can get MySQL source code here:
```
gmarciani: I think it's a good practice to include the extension of the final file within the erb file name because it makes it easier to imagine the expected result.
I mean `mysql_source_code.xyz.erb` => `mysql_source_code.txt.erb`

What do you think?

What do you think?

cookbooks/aws-parallelcluster-install/recipes/install_mysql_client.rb
```diff
@@ -15,55 +15,48 @@
...
   end
+else
 
-  ruby_block "Validate Repository Definition Checksum" do
-    block do
-      require 'digest'
-      checksum = Digest::MD5.file(local_file).hexdigest
+  package_installer = value_for_platform(
+    'default' => "yum install -y",
+    'ubuntu' => { 'default' => "apt install" }
```
gmarciani: Does `apt` require an equivalent `-y` flag?

cookbooks/aws-parallelcluster-install/recipes/install_mysql_repository.rb
```diff
@@ -0,0 +1,64 @@
...
+def install_repository_configuration_package(repository_name, definition_source, local_file, md5_hash)
+  remote_file local_file do
+    source definition_source
+    mode '0644'
+    retries 3
+    retry_delay 5
+    not_if { ::File.exist?(local_file) }
+  end
+
+  ruby_block "Validate Repository Definition Checksum" do
```
gmarciani: [Minor] This seems to be useful in general. Let's consider to move this block into a function that can be used by other recipes.

cookbooks/aws-parallelcluster-install/recipes/install_mysql_repository.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+# Recipe:: mysql
```
gmarciani: This recipe name is `install_mysql_repository` not `mysql`

attributes/default.rb
```diff
@@ -238,26 +241,55 @@
...
+    'default' => "8.0.31",
```
gmarciani, line 280: this is equal to `default['cluster']['mysql']['package']['source-version']`. Why do we need to redefine the version?

attributes/default.rb
davprat: Because, for Ubuntu (18|20).04 Arm, the version of mysql-common is not aligned with the versions of the other two packages (For Ubuntu Arm, the packages don't come from MySQL, but from an ec2 repo). What would be a better way to account for the version differences in this case?

attributes/default.rb
davprat: The source version doesn't include the release, but I could redefine the package versions to be ['version'] + ['release'], what do you think?

cookbooks/aws-parallelcluster-install/recipes/install_mysql_client.rb
davprat: I'll add it.

attributes/default.rb
gmarciani: So now we are taking `mysql-common` package from our artifacts on S3?
I do not have the whole context on the activity, so I may be wrong: what looks a bit fragile to me is taking packages that require to be aligned part from repos and part from our S3 bucket. If this is the case, isn't this approach exposed to packages misalignment in the future?

attributes/default.rb
gmarciani: Thanks for the explanation. This is a minor, we can also keep the current approach.

attributes/default.rb
```diff
@@ -238,26 +245,55 @@
...
+                                                           '20.04' => "ubuntu/20.04/x86_64",
+                                                           '18.04' => "ubuntu/18.04/x86_64",
+                                                         }
+                                                       )
+                                                     end
+default['cluster']['mysql']['package']['file-name'] = "mysql-community-client-#{node['cluster']['mysql']['package']['version']}.tar.gz"
+default['cluster']['mysql']['package']['archive'] = "#{node['cluster']['mysql']['package']['root']}/#{node['cluster']['mysql']['package']['platform']}/#{node['cluster']['mysql']['package']['file-name']}"
+default['cluster']['mysql']['package']['source'] = "#{node['cluster']['artifacts_s3_url']}/source/mysql-#{node['cluster']['mysql']['package']['source-version']}.tar.gz"
+
+# MySQL Validation
```
gmarciani, line 270: [minor] If I understood correctly, with "MySQL Validation" you mean the expectations during kitchen tests.
If this the case, please specify that these are not installed but used in Kitchen tests.

attributes/default.rb
jdeamicis: I would consider using a variable for this (and subsequent) version checks. This will definitely save us a build failure in the future when we forget to propagate the version bump in all the lines of this file :)

attributes/default.rb
```diff
@@ -238,26 +245,55 @@
   'default' => "659400f9842fffb8d64ae0b650f081b9",
   'ubuntu' => { 'default' => "c2b410031867dc7c966ca5b1aa0c72aa" }
 )
+
+# MySQL Packages
```
jdeamicis, line 249: @davprat please let's put a comment saying that we install our pinned version of the MySQL packages in every combination of OS/architecture except for Ubuntu ARM (where we rely in the OS repositories).

attributes/default.rb
gmarciani: Point clarified offline by @jdeamicis.

gmarciani (approved):
LGTM, please keep track of the non blocking minor comments to address them in a separated PR.

cookbooks/aws-parallelcluster-install/recipes/install_mysql_client.rb
```diff
@@ -15,55 +15,48 @@
...
-    dpkg_package repository_name do
-      source local_file
-    end
+  remote_file mysql_tar_file do
+    source mysql_archive_url
+    mode '0644'
+    retries 3
+    retry_delay 5
+    not_if { ::File.exist?(mysql_tar_file) }
+  end
```
jdeamicis, line 39: [minor] Could we please put a ruby expression to log the `mysql_archive_url` in the Cinc logs? At the moment this is not logged by the `remote_file` resource.

PR #1574, opened 2022-10-25
Remove validation of MySQL packages on Ubuntu 18.04/20.04 ARM as they are installed from OS repos.

gmarciani:
### Description of changes
1. Remove validation of MySQL packages on Ubuntu 18.04/20.04 ARM as they are installed from OS repos.
2. Fix logic to validate package versions so that it will consider only installed packages.
3. Add log line to know the URL the MYSQL package is downloaded from.

### Tests
1. Custom build on Ubuntu 18.04 ARM with validation enabled => succeeded

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -277,11 +277,8 @@ if arm_instance?
     }
   )
   default['cluster']['mysql']['repository']['expected']['version'] = value_for_platform(
-    'default' => "8.0.31",
-    'ubuntu' => {
-      '20.04' => "8.0.30-0ubuntu0.20.04.2",
-      '18.04' => "5.7.39-0ubuntu0.18.04.2",
-    }
+    'default' => "8.0.31"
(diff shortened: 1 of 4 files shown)


attributes/default.rb
```diff
@@ -278,9 +278,10 @@
   )
   default['cluster']['mysql']['repository']['expected']['version'] = value_for_platform(
     'default' => "8.0.31",
```
davprat: Does this one need to be escaped as well?

davprat (dismissed):

attributes/default.rb
gmarciani: Yes, you right.
It's not worth it to convert every attribute in escaped regex. 
So, for the time being let's remove this check for Ubuntu Arm.
Then will will revised the validation logic.

jdeamicis (approved):

PR #1595, opened 2022-11-15
[surface head node bootstrap errors] store error messages from custom scripts

delongmeng:
### Description of changes
During the bootstrap of the head node, if any of custom scripts (OnNodeStart or OnNodeConfigured) fails, a generic error message will be sent to CloudFormation stack without providing specific reasons of the failure. Here we write error messages into a dedicated file, which will be extracted and sent to CloudFormation stack.

Also updated the error file name from headnode_bootstrap_error_msg to bootstrap_error_msg, since this file will be used in both the head node and compute nodes.

### Tests
Manually tested by creating clusters and checking the error messages in the CloudFormation stack:  
* testing error in post-install script (OnNodeConfigured)  => expected error message in CloudFormation:
  - *'Failed to run postinstall, s3:/<xxx/.../xxx>.sh failed with non 0 return code: 1. If —rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details.'*
* testing error in pre-install script (OnNodeStart)  => expected error message in CloudFormation:
  - *'Failed to run preinstall, s3://<xxx/.../xxx>.sh failed with non 0 return code: 1. If —rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details.'*
* testing no access to s3  => expected error message in CloudFormation:
  - *'Failed to run postinstall, failed to download s3://<xxx/.../xxx>.sh using aws s3, return code: 1. If —rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details.'*
* testing wrong http url of the custom script  => expected error message in CloudFormation:
  - *'Failed to run postinstall, failed to download https://github.com/<xxx/.../xxx>.sh using wget, return code: 8. If —rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details.'*

### References
* [This PR](https://github.com/aws/aws-parallelcluster/pull/4540) shows how the error messages file is used.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
+++ b/cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
@@ -21,6 +21,7 @@ function error_exit () {
   script=`basename $0`
   echo "parallelcluster: ${script} - $1"
   logger -t parallelcluster "${script} - $1"
+  echo "${script} - $1" > /var/log/parallelcluster/bootstrap_error_msg
   exit 1
 }
 
@@ -31,12 +32,12 @@ function download_run (){
   tmpfile=$(mktemp)
   trap "/bin/rm -f $tmpfile" RETURN
   if [ "${scheme}" == "s3" ]; then
-    <%= node['cluster']['cookbook_virtualenv_path'] %>/bin/aws --region ${cfn_region} s3 cp ${url} - > $tmpfile || return 1
+    <%= node['cluster']['cookbook_virtualenv_path'] %>/bin/aws --region ${cfn_region} s3 cp ${url} - > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using aws s3, return code: $?"
   else
-    wget -qO- ${url} > $tmpfile || return 1
+    wget -qO- ${url} > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using wget, return code: $?"
   fi
-  chmod +x $tmpfile || return 1
-  $tmpfile "$@" || error_exit "Failed to run ${ACTION}, ${file} failed with non 0 return code: $?"
+  chmod +x $tmpfile || error_exit "Failed to run ${ACTION}, failed to run chmod +x ${tmpfile}, return code: $?"
+  $tmpfile "$@" || error_exit "Failed to run ${ACTION}, failed to execute file ${file}, return code: $?"
 }
 
 function get_stack_status () {
--- a/cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
+++ b/cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
@@ -21,7 +21,7 @@ function error_exit () {
   script=`basename $0`
   echo "parallelcluster: ${script} - $1"
   logger -t parallelcluster "${script} - $1"
-  echo "${script} - $1" > /var/log/parallelcluster/bootstrap_error_msg
+  echo "$1. If --rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details." > /var/log/parallelcluster/bootstrap_error_msg
   exit 1
 }
 
@@ -37,7 +37,7 @@ function download_run (){
     wget -qO- ${url} > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using wget, return code: $?"
   fi
   chmod +x $tmpfile || error_exit "Failed to run ${ACTION}, failed to run chmod +x ${tmpfile}, return code: $?"
-  $tmpfile "$@" || error_exit "Failed to run ${ACTION}, failed to execute file ${file}, return code: $?"
+  $tmpfile "$@" || error_exit "Failed to run ${ACTION}, ${file} failed with non 0 return code: $?"
 }
 
 function get_stack_status () {
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -543,7 +543,7 @@ def raise_and_write_chef_error(raise_message, chef_error = nil)
   unless chef_error
     chef_error = raise_message
   end
-  Mixlib::ShellOut.new("echo '#{chef_error}' > /var/log/parallelcluster/headnode_bootstrap_error_msg").run_command
+  Mixlib::ShellOut.new("echo '#{chef_error}' > /var/log/parallelcluster/bootstrap_error_msg").run_command
   raise raise_message
 end
 


cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -15,6 +15,7 @@ function error_exit () {
   script=`basename $0`
   echo "parallelcluster: ${script} - $1"
   logger -t parallelcluster "${script} - $1"
+  echo "${script} - $1" > /var/log/parallelcluster/headnode_bootstrap_error_msg
```
chenwany: Can we add here aslo ask users to check the corresponding logs for details? eg:"Additional info can be found in /var/log/.....log"?

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -25,12 +26,12 @@ function download_run (){
...
-      <%= node['cluster']['cookbook_virtualenv_path'] %>/bin/aws --region ${cfn_region} s3 cp ${url} - > $tmpfile || return 1
+      <%= node['cluster']['cookbook_virtualenv_path'] %>/bin/aws --region ${cfn_region} s3 cp ${url} - > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using aws s3, return code: $?"
     else
-      wget -qO- ${url} > $tmpfile || return 1
+      wget -qO- ${url} > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using wget, return code: $?"
     fi
-    chmod +x $tmpfile || return 1
-    $tmpfile "$@" || error_exit "Failed to run ${ACTION}, ${file} failed with non 0 return code: $?"
+    chmod +x $tmpfile || error_exit "Failed to run ${ACTION}, failed to run chmod +x on temporary file, return code: $?"
+    $tmpfile "$@" || error_exit "Failed to run ${ACTION}, failed to execute script ${file}, return code: $?"
```
chenwany: why do we change the message here? `${file} failed with non 0 return code: $?"`

chenwany (commented):
Can you describe what are the manual tests done in PR description?

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -25,12 +26,12 @@ function download_run (){
     tmpfile=$(mktemp)
     trap "/bin/rm -f $tmpfile" RETURN
     if [ "${scheme}" == "s3" ]; then
-      <%= node['cluster']['cookbook_virtualenv_path'] %>/bin/aws --region ${cfn_region} s3 cp ${url} - > $tmpfile || return 1
+      <%= node['cluster']['cookbook_virtualenv_path'] %>/bin/aws --region ${cfn_region} s3 cp ${url} - > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using aws s3, return code: $?"
```
chenwany: Instead of having error_exit everywhere, can we rely on the `error_exit "Failed to run preinstall"` in function `run_preinstall` and `run_postinstall`. To clarify the error is from which step, we can define a variable to store the message before return 1

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Yes we can.

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Just trying to keep the format consistent with the other branches: "failed to download xxx", "failed to run xxx", "failed to execute xxx", etc. Also because the name of file (the xxx part) is usually pretty long, so "failed to execute xxx" is slightly more readable than "xxx failed". Does that make sense?

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Do you mean utilizing line 44 and line 56 of [this script](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb#L44), instead of line 33? (Currently it seems that only line 33 is executed)

enrico-usai:
Suggestion: when writing "Manual tests" it's better to try to list the manual tests you executed, see this description as example: https://github.com/aws/aws-parallelcluster/pull/4462

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Added.

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Changed this line back to its original content, to avoid introducing problems for users who might parse the error message according to its original format in their workflows.

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Tried to do that by storing the message to a variable but found it tricky because the global variable couldn't be easily modified within a local function. Thus I keep this part as it is (directly calling the error_exit function).

delongmeng:
> 

Thank you! I've added some details to the manual tests performed.

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -25,11 +26,11 @@ function download_run (){
...
     else
-      wget -qO- ${url} > $tmpfile || return 1
+      wget -qO- ${url} > $tmpfile || error_exit "Failed to run ${ACTION}, failed to download ${url} using wget, return code: $?"
     fi
-    chmod +x $tmpfile || return 1
+    chmod +x $tmpfile || error_exit "Failed to run ${ACTION}, failed to run chmod +x on temporary file, return code: $?"
```
chenwany: minor: the message temporary file doesn't looks clear to me. Can we change it to: failed to make file $tmpfile executable

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -15,6 +15,7 @@ function error_exit () {
   script=`basename $0`
   echo "parallelcluster: ${script} - $1"
   logger -t parallelcluster "${script} - $1"
+  echo "${script} - $1." "Note: If --rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details." > /var/log/parallelcluster/headnode_bootstrap_error_msg
```
chenwany: nit: can we remove `fetch_and_run -`? Users doesn't know what is fetch_and_run, it is a function name in our code and I think adding it doesn't help with better understanding the problem if we have it in the cloudformation. And can we also remove "Note:" 
```
"$1. If --rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details."
```

chenwany (commented):
Please also tests what happen when compute nodes have bootstrap error

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: yes I can remove it although "fetch_and_run -" (coming from ${script}) has been exposed to users.

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Do we want to expose the exact file name of $tmpfile? I did that at the beginning but changed to "temporary file" because I felt that the exact name of a temporary file is more confusing than helpful. What do you think? But yes I agree to change "run chmod +x" to something "make the temporary file executable".

delongmeng: Do we want to show something like "... failed to make file /tmp/tmp.6J3xYo8DNK executable ..."?

chenwany: ah I see... what about failed to make the file executable

delongmeng:
> Please also tests what happen when compute nodes have bootstrap error   
 
After discussion we decided to skip this test because the code change here doesn't seem to add any risk to compute nodes.

chenwany (dismissed):

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -21,22 +21,23 @@ function error_exit () {
   script=`basename $0`
   echo "parallelcluster: ${script} - $1"
   logger -t parallelcluster "${script} - $1"
+  echo "$1. If --rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details." > /var/log/parallelcluster/bootstrap_error_msg
```
davprat, line 24: I think I would not add `. If --rollback-on-failure was set to false, check /var/log/cfn-init.log and /var/log/cloud-init-output.log for more details.` to the error message file - we should do that when we call cfn-signal.

davprat (dismissed):
Can we move the suggest trouble shooting step to user_data.sh in a new PR?

cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
delongmeng: Yeah I wanted to do that as well, but I realized the logs we want the customers to check might differ according to the sources: whether it's from BYOS recipes, other chef recipes, or custom scripts. For example, for chef recipes, we would suggest them also check `chef-client.log`.

davprat: 👍

davprat (approved):

PR #1329, opened 2022-01-19
Signal CloudFormation with an explicit error message

chenwany:
### Description of changes
* Describe *what* you're changing and *why* you're doing these changes.
* Link to impacted open issues.

### Tests
Manually test:
example error message in cloudformation:
```
WaitCondition received failed message: 'Failed when running HeadInit for the configured scheduler plugin. 
Additional info can be found in /var/log/chef-client.log, /var/log/parallelcluster/scheduler-plugin.out.log 
and /var/log/parallelcluster/scheduler-plugin.err.log.' for uniqueId: i-06b941a21c505b401
```

### References
* Link to related PRs in other packages (i.e. cookbook, node).
* Link to documentation useful to understand the changes.

### Checklist
- [ ] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [ ] Check all commits' messages are clear, describing what and why vs how.
- [ ] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [ ] Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1596, opened 2022-11-15
[surface head node bootstrap errors] create chef handler to store chef errors

delongmeng:
### Description of changes
During the bootstrap of the head node, if any of chef recipes fails, a generic error message will be sent to CloudFormation stack without providing specific reasons of the failure. Here we create a chef handler and apply to recipes. The handler will write error messages into a dedicated file, which will be extracted and sent to CloudFormation stack.

### Tests
Manually tested.
* Test chef handler handling chef errors:
  - Introduce an error in the config recipe (for example aws-parallelcluster-config::head_node_base.rb), then create a cluster, got error message in CloudFormation stack as expected:
  - _Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in /var/log/chef-client.log and /var/log/cloud-init-output.log.): Chef::Exceptions::FileNotFound: template[bad_file.txt] (aws-parallelcluster::exception_test line 5) had an error: Chef::Exceptions::FileNotFound: Cookbook aws-parallelcluster (3.4.0) does not contain a file at any of these locations:
  templates/host-ip-10-0-0-49.ec2.internal/shared_storages/bad_file.txt.erb
  ......_
* Test chef handler not affecting existing BYOS handler:
  - Introduce an error in BYOS plugin and create a cluster, the error message still stay the same as implemented in this PR
  - _Failed when running HeadConfigure for the configured scheduler plugin. Additional info can be found in /var/log/chef-client.log, /var/log/parallelcluster/scheduler-plugin.out.log and /var/log/parallelcluster/scheduler-plugin.err.log._

### References
* [This PR](https://github.com/aws/aws-parallelcluster/pull/4540) shows how the error messages file is used.
* [This PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/1329/files) handles BYOS plugin errors.

### Checklist

### DIFF
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -1,11 +1,35 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+#
+# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
 
 module WriteChefError
+  # this class is used to handle chef errors and write the errors into a certain file if the file does not exist yet
   class WriteChefError < Chef::Handler
     def report
+      # the run_status object is initialized by Chef Infra Client and keep track of status of a Chef Infra Client run
+      # the "failed?" property is evaluated to be true when a Chef Infra Client run fails
+      # reference: https://docs.chef.io/handlers/#run_status-object
       if run_status.failed?
-        error_file = '/var/log/parallelcluster/bootstrap_error_msg'
+        error_file = node['cluster']['bootstrap_error_path']
         unless File.exist?(error_file)
-          message = "Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in /var/log/chef-client.log and /var/log/cloud-init-output.log.):"
+          if node['cluster']['node_type'] == 'HeadNode'
+            more_details = "/var/log/chef-client.log and /var/log/cloud-init-output.log"
+          else
+            more_details = "/var/log/cloud-init-output.log"
+          end
+          message = "Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in '#{more_details}'.):"
           Mixlib::ShellOut.new("echo '#{message}' '#{run_status.formatted_exception}' > '#{error_file}'").run_command
         end
       end
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -3,7 +3,7 @@
 #
 # Cookbook:: aws-parallelcluster
 #
-# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright:: 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
 # License. A copy of the License is located at
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -3,7 +3,9 @@ module WriteChefError
   class WriteChefError < Chef::Handler
     def report
       if run_status.failed?
-        Mixlib::ShellOut.new("echo 'Failed when running chef recipes: #{run_status.formatted_exception}' > /var/log/parallelcluster/headnode_bootstrap_error_msg").run_command
+        unless File.exist?('/var/log/parallelcluster/headnode_bootstrap_error_msg')
+          Mixlib::ShellOut.new("echo 'Failed when running chef recipes: #{run_status.formatted_exception}' > /var/log/parallelcluster/headnode_bootstrap_error_msg").run_command
+        end
       end
     end
   end
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -18,19 +18,20 @@ module WriteChefError
   # this class is used to handle chef errors and write the errors into a certain file if the file does not exist yet
   class WriteChefError < Chef::Handler
     def report
+      extend Chef::Mixin::ShellOut
       # the run_status object is initialized by Chef Infra Client and keep track of status of a Chef Infra Client run
       # the "failed?" property is evaluated to be true when a Chef Infra Client run fails
       # reference: https://docs.chef.io/handlers/#run_status-object
       if run_status.failed?
         error_file = node['cluster']['bootstrap_error_path']
         unless File.exist?(error_file)
-          if node['cluster']['node_type'] == 'HeadNode'
-            more_details = "/var/log/chef-client.log and /var/log/cloud-init-output.log"
-          else
-            more_details = "/var/log/cloud-init-output.log"
-          end
+          more_details = if node['cluster']['node_type'] == 'HeadNode'
+                           "/var/log/chef-client.log and /var/log/cloud-init-output.log"
+                         else
+                           "/var/log/cloud-init-output.log"
+                         end
           message = "Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in '#{more_details}'.):"
-          Mixlib::ShellOut.new("echo '#{message}' '#{run_status.formatted_exception}' > '#{error_file}'").run_command
+          shell_out("echo '#{message}' '#{run_status.formatted_exception}' > '#{error_file}'")
         end
       end
     end
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -3,8 +3,10 @@ module WriteChefError
   class WriteChefError < Chef::Handler
     def report
       if run_status.failed?
-        unless File.exist?('/var/log/parallelcluster/headnode_bootstrap_error_msg')
-          Mixlib::ShellOut.new("echo 'Failed when running chef recipes: #{run_status.formatted_exception}' > /var/log/parallelcluster/headnode_bootstrap_error_msg").run_command
+        error_file = '/var/log/parallelcluster/headnode_bootstrap_error_msg'
+        unless File.exist?(error_file)
+          message = "Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in /var/log/chef-client.log and /var/log/cloud-init-output.log.):"
+          Mixlib::ShellOut.new("echo '#{message}' '#{run_status.formatted_exception}' > '#{error_file}'").run_command
         end
       end
     end
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -1,0 +1,10 @@
+
+module WriteChefError
+  class WriteChefError < Chef::Handler
+    def report
+      if run_status.failed?
+        Mixlib::ShellOut.new("echo 'Failed when running chef recipes: #{run_status.formatted_exception}' > /var/log/parallelcluster/chef_error_msg").run_command
+      end
+    end
+  end
+end
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -3,7 +3,7 @@ module WriteChefError
   class WriteChefError < Chef::Handler
     def report
       if run_status.failed?
-        Mixlib::ShellOut.new("echo 'Failed when running chef recipes: #{run_status.formatted_exception}' > /var/log/parallelcluster/chef_error_msg").run_command
+        Mixlib::ShellOut.new("echo 'Failed when running chef recipes: #{run_status.formatted_exception}' > /var/log/parallelcluster/headnode_bootstrap_error_msg").run_command
       end
     end
   end
--- a/files/default/event_handler/write_chef_error_handler.rb
+++ b/files/default/event_handler/write_chef_error_handler.rb
@@ -3,7 +3,7 @@ module WriteChefError
   class WriteChefError < Chef::Handler
     def report
       if run_status.failed?
-        error_file = '/var/log/parallelcluster/headnode_bootstrap_error_msg'
+        error_file = '/var/log/parallelcluster/bootstrap_error_msg'
         unless File.exist?(error_file)
           message = "Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in /var/log/chef-client.log and /var/log/cloud-init-output.log.):"
           Mixlib::ShellOut.new("echo '#{message}' '#{run_status.formatted_exception}' > '#{error_file}'").run_command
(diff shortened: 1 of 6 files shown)


chenwany (dismissed):
LGTM

recipes/config.rb
```diff
@@ -15,4 +15,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+# enable chef error handler
```
enrico-usai: You can avoid this comment because the action we're doing is already explicit in the name of the recipe we're including `enable_chef_error_handler` .

recipes/finalize.rb
```diff
@@ -15,4 +15,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+# enable chef error handler
```
enrico-usai: same here

recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,13 @@
+
+cookbook_file "#{node['cluster']['scripts_dir']}/write_chef_error_handler.rb" do
+  source 'event_handler/write_chef_error_handler.rb'
+  owner 'root'
+  group 'root'
+  mode '0755'
+end
+
+chef_handler 'WriteChefError::WriteChefError' do
```
enrico-usai, line 18: Two comments here:

1. It seems we're using a chef resource that is deprecated: https://supermarket.chef.io/cookbooks/chef_handler - did you check the alternatives [here](https://docs.chef.io/handlers/#reporting) ?
2. If we're adding a dependency probably we should upgrade the Bersksfile accordingly and upload the dep in the `thirdy-party` cookbook folder.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,14 @@
+
```
enrico-usai, line 16: Add licensing.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,14 @@
...
+module WriteChefError
+  class WriteChefError < Chef::Handler
```
enrico-usai, line 19: It's worth adding a comment here to describe what this class is doing, and when someone should use it.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,14 @@
...
+    def report
+      if run_status.failed?
+        error_file = '/var/log/parallelcluster/bootstrap_error_msg'
```
enrico-usai: this error_file should be defined in the default/attributes.rb file and we should also define a variable for `/var/log/parallelcluster/` and use it everywhere.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,14 @@
+
+module WriteChefError
+  class WriteChefError < Chef::Handler
+    def report
```
enrico-usai, line 20: let's rename report to have a more self-explicative name, if it's possible.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,14 @@
...
+      if run_status.failed?
```
enrico-usai, line 25: where is `run_status` coming from? if it's from `Chef::Handler` let's add this info in the description of the method/class, to be clear which are the inputs and the outputs of this method we're defining.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,14 @@
...
+        error_file = '/var/log/parallelcluster/bootstrap_error_msg'
+        unless File.exist?(error_file)
+          message = "Failed when running chef recipes (If --rollback-on-failure was set to false, more details can be found in /var/log/chef-client.log and /var/log/cloud-init-output.log.):"
```
enrico-usai: From the message it seems to me that this can be only used by the head node, right? If true let's explicit this limitation somewhere: comment or name of the function.

I'm saying this because chef-client.log is not available in the compute, so in that case the message should be different. We can for example instead point to: https://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting-v3.html

recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,13 @@
+
+cookbook_file "#{node['cluster']['scripts_dir']}/write_chef_error_handler.rb" do
+  source 'event_handler/write_chef_error_handler.rb'
```
enrico-usai: why do you need to create the `write_chef_error_handler.rb` file in the node? 
I mean this file is already part of the cookbook recipes installed with ParallelCluster cookbook. 

Can you just enable the existing file? What am I missing? If I'm missing something let's add a comment to explain why do we need to copy this file.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: The run_status object is initialized by Chef Infra Client and keep track of status of a Chef Infra Client run, and the "failed?" property will be evaluated to be true when a Chef Infra Client run fails. Reference: https://docs.chef.io/handlers/#run_status-object. I've added the description as comments in the script.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: I tried to rename the function to be more specific (such as write_error_file), but then the function seemed to fail to get invoked upon chef error. 
The report handler seems to be a built-in chef handler type, and within a custom handler example in chef handler documentation, "def report" is still used. (Reference: https://docs.chef.io/resources/chef_handler/, search for "class CookbookVersionsHandler")

delongmeng: Do you have any idea of how to explicitly invoke the specific function if we change the name? Right now, the report handler works without needing to be explicitly called.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: Yes I added a comment.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: Added.

recipes/config.rb
delongmeng: Removed.

recipes/finalize.rb
delongmeng: Removed.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: I added `default['cluster']['log_base_dir'] = '/var/log/parallelcluster'`, and `default['cluster']['bootstrap_error_path']` based on that. Also changed `default['cluster']['scheduler_plugin']['handler_log_out']` and `default['cluster']['scheduler_plugin']['handler_log_err']` to be based on `default['cluster']['log_base_dir']` accordingly.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: Sorry for the confusion. I do think this chef handler will also be used for compute nodes, because the handler is added to 3 common entry point recipes that are shared by both head node and compute node recipes (for example here https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-config/recipes/base.rb#L71). [BTW, shouldn't the cookbook [here](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-config/recipes/base.rb#L4) be `aws-parallelcluster-config`?]

With that being said, currently, users will only see this message in CloudFormation stack if the error comes from the head node, because we only have the cfn-signal in error_exit() in headnode/user_data.sh, see the code [here](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/resources/head_node/user_data.sh#L73). 

To avoid confusion for the error message stored in the error message file within the compute node, I added a check in the chef handler function to write slightly different message according to the node type. Does that make sense to you? I personally feel pointing to specific locations (if possible) is better than pointing to a general troubleshooting doc.

recipes/enable_chef_error_handler.rb
delongmeng: You are right! After some exploration I confirmed that we don't need the cookbook_file section. We just need to use the full absolute path of the recipe as the "source" for the chef_handler part. I've edited the code accordingly.

recipes/enable_chef_error_handler.rb
delongmeng: Thanks for the reminder. Correct me if I'm wrong (because I'm new to chef), but I feel that we are not using the deprecated chef_handler **cookbook**? 

(1) From your first link it says:

> "The chef_handler resource from **this cookbook** is **now shipping as part of Chef 14**. With **the inclusion of this resource into Chef itself** we are now deprecating this cookbook."

From the chef-client.log I see that we are using `chef_version=17.2.29`, so I'd assume we are not using the chef_handler **cookbook**? Instead, we are using resources from Chef itself? The alternative link you mentioned is indeed the webpage that I have been using as a reference. Another [webpage](https://docs.chef.io/resources/chef_handler/) about chef_handler resource also consistently mentions "New in Chef Infra Client 14.0". 

(2) If we were using the chef_handler **cookbook**, then where are the recipes from that cookbook (for example [these ones](https://github.com/chef-boneyard/chef_handler))? I don't find any in our codebase, or in the head node (under /etc/chef/cookbooks/), and also don't find clues when/where we download that cookbook.

Please let me know your thoughts and whether any additional actions are needed?

enrico-usai: Perfect, thanks for the detailed explanation. I'm aligned. We can keep it, thanks!

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+#
+# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
enrico-usai: minor: the copyright should be: 2022 Amazon.com because we created the file this year.

files/default/event_handler/write_chef_error_handler.rb
enrico-usai: Thanks!

files/default/event_handler/write_chef_error_handler.rb
enrico-usai: Let's keep it as is, thanks!

enrico-usai:
I cannot approve it because there are some issues in the ruby linters.

delongmeng:
> I cannot approve it because there are some issues in the ruby linters.

Yes I'll fix the license version and work on the failed check next Monday.

files/default/event_handler/write_chef_error_handler.rb
delongmeng: fixed

enrico-usai (approved):

recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,22 @@
...
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+chef_handler 'WriteChefError::WriteChefError' do
+  source "/etc/chef/cookbooks/aws-parallelcluster/files/default/event_handler/write_chef_error_handler.rb"
+  type exception: true
```
enrico-usai, line 20: According to [official doc](https://docs.chef.io/resources/chef_handler/#:~:text=a%20Ruby%20gem.-,type,of%20handler%20to%20register%20as%2C%20i.e.%20%3Areport%2C%20%3Aexception%20or%20both.,-Common%20Resource%20Functionality) by default both exception and report are enabled.

Probably you have to use `type :report => false, :exception => true` and after doing this you can remove the `if run_status.failed?` check in the handler, because it will be called only at exception case.

files/default/event_handler/write_chef_error_handler.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
```
awslitvit, line 1: Did you try to put this file under `libraries` folder?

In my POC I defined the custom handler under `libraries` and it was sufficient to enable it without specifying the source file.

delongmeng: I didn't try the libraries folder. Will probably give it try next week.

recipes/enable_chef_error_handler.rb
delongmeng: Yes, according to the doc, there are 3 types of handlers: exception, report, and start. Of them,
- An exception handler runs when the `failed?` property for the `run_status` object returns `true`.
- A report handler runs when the `success?` property for the `run_status` object returns `true`.

Do you know what's the difference between 
(1) specifying `type :report => false, :exception => true` in `enable_chef_error_handler.rb`, and 
(2) changing `def report` to `def exception` in `write_chef_error_handler.rb`?

I can play around with it (maybe next week), but wonder: if there're different ways to configure what we want, how do we know which is the best way?

delongmeng: I paused on this activity as I was back on the design doc last month. I'm now back on the Chef handler task and trying to address this comment within the new PR here: https://github.com/aws/aws-parallelcluster-cookbook/pull/1641

1. My conclusion 1: There is no need to specify `{type :report => false, :exception => true}`. It is true that the default value for the "type" property is `{"report"=>true, "exception"=>true}`, when it's _not specified_. As long as I specify it (for example here, using `type exception: true`), it will replace the default value _as a whole_ (note: NOT just partially replace the `"exception"=>true` and leave `"report"=>true`). In this case here, by specifying `type exception: true`, my handler will only be registered as an exception handler.
2. Support for conclusion 1: I checked chef-client logs for failed runs and successful runs, and confirmed that this handler will only be executed when the run fails:

- When the run fails, the handler `WriteChefError::WriteChefError` will be executed:
```
Running handlers:
[2023-01-13T01:36:58+00:00] ERROR: Running exception handlers
[2023-01-13T01:36:58+00:00] INFO: run_status.formatted_exception:
[2023-01-13T01:36:58+00:00] INFO: 'Timeout::Error: execution expired'
......
  - WriteChefError::WriteChefError
Running handlers complete
[2023-01-13T01:36:58+00:00] ERROR: Exception handlers complete
```
- When the run succeeds, the handler `WriteChefError::WriteChefError` will not be executed:
```
Running handlers:
[2023-01-12T04:08:20+00:00] INFO: Running report handlers
Running handlers complete
[2023-01-12T04:08:20+00:00] INFO: Report handlers complete
```
3. This stays true even after I removed the `if run_status.failed?` check. So this comes to my 2nd conclusion: the `if run_status.failed?` check can be safely removed since the handler is an exception handler and will only be executed when the run fails.
4. My previous thought of changing `def report` to `def exception` would probably not work, because the type of the handler (whether a report handler or exception handler) is not defined by the name of the function.
5. Based on this I will remove the `if run_status.failed?` check in the new PR (https://github.com/aws/aws-parallelcluster-cookbook/pull/1641).

Does it address your concern @enrico-usai?

files/default/event_handler/write_chef_error_handler.rb
delongmeng: Sorry I paused this activity as I went back to work on the design doc last month. Now I've resumed the Chef handler activity and tried move the chef handler script to the `libraries` folder as you suggested. I confirmed that it worked and I didn't need to specifying the source file.

The code change can be found within a new PR: https://github.com/aws/aws-parallelcluster-cookbook/pull/1641/files

PR #1598, opened 2022-11-16
3.x Add configuration to clustermgtd.conf for collecting compute console output.

davprat:
### Description of changes
* Generate configuration for compute console output logging in clustermgtd.conf on cluster creation
* Update configuration for compute console output logging in clustermgtd.conf on cluster update
* Configure CloudWatch Agent to publish compute console output logs from head node.
### Testing
* Manually tested happy path (Monitoring enabled) cluster creation.
* Manually tested CloudWatch Agent configuration by making sure new log file was correctly exported to CloudWatch.
* No testing on update yet. 
* No testing on Monitoring disabled.

### Checklist

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -640,6 +640,8 @@ default['cluster']['enable_nss_slurm'] = node['cluster']['directory_service']['e
 default['cluster']['realmemory_to_ec2memory_ratio'] = 0.95
 default['cluster']['slurm_node_reg_mem_percent'] = 75
 default['cluster']['slurmdbd_response_retries'] = 30
+default['cluster']['slurm_console_logging']['default']['enabled'] = 'True'
+default['cluster']['slurm_console_logging']['default']['sample_size'] = 1
 
 # Official ami build
 default['cluster']['is_official_ami_build'] = false
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -640,8 +640,8 @@ default['cluster']['enable_nss_slurm'] = node['cluster']['directory_service']['e
 default['cluster']['realmemory_to_ec2memory_ratio'] = 0.95
 default['cluster']['slurm_node_reg_mem_percent'] = 75
 default['cluster']['slurmdbd_response_retries'] = 30
-default['cluster']['slurm_console_logging']['default']['enabled'] = 'True'
-default['cluster']['slurm_console_logging']['default']['sample_size'] = 1
+default['cluster']['slurm_console_logging']['enabled'] = 'true'
+default['cluster']['slurm_console_logging']['sample_size'] = 1
 
 # Official ami build
 default['cluster']['is_official_ami_build'] = false
--- a/cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_log_files.json
+++ b/cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_log_files.json
@@ -158,6 +158,23 @@
     },
     {
       "timestamp_format_key": "default",
+      "file_path": "/var/log/parallelcluster/compute_console_output.log",
+      "log_stream_name": "compute_console_output",
+      "schedulers": [
+        "slurm"
+      ],
+      "platforms": [
+        "amazon",
+        "centos",
+        "ubuntu"
+      ],
+      "node_roles": [
+        "HeadNode"
+      ],
+      "feature_conditions": []
+    },
+    {
+      "timestamp_format_key": "default",
       "file_path": "/var/log/parallelcluster/computemgtd",
       "log_stream_name": "computemgtd",
       "schedulers": [
--- a/cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
@@ -173,17 +173,10 @@ replace_or_add "update node replacement timeout" do
   replace_only true
 end
 
-replace_or_add "update compute console logging enabled" do
+replace_or_add "Update compute console logging" do
   path "/etc/parallelcluster/slurm_plugin/parallelcluster_clustermgtd.conf"
   pattern "node_replacement_timeout*"
-  line "compute_console_logging_enabled = #{node['cluster']['config'].dig(:Monitoring, :ComputeConsoleLoggingEnabled) || 'True'}"
-  replace_only true
-end
-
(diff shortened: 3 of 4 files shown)


davprat:
Check failures due to know issue with stunnel package:
```
#20 567.0     Error executing action `create` on resource 'remote_file[/opt/parallelcluster/sources/stunnel-5.67.tar.gz]'
#20 567.0     ================================================================================
#20 567.0     
#20 567.0     OpenSSL::SSL::SSLError
#20 567.0     ----------------------
#20 567.0     SSL Error connecting to https://www.stunnel.org/downloads/stunnel-5.67.tar.gz - SSL_connect returned=1 errno=0 state=error: certificate verify failed (certificate has expired)
#20 567.0     
#20 567.0     Resource Declaration:
#20 567.0     ---------------------
#20 567.0     # In /etc/chef/local-mode-cache/cache/cookbooks/aws-parallelcluster-install/recipes/efs.rb
#20 567.0     
#20 567.0      91:   remote_file stunnel_tarball do
#20 567.0      92:     source node['cluster']['stunnel']['url']
#20 567.0      93:     mode '0644'
#20 567.0      94:     retries 3
#20 567.0      95:     retry_delay 5
#20 567.0      96:     not_if { ::File.exist?(stunnel_tarball) }
#20 567.0      97:   end
#20 567.0      98: 
```

attributes/default.rb
```diff
@@ -638,6 +638,8 @@
 default['cluster']['realmemory_to_ec2memory_ratio'] = 0.95
 default['cluster']['slurm_node_reg_mem_percent'] = 75
 default['cluster']['slurmdbd_response_retries'] = 30
+default['cluster']['slurm_console_logging']['default']['enabled'] = 'True'
```
chenwany: question: Could you explain the benefit of using multiple nested dict here to store the default value and enable? What about simply use `default['cluster']["compute_console_logging_max_sample_size"]` and `default['cluster']["compute_console_logging_enabled"]` because the clustermgtd configuration is also not nested?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_log_files.json
```diff
@@ -156,6 +156,23 @@
       ],
       "feature_conditions": []
     },
+    {
+      "timestamp_format_key": "default",
```
chenwany, line 160: Is this `timestamp_format_key` manually tested? Asking because the console_output.logs are from system messge. And system message are in the format of
```
"timestamp_format_key": "month_first",
```

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
```diff
@@ -173,6 +173,13 @@ def update_nodes_in_queue(strategy, queues)
   replace_only true
 end
 
+replace_or_add "Update compute console logging" do
+  path "/etc/parallelcluster/slurm_plugin/parallelcluster_clustermgtd.conf"
+  pattern "node_replacement_timeout*"
```
chenwany: Is the pattern need to be update here?

chenwany (commented):
General question that not related to this PR: If there's a long lasting cluster, the logs will keep growing, how are we going to handle that? Is there a way to delete the logs when they are uploaded to the log group? Maybe we can leave it to when we go to log rotation story

attributes/default.rb
chenwany: Should it be enable only when https://docs.aws.amazon.com/parallelcluster/latest/ug/Monitoring-v3.html#yaml-Monitoring-Logs-CloudWatch-Enabled is enable?

davprat: It will use Monitoring-Logs-CloudWatch-Enabled if it is present, otherwise it defaults to True ([This line](https://github.com/aws/aws-parallelcluster-cookbook/pull/1598/files#diff-694753d857e75e10dfab299ef7cdb79106b6a6fb2b6812f637b490c9cc16b5d9R179)).

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
davprat: oof, yes you are correct.

delongmeng (commented):
Just a general question, what tests (maybe just some manual tests) have been done for the code change in this PR?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_log_files.json
davprat: I believe it is the same as the time format as clustermgtd - this entry is picking up the log file written out by a python logger that I've defined alongside clustermgtd's root logger and I've specified the format to be the same as clustermgtd.
So CloudWatch Agent will be parsing the event as written by clustermgtd and not parsing the contents (which is the system output) I think?

attributes/default.rb
davprat: As to the first question, to me it seems more readable and kind of groups things by purpose, but I'm not too opinionated on this.

davprat:
> General question that not related to this PR: If there's a long lasting cluster, the logs will keep growing, how are we going to handle that? Is there a way to delete the logs when they are uploaded to the log group? Maybe we can leave it to when we go to log rotation story

Yes, I was thinking of maybe adding log rotation for this specific log  to this story as a first step/experiment. I might do this after the beta goes out.

davprat:
> Just a general question, what tests (maybe just some manual tests) have been done for the code change in this PR?

Good call out, I should have documented this in the PR.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_log_files.json
chenwany: ok sounds good, let's test it with the node package PR

attributes/default.rb
davprat: removed ['default'] sub-tree.

cookbooks/aws-parallelcluster-slurm/recipes/update_head_node.rb
davprat: fixed

chenwany (dismissed):
LGTM! lets test the timestamp for the log stream before merging it

chenwany (dismissed):

chenwany (approved):

PR #1641, opened 2023-01-13
[head node bootstrap error] revise the chef handler to detect mount failure otherwise extract recipe info

delongmeng:
### Description of changes
This is a follow-up effort on a previous PR https://github.com/aws/aws-parallelcluster-cookbook/pull/1596, where we only extracted exception information of failed Chef runs. In this new PR, we:
* Detect storage mount failures of EBS, EFS, FSX, and RAID from Chef recipe exceptions
* Otherwise extract chef recipe name and code line
* Also address some previous comments from https://github.com/aws/aws-parallelcluster-cookbook/pull/1596, regarding Chef handler type and location of the Chef handler script.

### Tests
Manually tested.

* Test chef handler not executed when the cluster creation succeeds:
  - Created a cluster using pcluster CLI, checked the chef-client.log file
  - Outcome: the Chef handler script was not executed
* Test chef handler handling storage mount errors:
  - Create a cluster with invalid EBS volume id, got error message in CloudFormation stack as expected:
  - _WaitCondition received failed message: 'Failed to mount EBS volume. Please check /var/log/chef-client.log in the head node, or check the chef-client.log in CloudWatch logs. Please refer to https://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting-v3.html#troubleshooting-v3-get-logs for more details on ParallelCluster logs.' for uniqueId: i-03d07c8dbf3a6d48d_
  - Similarly, for RAID mount issue, got "Failed to mount RAID array......"; for EFS mount issue, got "Failed to mount EFS......"; for FSX mount issue, got "Failed to mount FSX......".
* Test chef handler handling other Chef failures:
  - Introduce an error in the config recipe (for example aws-parallelcluster-config::head_node_base.rb), then create a cluster, got error message in CloudFormation stack as expected:
  - _WaitCondition received failed message: 'Failed to run chef recipe aws-parallelcluster-config::head_node_base line 102. Please check /var/log/chef-client.log in the head node, or check the chef-client.log in CloudWatch logs. Please refer to https://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting-v3.html#troubleshooting-v3-get-logs for more details on ParallelCluster logs.' for uniqueId: i-0d5ad5b16630327ec_

### References
* Related PR: https://github.com/aws/aws-parallelcluster-cookbook/pull/1596)
* Related resources: https://docs.chef.io/resources/chef_handler/, https://docs.chef.io/handlers/#custom-handlers

### Checklist


libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+#
+# Copyright:: 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
chenwany: 2023？

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,99 @@
...
+          ' for more details on ParallelCluster logs.'
+
+        # get the failed action records using the chef function filtered_collection
+        # reference: https://github.com/cinc-project/chef/blob/stable/cinc/lib/chef/action_collection.rb#L107
+        failed_action_collection = action_collection.filtered_collection(
+          up_to_date: false, skipped: false, updated: false, failed: true, unprocessed: false
+        )
+
+        # define a mapping from the mount-related resource name to the error message we would like to display
+        mount_message_mapping = {
```
chenwany, line 44: Instead of putting it in the single function, can we have a function use for checking share storage error?

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,99 @@
...
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+module WriteChefError
+  # this class is used to handle chef errors and write the errors into a certain file if the file does not exist yet
+  class WriteChefError < Chef::Handler
+    def report
+      extend Chef::Mixin::ShellOut
+      error_file = node['cluster']['bootstrap_error_path']
```
chenwany, line 22: Why run_status.failed is not checking here?

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,99 @@
...
+          "mount fsx" => "Failed to mount FSX.",
+        }
+
+        failed_action_collection.each do |action_record|
+          # there might be multiple failed action records
+          # here we only look at the outermost layer resource by setting nesting_level = 0
+          # with the assumption that there is only one failed action record with nesting_level = 0
+          next unless action_record.nesting_level == 0
+          # we first check if it is a storage mounting error
+          if action_record.action == :mount
```
chenwany: what if when the action fail with `export`?

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,99 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+module WriteChefError
+  # this class is used to handle chef errors and write the errors into a certain file if the file does not exist yet
+  class WriteChefError < Chef::Handler
+    def report
+      extend Chef::Mixin::ShellOut
```
chenwany, line 21: Why do we need to add `extend Chef::Mixin::ShellOut` here?

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,99 @@
...
+          "add raid" => "Failed to mount RAID array.",
+          "mount efs" => "Failed to mount EFS.",
+          "mount fsx" => "Failed to mount FSX.",
+        }
+
+        failed_action_collection.each do |action_record|
+          # there might be multiple failed action records
+          # here we only look at the outermost layer resource by setting nesting_level = 0
+          # with the assumption that there is only one failed action record with nesting_level = 0
+          next unless action_record.nesting_level == 0
```
chenwany, line 55: are we sure that nesting_level is 0 for all mount, export actions?

libraries/write_chef_error_handler.rb
delongmeng: fixed

libraries/write_chef_error_handler.rb
delongmeng: After discussion we decided not to use a function now and leave it to the future when needed.

libraries/write_chef_error_handler.rb
delongmeng: It's not needed because this is an exception handler and will only be executed when run_status.failed is true.
Please refer to this discussion here: https://github.com/aws/aws-parallelcluster-cookbook/pull/1596#discussion_r1072881701

libraries/write_chef_error_handler.rb
delongmeng: In the previous PR, I used `Mixlib::ShellOut.new(xxx).run_command` at the beginning but was required to change to `shell_out` instead. To use the `shell_out` function, `extend Chef::Mixin::ShellOut` is needed.

libraries/write_chef_error_handler.rb
delongmeng: After discussion, we agree that using `nesting_level == 0` is the best option for now.

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,76 @@
...
+      # if the error file already exists we don't take any additional action here
+      unless File.exist?(error_file)
+        message_error = 'Failed to run chef recipe.'
+        node_type_to_log = {
+          'HeadNode' => 'Please check /var/log/chef-client.log in the head node, or check the chef-client.log in CloudWatch logs.',
+          'ComputeFleet' => 'Please check the cloud-init-output.log in CloudWatch logs.',
+        }
+        message_logs_to_check = node_type_to_log[node['cluster']['node_type']]
+        message_troubleshooting_link = 'Please refer to'\
+          ' https://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting-v3.html#troubleshooting-v3-get-logs'\
```
gmarciani, line 34: We should avoid the use of static URLs in error messages because public docs content may be moved and even may not be available at all. As an alternative: let's refer to a section in a doc without URLs. In this case: `Please refer to the Troubleshooting section in AWS ParallelCluster docs for more details on ParallelCluster logs.`

I know that we refer to public docs URLs in other places (eg error messages from CLI validators), but we should avoid it.

libraries/write_chef_error_handler.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      end
+    end
+
+    def get_recipe_info(action_record)
+      # use the built-in function "defined_at" of Chef::Resource to get the recipe information
+      # when source_line is not available it will return "dynamically defined" and we replace it with empty string
+      # reference: https://github.com/cinc-project/chef/blob/stable/cinc/lib/chef/resource.rb#L1436
+      recipe_info = action_record.new_resource.defined_at
+      recipe_info == "dynamically defined" ? "" : recipe_info
```
gmarciani: Isn't it possible to at least return the recipe name even if the precise failing source line is not available?

delongmeng: According to Chef codebase: 

- The `source_line` is the caller of the resource: https://github.com/cinc-project/chef/blob/stable/cinc/lib/chef/dsl/declare_resource.rb#L263 (It's usually available)

- A resource is considered as "dynamic resource" when `source_line` is null:
https://github.com/cinc-project/chef/blob/stable/cinc/lib/chef/formatters/error_inspectors/resource_failure_inspector.rb#L98 

- A resource usually falls into the definition of "Dynamically defined resource" when the cookbook and recipe name are not available. Reference: 
https://github.com/cinc-project/chef/blob/stable/cinc/lib/chef/formatters/doc.rb#L227

- All of the above information and the logic of the `defined_at` built-in method seems to assume that when `source_line` is not available, recipe information will also not be available anyway.

This is consistent with our observation that we've seen situations when we have `source_line` but not recipe name, but we never see any situation when the `source_line` is missing but the recipe name is available. Based on this I think we can keep use the `defined_at` method to avoid the code to be too complicated. Worst case scenario (in some very rare corner case and we can dig into it later), the returned error message is just the general catch-all message. Does that sound good to you?

libraries/write_chef_error_handler.rb
delongmeng: Thanks for pointing this out. This was a team decision during the design stage. After discussion we agreed that it's not a blocking issue for now, and we will consider it in the future.

chenwany (approved):

PR #1650, opened 2023-01-17
[Isolated Regions] Add support for isolated regions: us-iso-* and us-isob-*

gmarciani:
### Description of changes
Add support for US isolated regions: us-iso-* and us-isob-*.

In particular, we added:
1. support for the AWS domain in these regions;
2. configuration of the AWS CLI so that it uses the CA bundle dedicated to these regions,
stored at '/etc/pki/#{region}/certs/ca-bundle.pem'. Such configuration is applied at config time
to all the system users that are supposed to interact with AWS:
root, ec2-user (or equivalent), pcluster-admin, pcluster-scheduler-plugin and dcv.

### Tests
* New unit tests
* Custom build with validation enabled
* Cluster creation with custom built AMI
* Full regression tests will be done by the authoritative pipeline

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-config
+# Recipe:: aws_cli
+#
+# Copyright:: 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Users to configure the AWS CLI for.
+# We configure the AWS CLI for all the users that are created by ParallelCluster and must interact with AWS.
+users = node['cluster']['head_node_imds_allowed_users']
+
+# AWS CLI configurations to set.
+# Note: only the configurations with a not nil value will be set.
+region = node['cluster']['region']
+configurations = {
+  'ca_bundle' =>  region.start_with?('us-iso') ? "/etc/pki/#{region}/certs/ca-bundle.pem" : nil,
+}
+
+def set_aws_cli_configuration(config_name, expected_value, user)
+  execute "Setting AWS CLI configuration for user #{user}: #{config_name} set to '#{expected_value}'" do
+    user user
+    login true
+    command "aws configure set #{config_name} #{expected_value}"
+  end
+end
+
+# AWS CLI is explicitly configured only in US iso regions.
+if region.start_with?('us-iso')
+  users.each do |user|
+    configurations.each do |config_name, config_value|
+      next if config_value.nil?
+      set_aws_cli_configuration(config_name, config_value, user)
+    end
+  end
+end
--- a/cookbooks/aws-parallelcluster-install/recipes/users.rb
+++ b/cookbooks/aws-parallelcluster-install/recipes/users.rb
@@ -28,7 +28,7 @@ user node['cluster']['cluster_admin_user'] do
   system true
   shell '/bin/bash'
   home "/home/#{node['cluster']['cluster_admin_user']}"
-  manage_home false
+  manage_home node['cluster']['node_type'] == 'HeadNode'
 end
 
 # Calling user_ulimit will override every existing limit
--- a/system_tests/install_cinc.sh
+++ b/system_tests/install_cinc.sh
@@ -31,12 +31,19 @@ elif [ `echo "${OS}" | grep -E '^ubuntu'` ]; then
   PLATFORM='DEBIAN'
 fi
 
-BUCKET="s3.amazonaws.com"
-[[ ${AWS_Region} =~ ^cn- ]] && BUCKET="s3.cn-north-1.amazonaws.com.cn/cn-north-1-aws-parallelcluster"
+AWS_DOMAIN="amazonaws.com"
+[[ ${AWS_Region} =~ ^cn- ]] && AWS_DOMAIN="amazonaws.com.cn"
+[[ ${AWS_Region} =~ ^us-iso- ]] && AWS_DOMAIN="c2s.ic.gov"
+[[ ${AWS_Region} =~ ^us-isob- ]] && AWS_DOMAIN="sc2s.sgov.gov"
+
+S3_ENDPOINT="s3.${AWS_Region}.${AWS_DOMAIN}"
+
+BUCKET="cloudformation-examples"
+[[ ${AWS_DOMAIN} != "amazonaws.com" ]] BUCKET="${AWS_Region}-aws-parallelcluster/cloudformation-examples"
 if [[ ${OS} =~ ^(ubuntu2004)$ ]]; then
-  CfnBootstrapUrl="https://${BUCKET}/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz"
+  CfnBootstrapUrl="https://${S3_ENDPOINT}/${BUCKET}/aws-cfn-bootstrap-py3-latest.tar.gz"
 else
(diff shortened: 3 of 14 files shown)


dreambeyondorange (dismissed):

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
```diff
@@ -0,0 +1,42 @@
...
+    user user
+    login true
+    command "aws configure set #{config_name} #{expected_value}"
+  end
+end
+
+users.each do |user|
+  configurations.each do |config_name, config_value|
+    next if config_value.nil? || config_value.empty?
+    set_aws_cli_configuration(config_name, config_value, user)
```
dreambeyondorange: We probably only want to do this if we are in the ISO regions right?

cookbooks/aws-parallelcluster-install/recipes/users.rb
```diff
@@ -28,7 +28,7 @@
   system true
   shell '/bin/bash'
   home "/home/#{node['cluster']['cluster_admin_user']}"
-  manage_home false
+  manage_home node['cluster']['node_type'] == 'HeadNode'
```
dreambeyondorange, line 31: Do we want to switch on this option based on the region?

system_tests/install_cinc.sh
```diff
@@ -33,6 +33,8 @@ fi
 
 BUCKET="s3.amazonaws.com"
```
dreambeyondorange: should we use the domain helper below?

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
gmarciani: Actually it does only for ISO regions.
The block is able to apply the AWS CLI configurations according to the dictionary `configurations` below, but it does only for config names that has a value and only ISO regions do have.

```
configurations = {
  "ca_bundle" =>  region.start_with?("us-iso") ? "/etc/pki/#{region}/certs/ca-bundle.pem" : "",
}
```

Do you think it's worth it to wrap the overall block with an if statement that executes it only for ISO?

cookbooks/aws-parallelcluster-install/recipes/users.rb
gmarciani: I think we can create the home directory for the user: this is also aligned to what we already do for other users, see the slurm user [here](https://github.com/aws/aws-parallelcluster-cookbook/blob/16100b20204ec4e4949d79bfab9c884028f27108/cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb#L42-L51)

system_tests/install_cinc.sh
gmarciani: no because this is used only by the Dockerfile that are used in system tests.

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
```
lukeseawalker: wrong cookbook name

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
```diff
@@ -0,0 +1,42 @@
...
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Users to configure the AWS CLI for.
+# We configure the AWS CLI for all the users that are created by ParallelCluster and must interact with AWS.
+users = node["cluster"]["head_node_imds_allowed_users"]
```
lukeseawalker: I think we are using single quotes to wrap variable names. I don't get why the linter doesn't notify about this

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
```diff
@@ -0,0 +1,42 @@
...
+
+def set_aws_cli_configuration(config_name, expected_value, user)
+  execute "Setting AWS CLI configuration for user #{user}: #{config_name} set to '#{expected_value}'" do
+    user user
+    login true
+    command "aws configure set #{config_name} #{expected_value}"
+  end
+end
+
+users.each do |user|
```
lukeseawalker: is there a reason why this is set at per user level and not at global os level?
I don't get the connection between the IDMS user Vs the user for which you are setting the ca_bundle into the AWS  configuration. Wasn't this setting supposed to be useful for every user in the system? I'm think about AD users for example, or any other custom user created

cookbooks/aws-parallelcluster-install/recipes/users.rb
lukeseawalker: what if ['node_type'] is nil as [your linked reference](https://github.com/aws/aws-parallelcluster-cookbook/blob/16100b20204ec4e4949d79bfab9c884028f27108/cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb#L42-L51)?

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
gmarciani: Thought about it. I will add the extra filter because if you asked the question then it means that the implicit filter on the configuration dictionary is not enough.

cookbooks/aws-parallelcluster-install/recipes/users.rb
gmarciani: I think the home directory should be created if and only if the resource is executed in the head node.
I'm not aware of any case where node_type equal to nil represent a meaningful situation.
So, I think we should leave the condition as it is. Do you agree?

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
gmarciani: Fixing thank you

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
gmarciani: got it fixing.

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
gmarciani: I completely agree with you: the configuration should be for every present and future system user.
Nevertheless, the AWS CLI does not provide a way to configure something "globally" using 'aws configure'.
A way to set a "global" configuration is to set the corresponding environment variables for every system users.
I've explored and tested this approach: it works, but I've excluded it because:
1. relying on environment variables is more error prone and less speaking than using dedicated configuration commands:
2. it requires changes to user profiles and sudoers settings, to cover system users.
3. it requires changes to the PAM scripts executed to set the home directory up.

**Regarding AD users:** we can still rely on the aws configure approach by including in this PR the step (3).
I've preferred to exclude to cover AD users because the customer is responsible for them.
Do you think we should include them anyway?

**Regarding IMDS users:** Assuming that we want to configure the AWS CLI on a per user basis, then it makes sense to configure it only for the users that we create and that we assume they need to have access to AWS.

cookbooks/aws-parallelcluster-config/recipes/aws_cli.rb
dreambeyondorange: sorry for the delayed response, I was actually thinking we wouldn't want this object `ca-bundle` to exist at all for non adc regions since we didn't need it before.

cookbooks/aws-parallelcluster-install/recipes/users.rb
dreambeyondorange: I am only mentioning this because I would err on the side of caution and avoid changing the default behavior for non adc clusters

lukeseawalker: ['node_type'] should be 'nil' at install time

gmarciani: Discussed with Luca about it: it's required to cover the nil case because at installation time we are not aware of the node type.

gmarciani:
Closed in favor of https://github.com/aws/aws-parallelcluster-cookbook/pull/1761, which directly targets the release branch.
We will back-port the final changes to develop once approved for the release branch.

PR #1662, opened 2023-01-20
Create a nfs resource for each OS and add support for redhat8

enrico-usai:
### Description of changes
* Rename default action from `prepare` to `setup`.
* [Documentation](https://docs.chef.io/custom_resources/) sais that the default action is the first action listed, but it's not clear in case of usage of partial files, for this reason set `default_action` to be sure which is the default.
* Create a nfs resource for each OS and add support for redhat8
* Delete useless example resource

### Tests
* Improve NFS resource tests

### Checklist


test/resources/controls/aws_parallelcluster_install/nfs_spec.rb
```diff
@@ -8,8 +8,9 @@
     end
 
     # Check version of NFS
-    describe command("rpcinfo -p localhost") do
-      its('stdout') { should match(/4.*tcp.*2049.*nfs/) }
+    nfs_version = command("rpcinfo -p localhost | awk '{print $5$2}' | grep nfs4")
+    describe nfs_version do
```
awslitvit: Does it improve the description in case of failure? Could you specify what is the message in that case?

awslitvit (dismissed):

test/resources/controls/aws_parallelcluster_install/nfs_spec.rb
enrico-usai: The error would be:
``` 
     ×  Command: `rpcinfo -p localhost | awk '{print $5$2}' | grep nfs4` stdout is expected to match "nfs4"
     expected "" to match "nfs4"
```

I agree this is not the best failure message. I read about `expect` vs `should` that permits to set more clear error message, but I didn't want to spend too much time on simplifying this within this patch.

This is a small improvement because having the variable `nfs_version` is at least clear what we're searching for. 
If you have simple suggestions to improve this check please share, thanks!

enrico-usai: Update, after adding the external `describe` block:
``` 
     ×  Verify installed NFS version is 4
      Command: `rpcinfo -p localhost | awk '{print $5$2}' | grep nfs4` stdout is expected to match "nfs4"
     expected "" to match "nfs4"
```

awslitvit (approved):

PR #1679, opened 2023-01-25
Add node attributes json generation in /etc/chef/node_attributes.json

francesco-giordano:
### Description of changes
* The function will export all the node attribute to a json file at the beginning of the install phase. Doing so other scripts can read the file instead of point directly to the default.rb

### Tests
* Tested in Jenkins

### References
* https://github.com/aws/aws-parallelcluster/pull/4823


awslitvit:
- The resource must be a recipe.
- We should export only specific values, decoupled from the internal attributes structure.

awslitvit (approved):

PR #1702, opened 2023-01-31
RHEL8: Define new recipe to automatically import recipes and resources deps

enrico-usai:
### Description of changes

* This new recipe will permit to define tests requiring multiple dependencies in term of resources and recipes.
  * Intead of using the run_list attribute to define the list of recipes we are now using the dependencies attribute, on which it's possible to define a list of deps, identified by the recipe: or resource: prefix.

* Rename sudo, users, awscli controls to have a unique convention
  * The name of the control for the inspec test should describe the test we're doing and should not correspond to the name of the recipe.

### Tests
Adapted existing tests accordingly.

### DIFF
--- a/kitchen.recipes.yml
+++ b/kitchen.recipes.yml
@@ -12,6 +12,11 @@ verifier:
     - test/recipes
     - test/resources
 
+# If you need to test a recipe with recipe/resources dependencies,
+# add recipe[aws-parallelcluster::add_dependencies] as first item in the run_list
+# then define a dependencies attribute, listing them with recipe: or resource: prefix.
+# You can find an example in the add_depdendencies.rb file.
+
 suites:
   - name: sudo
     run_list:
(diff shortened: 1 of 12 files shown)


kitchen.recipes.yml
```diff
@@ -12,21 +12,29 @@ verifier:
...
+        - sudo_installed
+        - sudoers_file_configured
   - name: users
     run_list:
-      - recipe[aws-parallelcluster::test_dummy]
+      - recipe[aws-parallelcluster::add_dependencies]
       - recipe[aws-parallelcluster-install::users]
     verifier:
+      dependencies:
+        - recipe:aws-parallelcluster::test_dummy]
```
francesco-giordano: **typo:** is the last `]` a typo?

enrico-usai: fixed, thanks!

francesco-giordano (approved):

PR #1710, opened 2023-02-02
[RHEL8][Slurm] Isolate mysql attributes

NSsirena:
### Description of changes
* Removes MySql related attributes from main
* Defines an action class with common behavior
* Defines a new partial to replace platform depended attributes with actions
* Revise mysql validation

### Tests
* Testes successfully both on Docker and EC2.

Note: since the validation is launched from a different endpoint, to force its tests I used this temporary config
in `kitchen.recipes.yaml`
```
  - name: validate-mysql
    run_list:
      - recipe[aws-parallelcluster::add_dependencies]
      - recipe[aws-parallelcluster-test::validate_mysql]
    verifier:
      controls:
        - mysql_client_installed
        - mysql_client_source_node_created
    attributes:
      dependencies:
        - recipe:aws-parallelcluster::test_dummy
        - recipe:aws-parallelcluster-install::directories
        - resource:mysql_client
```

### Checklist
- [ ] ~~Check if documentation is impacted by this change.~~


libraries/helpers.rb
```diff
@@ -477,6 +470,16 @@ def get_metadata_with_token(token, uri)
...
+def configure_gc_thresh_values
+  (1..3).each do |i|
+    # Configure gc_thresh values to be consistent with alinux2 default values
+    sysctl "net.ipv4.neigh.default.gc_thresh#{i}" do
+      value node['cluster']['sysctl']['ipv4']["gc_thresh#{i}"]
+      action :apply
+    end
+  end
+end
+
```
NSsirena: Please double check if this is required.
It isn't part of my changes. It's likely a leftover of a rebase.

enrico-usai: You can remove this part, it's now moved in the cookbooks/aws-parallelcluster-install/recipes/gc_thresh_values.rb recipe

libraries/helpers.rb
```diff
@@ -281,13 +281,6 @@ def reload_network_config
   end
 end
 
-#
-# Check if this is an ARM instance
-#
-def arm_instance?
```
enrico-usai, line 294: This is used by a lot of recipes, I don't think we can move it.

cookbooks/aws-parallelcluster-test/metadata.rb
```diff
@@ -14,3 +14,4 @@
 supports 'ubuntu', '>= 18.04'
 
 depends 'aws-parallelcluster-common', '~> 3.6.0'
+depends 'aws-parallelcluster-install', '~> 3.6.0'
```
enrico-usai: Why are we adding a dependency for -install cookbook?

cookbooks/aws-parallelcluster-install/resources/mysql_client/mysql_client.rb
```diff
@@ -38,3 +39,47 @@
...
+          '20.04' => "ubuntu/20.04/x86_64",
+          '18.04' => "ubuntu/18.04/x86_64",
+        }
+      )
+    end
+  end
+
+  def expected_version
+    if arm_instance?
+      "8.0.31"
```
enrico-usai, line 62: If we want to change the version of mysql we need to change this value in multiple places: expected_version, package_version and package_source_version

Can we use it the `package_source_version`  variable?

libraries/helpers.rb
NSsirena: I moved it in parallelcluster-commong that should be included in all recipes

cookbooks/aws-parallelcluster-test/metadata.rb
NSsirena: It was needed to test the change in the mysql-validate. Probably it is not needed anymore

cookbooks/aws-parallelcluster-install/resources/mysql_client/mysql_client.rb
NSsirena: This was and it is used by the mysql-validate. I kept it since I assumed we needed a different value for checking.

We can remove it in a future iteration if we consider it a duplicate.

awslitvit (approved):

PR #1724, opened 2023-02-06
Split package_repos resource and extend docker testing

enrico-usai:
### Description of changes
* Thanks to `cookbooks/aws-parallelcluster-test/recipes/docker_mock.rb` the actions are mocked so we can remove the virtualized check.
* gc_thresh_values and openssh are now tested on docker too.
* Permit to specify the action name when adding a resource as dep
* Split repos setup and update actions in the package_repos resource. Call the update actions for openssh test, requiring only apt_update call for Ubuntu 20.

### DIFF
--- a/kitchen.resources.yml
+++ b/kitchen.resources.yml
@@ -36,7 +36,6 @@ suites:
 
 <% [
   'package_repos',
-  'nfs',
   ].each do |resource| %>
   - name: <%= resource %>
     run_list:
@@ -51,8 +50,6 @@ suites:
 <%end %>
   - name: c_states
     run_list:
-      # This is a test recipe executing the default action of the resource you pass as `resource` attribute,
-      # with default properties
       - recipe[aws-parallelcluster-install::test_resource]
     verifier:
       controls:
@@ -82,3 +79,13 @@ suites:
         - recipe:aws-parallelcluster::test_dummy
         - recipe:aws-parallelcluster-install::directories
       resource: mysql_client
+  - name: nfs
+    run_list:
+      - recipe[aws-parallelcluster-install::test_resource]
+    verifier:
+      controls:
+        - nfs_installed_with_right_version
+    attributes:
+      resource: nfs
+      dependencies:
+        - recipe:aws-parallelcluster-test::docker_mock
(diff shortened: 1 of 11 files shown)


kitchen.resources.yml
```diff
@@ -36,7 +36,6 @@ suites:
 
 <% [
```
awslitvit: Maybe we can remove this erb now?

test/recipes/controls/aws_parallelcluster_install/openssh_spec.rb
```diff
@@ -25,10 +23,15 @@
   end
 
   files = %w(/etc/ssh/sshd_config /etc/ssh/ca_keys /etc/ssh/revoked_keys)
+  file_permission = if os.debian?
+                      '0644'
+                    else
+                      '0600'
+                    end
```
awslitvit, line 30: Just out of curiosity: why?

recipes/add_dependencies.rb
```diff
@@ -16,14 +17,22 @@
 #    dependencies:
 #      - recipe:aws-parallelcluster::test_dummy
 #      - recipe:aws-parallelcluster-install::directories
-#      - resource:ec2_udev_rules
+#      - resource:package_repos:update
```
awslitvit: Nitpick. I would keep both examples, the one with the default action and the one with the explicit action.

awslitvit (dismissed):

kitchen.resources.yml
enrico-usai: Removed

recipes/add_dependencies.rb
enrico-usai: Restored previous example too.

test/recipes/controls/aws_parallelcluster_install/openssh_spec.rb
enrico-usai: I verified it in the OS and this is the default. 
Not sure why in Ubuntu this is not aligned with RHEL but the root group has the possibility to read these files.

awslitvit (approved):

PR #1725, opened 2023-02-06
Add test for CloudWatch Agent configuration recipe.

ermanno:
### Description of changes
Add test for CloudWatch Agent configuration recipe.

### Tests
Tested on Docker and EC2.

*Note that on RHEL8 Docker we skip installation/configuration of CW Agent as well as the config test, because we have a limitation that we cannot install PyDev, which is also the reason `aws-parallelcluster-install::python` and its test skips execution on RHEL8 Docker.*

**The code has been run in an EC2 instance with RHEL8, and the test validates that the CW Agent is running.**

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-install/recipes/cloudwatch_agent.rb
+++ b/cookbooks/aws-parallelcluster-install/recipes/cloudwatch_agent.rb
@@ -17,6 +17,10 @@
 
 # Download the cloudwatch agent's public key. Note that the domain used to get
 # the public key must NOT be regionalized.
+
+# TODO: find a way to make this code work on ubi8
+return if redhat_ubi?
+
 remote_file node['cluster']['cloudwatch']['public_key_local_path'] do
   source node['cluster']['cloudwatch']['public_key_url']
   retries 3
@@ -95,6 +99,6 @@ when 'ubuntu'
   dpkg_package package_path do
     source package_path
   end
-when 'amazon', 'centos'
+when 'amazon', 'centos', 'redhat'
   package package_path
 end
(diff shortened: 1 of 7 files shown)


test/recipes/controls.aws_parallelcluster_config/cloudwatch_agent_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  describe file('/usr/local/bin/write_cloudwatch_agent_json.py') do
+    it { should exist }
+    its('mode') { should cmp '0755' }
+    its('owner') { should eq 'root' }
+    its('group') { should eq 'root' }
+    its('content') { should_not be_empty }
+  end
+
+  describe file('/usr/local/etc/cloudwatch_log_files.json') do
```
enrico-usai: nitpick: in cases like this you can use a loop:

```ruby 
files = %w(/usr/local/etc/cloudwatch_log_files.json /usr/local/etc/cloudwatch_log_files_schema.json /usr/local/bin/cloudwatch_log_configs_util.py)
  files.each do |file|
    describe file(file) do
      it { should exist }
      its('mode') { should cmp '0644' }
      its('owner') { should eq 'root' }
      its('group') { should eq 'root' }
    end
  end
```

cookbooks/aws-parallelcluster-install/recipes/cloudwatch_agent.rb
```diff
@@ -95,6 +99,6 @@
   dpkg_package package_path do
     source package_path
   end
-when 'amazon', 'centos'
+when 'amazon', 'centos', 'redhat'
```
enrico-usai, line 102: When there are if/else related to the Oses we are moving the code into a resource.

We should create a cloudwatch_agent resource, with two actions: setup and configure, move it in aws-parallelcluster-common/resources and call setup in the aws-parallelc-luster_install case and prepare in the config one.

ermanno: As discussed off-line, this will be part of a future task.

enrico-usai (approved):

PR #1732, opened 2023-02-07
[headnode metrics] add headnode disk and memory metrics to CloudWatch agent configuration

delongmeng:
### Description of changes
* Add new CloudWatch metrics in order to monitor memory usage and disk usage of the root volume on head node.
* Use "InstanceId" as append dimension, so that we can use these metrics later by instance id for downstream usage of the metrics such as visualization and alarms.
* Use "InstanceId, path" as an aggregation dimension so that we can use InstanceId and path to specify a disk metric without the need to retrieve device and file system type information.

### Tests
Manual tests: 

After code change, create a new cluster, and validate that new metrics (under the "CWAgent" namespace) show up in CloudWatch as expected:

 * Under the "InstanceId, device, fstype, path" dimension
    * disk_used_percent
 * Under the "InstanceId" dimension
    * mem_used_percent
  * Under the "InstanceId, path" dimension (aggregated)
    * disk_used_percent 

For a head node of c5.large Nitro instance, confirmed that the "reboot" of the head node didn't affect the metrics.

### References
* [PR to add CloudWatch Dashboard graphs based on these metrics](https://github.com/aws/aws-parallelcluster/pull/4869)
* [PR to add CloudWatch alarms based on these metrics](https://github.com/aws/aws-parallelcluster/pull/4877)

### Checklist


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 
 **ENHANCEMENTS**
 - Enforce the DCV Authenticator Server to use at least `TLS-1.2` protocol when creating the SSL Socket.
+- Add CloudWatch metrics: disk usage percentage (root volume) and memory usage percentage of the head node.
```
chenwany: What about: Track head node memory and root volume disk utilization using the `mem_used_percent` and `disk_used_percent` metrics collected trough the CloudWatch Agent.

chenwany (commented):
Could you add in the PR description why we need to specify `append_dimensions` and `aggregation_dimemsions` in the config?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -177,14 +178,74 @@ def filter_output_fields(configs):
...
+    metrics_collected = {}
+    metrics_collected = add_metrics_mem(metrics_collected, args.node_role)
+    metrics_collected = add_metrics_disk(metrics_collected, args.node_role)
+    metric_configs["metrics_collected"] = metrics_collected
+    return metric_configs
+
+
+def add_metrics_mem(metrics_collected, node_role):
+    """Add memory metrics for the metrics_collected field of CloudWatch Agent Metrics section."""
+    if node_role == "HeadNode":
```
chenwany: As it is discussed, can we do this with metrics schema like what we do for cloudwatch logs schema so that in the future it will be easier for us to append more metrics?

CHANGELOG.md
delongmeng: Sounds good. Revised accordingly.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
delongmeng: Yes. I moved the information of metrics to the configuration json file, updated the json schema, and renamed the files accordingly (from specific for "log_files" to general for CloudWatch agent).

delongmeng:
> Could you add in the PR description why we need to specify `append_dimensions` and `aggregation_dimemsions` in the config?

Revised the description to explain the dimensions.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -177,14 +178,68 @@ def filter_output_fields(configs):
     return [{desired_key: config[desired_key] for desired_key in desired_keys} for config in configs]
 
 
-def create_config(log_configs):
+def create_metrics_collected(selected_configs):
+    """Create the "metrics_collected" section in metrics configuration for selected metrics."""
+    metrics_collected = {}
```
chenwany: What about simplify the code to be something like:

(draft, to be verified)
```
def create_metrics_collected(selected_configs):
    “”"Create the “metrics_collected” section in metrics configuration for selected metrics.“”"
    desired_keys = [
        “measurement”,
        “metrics_collection_interval”,
        “resources”
    ]

    def _collect_metric(metric_config):
        collected = {key: metric_config.get(key) for key in metric_config if key in desired_keys}
        if “append_dimensions” in metric_config and “ClusterName” in metric_config[“append_dimensions”]:
            collected[“append_dimensions”] = {“ClusterName”: get_node_info().get(“stack_name”)}
        return collected

    return {metric_config[“metric_type”]: _collect_metric(metric_config) for metric_config in selected_configs}
```

Also, do we really need to append `ClusterName` dimension for now? It is not used in the cli to generate dashboard or create alarm?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -177,14 +178,68 @@ def filter_output_fields(configs):
...
+        metric_key = metric_config["metric_type"]
+        metrics_collected[metric_key] = {
+            "measurement": metric_config["measurement"],
+            "metrics_collection_interval": metric_config.get(
+                "metrics_collection_interval", DEFAULT_METRICS_COLLECTION_INTERVAL
+            ),
+        }
+        if "resources" in metric_config.keys():
+            metrics_collected[metric_key]["resources"] = metric_config["resources"]
+        if "append_dimensions" in metric_config.keys() and "ClusterName" in metric_config["append_dimensions"]:
```
chenwany: nit: `if "append_dimensions" in metric_config.keys()` can be simplified as if "append_dimensions" in metric_config

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -177,14 +178,68 @@ def filter_output_fields(configs):
...
+        "InstanceType": "${aws:InstanceType}",
+        "AutoScalingGroupName": "${aws:AutoScalingGroupName}",
+    }
+    append_dimensions = {key: valid_append_dimensions[key] for key in dimensions}
+    return append_dimensions
+
+
+def add_append_dimensions(metric_configs, configs):
+    """Add the "append_dimensions" filed for the CloudWatch Agent Metrics section."""
+    if "append_dimensions" in configs.keys() and len(metric_configs["metrics_collected"]) > 0:
```
chenwany: nit: we can remove `.keys()` here

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -177,14 +178,68 @@ def filter_output_fields(configs):
...
+        metric_configs["append_dimensions"] = select_append_dimensions(configs["append_dimensions"])
+    return metric_configs
+
+
+def add_aggregation_dimensions(metric_configs, configs):
+    """Add the "aggregation_dimensions" filed for the CloudWatch Agent Metrics section."""
+    if "aggregation_dimensions" in configs.keys() and len(metric_configs["metrics_collected"]) > 0:
```
chenwany: same comment here for `.keys()`

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
```diff
@@ -620,5 +620,49 @@
...
+        ],
+        "node_roles": [
+          "HeadNode"
+        ]
+      }
+    ],
+    "append_dimensions": [
+      "InstanceId"
+    ],
+    "aggregation_dimensions": [
```
chenwany, line 653: Do we need aggregation_dimensions instance Id and cluster name here? Are they being used in the dashboard and alarm?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config_schema.json
```diff
@@ -0,0 +1,101 @@
...
+              "append_dimensions": {
+                "type": "array",
+                "items": {"type": "string", "enum": ["ClusterName"]}
+              },
+              "node_roles": {
+                "type": "array",
+                "items": {"type": "string", "enum": ["HeadNode", "ComputeFleet"]}
+              }
+            },
+            "required": [
```
davprat, line 75: It looks like we don't have a `feature_conditions` for metrics. Is there a backlog task to add this?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config_util.py
```diff
@@ -1,11 +1,11 @@
 """
-Validate and modify the data in the cloudwatch_log_files.json cookbook file.
+Validate and modify the data in the cloudwatch_agent_config.json cookbook file.
```
davprat, line 2: I like the new name. Just out of curiosity, have any customers customized this file? It looks like for pcluster >= 3.0.0, modifying this file is undocumented (https://github.com/aws/aws-parallelcluster/wiki/CloudWatch-Logs) so we're probably good. @chenwany What do you think?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
davprat: If we don't user `ClusterName` in the dashboard, what happens if a user creates two clusters in the same account? Will the dashboards pickup metrics from the other cluster or do we have another way of keeping them separate?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
davprat: Also too, I think you don't need `and len(metric_configs["metrics_collected"]) > 0:` as just `and metric_configs["metrics_collected"]:` should work I think?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
chenwany: Instance id or headnode private ip are used for metrics and alarm. We can use that to keep them separate

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config_util.py
chenwany: I think there are some users customized logfile, TBH I don't know if they are using `cloudwatch_log_files.json`. I think `cloudwatch_log_files.json` is not documented and is not exposed to our customers.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
delongmeng: as discussed, we will remove them for now

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config_schema.json
delongmeng: For now we don't need feature_conditions for the current metrics, and we can add that section when needed.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config_util.py
delongmeng: As discussed, we think it makes sense to use the new file name as it's more accurate. This document should not need to be customized by users.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
delongmeng: Good point, thanks!

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
delongmeng: I will remove ClusterName from the json configuration file and keep it in the schema and this python script, so that in the future we can still use it for metrics if needed. 
@davprat Right now, we have `"append_dimensions": ["InstanceId"]` so different clusters can be differentiated by their instance ids.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
delongmeng: sure

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
delongmeng: Yes that should work as well. Will do so.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
delongmeng: code has been refactored as you suggested @chenwany

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
delongmeng: Tried remove the aggregation dimensions. Memory metric was still fine but the graph for disk metric was broken. The reason is in Graphs and Alarms we use Instance Id as the dimension for these metrics. With aggregation dimension of Instance Id, both memory and disk metrics are fine. However, without it, disk metric has other dimensions beside Instance Id, such as device, file system, and path; memory metric doesn't have other dimensions so it was fine. The easiest way to fix it is to still use Instance Id as aggregation dimension. We don't need ClusterName so it can be safely removed.

chenwany: When using aggregation metric if InstanceId for disk metric, does it only include the value of “/” volume? Did you test if with adding "/" and "instance_id" without aggregation metrics in cloudwatch dashboard enough for monitoring disk usage for root volume?

chenwany (approved):
LGTM!

PR #489, opened 2020-01-13
FSx Lustre on Ubuntu 16/18

sean-smith:
Signed-off-by: Sean Smith <seaam@amazon.com>

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1740, opened 2023-02-08
[bandit] Validate args before subprocess calls

EddyMM:
### Description of changes
*  Validate device name arguments against a regex pattern before using it in subprocess call
   * N/B: The `manageVolume` and `ec2_dev_2_volid ` scripts are created in different directories (`/usr/local/sbin` and `/sbin` respectively) hence the use of their own independent validations (until we find a way to share functions, constants, modules between Python file in the cookbook)
* Add comments on why security exclusions remain valid (e.g. some subprocess calls are safe since they do not use any external arguments)

### Tests
* Unit tests
* Running existing Integrations tests

### References
* [[3.x] Reducing risks associated to Bandit findings #489](https://github.com/aws/aws-parallelcluster-node/pull/489)

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
+++ b/cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
@@ -1,5 +1,6 @@
 import configparser
 import os
+import re
 import sys
 import syslog
 import time
@@ -24,11 +25,28 @@ def get_imdsv2_token():
     return headers
 
 
+def validate_device_name(device_name):
+    """
+    Validate an argument used to build a subprocess command against a regex pattern.
+
+    The validation is done after forcing the encoding to be the standard Python Unicode / UTF-8
+    :param device_name: an argument string to validate
+    :raise: Exception if the argument fails to match the patter
+    :return: True if the argument matches the pattern
+    """
+    device_name = (str(device_name).encode("utf-8", "ignore")).decode()
+    match = re.match(r"^(\w)+$", device_name)
+    if not match:
+        raise ValueError("Device name provided argument has an invalid pattern.")
+    return True
+
+
 def main():
     syslog.syslog("Starting ec2_dev_2_volid.py script")
     # Get dev
     try:
         dev = str(sys.argv[1])
+        validate_device_name(dev)
         syslog.syslog(f"Input block device is {dev}")
     except IndexError:
         syslog.syslog(syslog.LOG_ERR, "Provide block device i.e. xvdf")
--- a/cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
+++ b/cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
@@ -4,6 +4,7 @@
 import argparse
 import configparser
 import os
+import re
 import subprocess  # nosec B404
 import sys
 import time
@@ -15,6 +16,22 @@ from botocore.config import Config
 METADATA_REQUEST_TIMEOUT = 60
 
 
+def validate_device_name(device_name):
+    """
+    Validate an argument used to build a subprocess command against a regex pattern.
+
+    The validation is done after forcing the encoding to be the standard Python Unicode / UTF-8
+    :param device_name: an argument string to validate
+    :raise: Exception if the argument fails to match the patter
+    :return: True if the argument matches the pattern
+    """
+    device_name = (str(device_name).encode("utf-8", "ignore")).decode()
+    match = re.match(r"^/dev/(\w)+$", device_name)
+    if not match:
+        raise ValueError("Device name provided argument has an invalid pattern.")
+    return True
+
+
 def convert_dev(dev):
     # Translate the device name as provided by the OS to the one used by EC2
     # FIXME This approach could be broken in some OS variants, see  # pylint: disable=fixme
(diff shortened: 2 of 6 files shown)


cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -24,11 +25,27 @@ def get_imdsv2_token():
...
+    Validate an argument used to build a subprocess command against a regex pattern.
+
+    The validation is done after forcing the encoding to be the standard Python Unicode / UTF-8
+    :param device_name: an argument string to validate
+    :raise: Exception if the argument fails to match the patter
+    :return: True if the argument matches the pattern
+    """
+    match = re.match(r"^(\w){2,}$", device_name)
+    if not match:
+        raise ValueError("Device name provided argument contains has an invalid pattern.")
```
gmarciani: Typo: contains + has

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -24,11 +25,27 @@ def get_imdsv2_token():
...
+def validate_device_name(device_name):
+    """
+    Validate an argument used to build a subprocess command against a regex pattern.
+
+    The validation is done after forcing the encoding to be the standard Python Unicode / UTF-8
+    :param device_name: an argument string to validate
+    :raise: Exception if the argument fails to match the patter
+    :return: True if the argument matches the pattern
+    """
+    match = re.match(r"^(\w){2,}$", device_name)
```
gmarciani: Why` {2,}`?

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
```diff
@@ -15,6 +16,21 @@
 METADATA_REQUEST_TIMEOUT = 60
 
 
+def validate_device_name(device_name):
```
gmarciani, line 19: Why duplicating this logic (an almost equal logic has been defined in `cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py`.

What about putting the logic in common with parameter about the prefix of the accepted pattern?

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -24,11 +25,27 @@ def get_imdsv2_token():
     return headers
 
 
+def validate_device_name(device_name):
```
gmarciani, line 28: Is the name of this file correct? `ec2_dev_2_volid` shouldn't be `valid` instead?

EddyMM: The name is correct `volid` -> Volume ID.
The script returns the volumen ID given the Block Device Name.

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
EddyMM: I agree. At the moment, these scripts are not in the form of a library (or in the same folder at execution time).

From the PR description:
The `manageVolume` and `ec2_dev_2_volid ` scripts are created in different directories (`/usr/local/sbin` and `/sbin` respectively) hence the use of their own independent validations (until we find a standard way to share functions, constants, modules between Python file in the cookbook)

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
EddyMM: The block device names all have at least 2 alphanumeric chars [persistent_block_device_naming](https://wiki.archlinux.org/title/persistent_block_device_naming). For security purposes though, this regex constraint isn't really a hard requirement though. I can remove it.

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
EddyMM: Some of the Python scripts in the cookbook are built as standalone scripts (with little/no code sharing capabilities).
Some suggestions to address this would be placing them in the same directory (which fixes the problem for scripts that can be in the same folder), or having a shared module installed in the virtualenv.

Either way, adding these changes seemed a bit too invasive for a simple fix to address a bandit finding. A cookbook-wide solution will need to be evaluated.

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
gmarciani: Since the goal of this change is security enforcement I would avoid introducing formatting constraints that may be out of scope and introduce risk (e.g. what if the name format changes in future? not highly probably I'd say but still can be avoided)

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/manageVolume.py
gmarciani: Ok, considering the urgency of this Pr and the refactoring that this comment would imply, let's consider this as out of scope for this PR.

cookbooks/aws-parallelcluster-install/files/default/ec2_udev_rules/ec2_dev_2_volid.py
gmarciani: ahahah ok thanks

gmarciani (approved):

PR #1743, opened 2023-02-09
Define new fs_update recipe and add tests for it

enrico-usai:
### Description of changes

The shared_storages_data.yaml file is created starting from an erb template adding information for ebs, raid, efs and fsx storages only when defined.

Added fs_update and fs_update_default_values to test the two cases

### Tests
Tested on RH8 and Ubuntu20 on docker and EC2

### DIFF
--- a/kitchen.recipes.yml
+++ b/kitchen.recipes.yml
@@ -214,6 +214,50 @@ suites:
 #        - recipe:aws-parallelcluster-install::directories
 #        - recipe:aws-parallelcluster-install::python
 #        - recipe:aws-parallelcluster-install::cloudwatch_agent
+  - name: fs_update
+    run_list:
+      - recipe[aws-parallelcluster::add_dependencies]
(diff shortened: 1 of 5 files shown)


ermanno (dismissed):

ermanno (approved):

kitchen.recipes.yml
```diff
@@ -220,6 +220,50 @@ suites:
 #        - recipe:aws-parallelcluster-install::directories
 #        - recipe:aws-parallelcluster-install::python
 #        - recipe:aws-parallelcluster-install::cloudwatch_agent
+  - name: fs_update
+    run_list:
+      - recipe[aws-parallelcluster::add_dependencies]
```
gmarciani, line 225: If I remember correctly, we have a recipe that sets up the envars for Chef.
Shouldn't that recipe be on top of every run_list?

PR #1753, opened 2023-02-10
Transform install_munge recipe into a resource

NSsirena:
### Description of changes
* Transform install_munge recipe into a resource

### Tests
* Successfully tested on all platform apart from redhat8

### Checklist


cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
```diff
@@ -0,0 +1,118 @@
...
+# Munge
+munge_version = '0.5.14'
+munge_url = "https://github.com/dun/munge/archive/munge-#{munge_version}.tar.gz"
+munge_tarball = "#{node['cluster']['sources_dir']}/munge-#{munge_version}.tar.gz"
+munge_user = 'munge'
+munge_user_id = node['cluster']['reserved_base_uid'] + 2
+munge_group = munge_user
+munge_group_id = munge_user_id
+
+action :setup do
```
jdeamicis, line 27: Could we call this action in a more speaking way? Something like `install_and_configure`?

cookbooks/aws-parallelcluster-slurm/recipes/install.rb
```diff
@@ -16,5 +16,5 @@
 # limitations under the License.
 
 include_recipe 'aws-parallelcluster-slurm::install_pmix'
-include_recipe 'aws-parallelcluster-slurm::install_munge'
+munge 'Install munge'
```
jdeamicis: I would explicitly add the performed action for more clarity here, even if it's the default one.

Actually, maybe we shouldn't even have a default action for such a special resource (or maybe downloading the source code, installing and configuring may not be a good default action). I imagine we could use the resource also to restart the munge daemon, or to rotate the munge key, or whatever, so maybe I wouldn't configure the default action of the `munge` resource to redownload and reinstall it.

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
```diff
@@ -0,0 +1,118 @@
...
+  # Create required directories for munge
+  dirs = ["/var/log/munge", "/etc/munge", "/var/run/munge"]
+  dirs.each do |dir|
+    directory dir do
+      action :create
+      owner munge_user
+      group munge_group
+    end
+  end
+end
```
jdeamicis, line 118: I believe here we are missing the [setup_munge_head_node](https://github.com/aws/aws-parallelcluster-cookbook/blob/1a57f16b59968528609ae51765ce98e4405f7d76/libraries/helpers.rb#L409-L422), which I believe functionally belongs to this `munge` resource. Should we move it here and then use it in `config_head_node` as a `generate_munge_key` action of the `munge` resource?

jdeamicis: Actually there are additional operations performed in that helpers.rb.

We can even leave them where they are and move them to the new resource at a later stage if we prefer (but this refactoring seems the best opportunity to me).

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
```diff
@@ -0,0 +1,118 @@
...
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Munge
+munge_version = '0.5.14'
+munge_url = "https://github.com/dun/munge/archive/munge-#{munge_version}.tar.gz"
+munge_tarball = "#{node['cluster']['sources_dir']}/munge-#{munge_version}.tar.gz"
+munge_user = 'munge'
+munge_user_id = node['cluster']['reserved_base_uid'] + 2
```
jdeamicis: What about defining a `reserved_uid_offset_munge` parameter equal to 2 in the global `default.rb` and then referencing it here?

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
NSsirena: These are part of another task and will be addressed in another PR.

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
NSsirena: Would make sense,
however after working at the install_slurm PR I discovered that we have also other user/groups related to Slurm.
So I'd prefer to define a `base_slurm_user_id` (and possibly a reserved interval of ids) and then define the slurm user, munge user, and so on at given offsets.

I didn't do it yet because I still don't have a full picture of how many users are required.

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
NSsirena: I agree that `setup` is a very broad term that can be replaced with a more descriptive name, however as you stated in the following comment is not yet clear `what the default action should be` so I'll keep this name since it is the `standard` we agreed upon for the refactoring and leave the "optimizations" to a later time when the usage requirement will be more clear.

cookbooks/aws-parallelcluster-slurm/recipes/install.rb
NSsirena: I'll revisit the code to explicitly state the name of the action if this makes it more clear.

However, even if I'd like to review and improve what the default action would be, and how to group some basic actions into high-level actions that would be a better fit in our workflows, .. this is beyond the size and scope of this task so I'll create a NiceToHave task for addressing these request.

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
jdeamicis: You can gain a full picture by looking at the number of UUID variables in the big old `default.rb`. For Slurm we should need 2 users, and around 10 UUIDs or less in total.

cookbooks/aws-parallelcluster-slurm/resources/munge/partial/_munge_actions.rb
jdeamicis: Resolved for now.

test/resources/controls/aws_parallelcluster_slurm/munge_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    its('group') { should eq 'root' }
+    its('content') do
+      should match("USER=#{munge_user}")
+      should match("GROUP=#{munge_group}")
+    end
+  end unless os_properties.redhat_ubi?
+end
+
+control 'munge_folders_created' do
+  title 'Munge folder have been created'
```
jdeamicis, line 54: folders

jdeamicis (approved):
Approved with minor comment.

PR #1755, opened 2023-02-10
Add test for stunnel installation

ermanno:
### Description of changes
Add a resource for `stunnel` installation.

### Tests
Tested on Docker and EC2. 

Notes:
- **_AL2_** is excluded because there `stunnel` is installed via the package manager as a dependency of `amazon-efs-utils`.
- _**RHEL8 on Docker**_ is excluded because there we don't perform the installation of the base packages, as they are not supported on UBIs, so we cannot build `stunnel`.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-install/attributes/default_redhat8.rb
+++ b/cookbooks/aws-parallelcluster-install/attributes/default_redhat8.rb
@@ -10,7 +10,7 @@ default['cluster']['base_packages'] = %w(vim ksh tcsh zsh openssl-devel ncurses-
                                              libical-devel postgresql-devel postgresql-server sendmail libxml2-devel libglvnd-devel
                                              mdadm python2 python2-pip libgcrypt-devel libevent-devel glibc-static bind-utils
                                              iproute NetworkManager-config-routing-rules python3 python3-pip iptables libcurl-devel yum-plugin-versionlock
-                                             coreutils moreutils sssd sssd-tools sssd-ldap curl environment-modules)
+                                             coreutils moreutils sssd sssd-tools sssd-ldap curl environment-modules gcc gcc-c++)
 
 # Needed by hwloc-devel blas-devel libedit-devel and glibc-static packages
 default['cluster']['extra_repos'] = 'codeready-builder-for-rhel-8-rhui-rpms'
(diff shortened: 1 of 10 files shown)


cookbooks/aws-parallelcluster-install/attributes/default_redhat8.rb
```diff
@@ -10,7 +10,7 @@
                                              libical-devel postgresql-devel postgresql-server sendmail libxml2-devel libglvnd-devel
                                              mdadm python2 python2-pip libgcrypt-devel libevent-devel glibc-static bind-utils
                                              iproute NetworkManager-config-routing-rules python3 python3-pip iptables libcurl-devel yum-plugin-versionlock
-                                             coreutils moreutils sssd sssd-tools sssd-ldap curl environment-modules)
+                                             coreutils moreutils sssd sssd-tools sssd-ldap curl environment-modules gcc gcc-c++)
```
ermanno, line 13: @awslitvit @francesco-giordano I'm highlighting this since we discussed it off-line.

We discussed the idea of installing the compiler for RHEL8 as a `stunnel` dependency (differently from what we do, for example, [in the case of AL2](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-install/attributes/default_amazon2.rb#L10)) and skip the installation of the _base packages_ when we execute the test.

However I then saw that both for RHEL8 and for all the other OSs we also need the _base packages_ in order to successfully compile `stunnel`, so we cannot skip this phase, thus I added these packages here, like for AL2.

cookbooks/aws-parallelcluster-install/resources/stunnel/stunnel_centos7.rb
```diff
@@ -0,0 +1,27 @@
...
+unified_mode true
+default_action :setup
+
+use 'partial/_setup'
+
+action_class do
+  def dependencies
+    %w(tcp_wrappers-devel)
+  end
+end
```
francesco-giordano, line 27: **Thought:** what about move this package in the list of the *base packages*?

Alternatively remove completely this dependency to have all the OSes aligned? 
Since this feature is not used by `efs-utils` and I do not think that this would be an breaking change since the software should be used only by `efs-utils`, I would keep the installation lighter removing the package.

ermanno: - If they are not needed as base dependencies, I would keep them in the resource/recipes where they are used: this is the approach we have taken for all such dependencies that we use only when installing specific software.
- I would keep the `tcp_wrappers-devel` because it's listed as dependency in the official documentation, despite it maybe not being strictly necessary... I would make only the changes needed to support RHEL8 and not diverge from the official documentation when not strictly needed.

francesco-giordano (approved):

PR #1757, opened 2023-02-10
Refactor install_slurm recipe

NSsirena:
### Description of changes
* Explicitly list slurm dependencies
* move version definition in slurm cookbook
* make use of the newly defined munge resource
* small code clean up 
* Add tests for install_slurm recipe

### Tests
* Tests validated on all platforms on EC2.

### Checklist


awslitvit:
I would also change the PR description, as you did much more than just adding tests.

cookbooks/aws-parallelcluster-slurm/attributes/slurm_attributes.rb
```diff
@@ -0,0 +1,7 @@
+# Slurm attributed shared between install_slurm and configure_slurm_accounting
+default['cluster']['reserved_base_uid'] = 400
```
jdeamicis: This should be global. Please remove it and leave it under the global defaults.

cookbooks/aws-parallelcluster-slurm/attributes/slurm_attributes.rb
```diff
@@ -0,0 +1,7 @@
...
+default['cluster']['slurm']['user'] = 'slurm'
+default['cluster']['slurm']['user_id'] = node['cluster']['reserved_base_uid'] + 1
```
jdeamicis: Same comment as [this one for the munge recipes](https://github.com/aws/aws-parallelcluster-cookbook/pull/1753#discussion_r1104317554). Could we please have the uid offsets managed in a central location under the global defaults or whatever other location we will be using for global parameters?

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
```diff
@@ -17,46 +17,59 @@
...
+slurm_version = '22-05-8-1'
+slurm_commit = ''
+slurm_tar_name = if slurm_commit.empty?
+                   "slurm-#{slurm_version}"
+                 else
+                   "#{slurm_commit}"
+                 end
+slurm_tarball = "#{node['cluster']['sources_dir']}/#{slurm_tar_name}.tar.gz"
+slurm_url = "https://github.com/SchedMD/slurm/archive/#{slurm_tar_name}.tar.gz"
+slurm_sha256 = '8c8f6a26a5d51e6c63773f2e02653eb724540ee8b360125c8d7732314ce737d6'
```
jdeamicis: Why are these here and not in `attributes/slurm_attributes.rb`? Surely at least the version, commit and sha256 variables must be exposed in the attributes file. The other ones are mostly internal logic variables and can be part of the internal of the recipe/resource.

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
```diff
@@ -17,46 +17,59 @@
...
 remote_file slurm_tarball do
-  source node['cluster']['slurm']['url']
+  source slurm_url
   mode '0644'
   retries 3
   retry_delay 5
-  not_if { ::File.exist?(slurm_tarball) }
-end
+  action :create_if_missing
+end unless redhat_ubi?
```
jdeamicis: Why would this be different for redhat? Or am I missing something about `redhat_ubi?`? This applies also to the other steps below.

kitchen.recipes.yml
```diff
@@ -397,3 +397,22 @@ suites:
...
+        - slurm_library_shared
+    attributes:
+      dependencies:
+        - recipe:aws-parallelcluster::test_dummy
+        - recipe:aws-parallelcluster-install::directories
+        - resource:package_repos
+        - resource:install_packages
+        - recipe:aws-parallelcluster-install::python
+        - recipe:aws-parallelcluster-slurm::install_jwt
+        - recipe:aws-parallelcluster-slurm::install_pmix
```
jdeamicis, line 422: munge is also required for the installation of Slurm. Please add it to the dependencies.

kitchen.recipes.yml
```diff
@@ -397,3 +397,22 @@ suites:
...
+  - name: slurm_installed
+    run_list:
+      - recipe[aws-parallelcluster::add_dependencies]
+      - recipe[aws-parallelcluster-slurm::install_slurm]
+    verifier:
+      controls:
+        - slurm_installed
+        - slurm_user_and_group_created
+        - slurm_licence_configured
+        - slurm_library_shared
```
jdeamicis, line 413: This is probably an extra task, but we may want to have a further test that parses the `config.log` produced by the the Slurm configure step to check that Slurm is built with all the functionalities we expect to be there.

If we don't want to address this right now, please create a task for it.

test/recipes/controls/aws_parallelcluster_slurm/install_slurm_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+slurm_user = 'slurm'
+slurm_group = slurm_user
+slurm_license_path = '/opt/parallelcluster/licenses/slurm'
+
+control 'slurm_installed' do
```
jdeamicis, line 17: We may want to expand this:
- we should be able to find the most common binaries under `/opt/slurm/bin` (e.g. `srun`,`sbatch`, `scontrol`, etc.);
- we may search for specific libraries we expect to be there (for example `accounting_storage_mysql.so` to check that the mysql plugin has been compiled);
- later on we could think of more stuff.

jdeamicis (commented):
Looks good to me, I left some comments around the code for improvements/clarifications.

cookbooks/aws-parallelcluster-slurm/attributes/slurm_attributes.rb
NSsirena: Yep. 
Good catch .. I moved here for a test and forgot to remove it.
Fixing it.

cookbooks/aws-parallelcluster-slurm/attributes/slurm_attributes.rb
NSsirena: Same as above ..fixing.

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
NSsirena: These aren't used elsewhere so I moved them in the smallest required scope.
I can move them back in the attributes section if you find it more readable.

cookbooks/aws-parallelcluster-slurm/recipes/install_slurm.rb
NSsirena: Yep.. unfortunately  the repositories in `redhat_ubi` don't have all the required packages.
So we agreed for this first round to skip redhat tests on docker and address this issue at a later time.

kitchen.recipes.yml
NSsirena: I will add it for completeness, however I will need to rebase this after we merge the change that transforms the munge recipe into a resource

kitchen.recipes.yml
NSsirena: Yes. Improving the current recipes is beyond the scope of the task...but, I agree we should add all the required checks. I'll create a task for this.

Note: config.log will be present only after the configuration phase, that still hasn't been addressed.

test/recipes/controls/aws_parallelcluster_slurm/install_slurm_spec.rb
NSsirena: I only partially agree.
I do not think we should test for the presence of the binaries, we are building the code from the sources ...so this check would actually test the `slurm makefile` and install process ..that is outside our scope.

I however agree on testing for the presence of the expected libraries since this may depend on how we configure the environment while building slurm, so something may be missing if we have some bug or regression.

I'll add this section on the NiceToHave task.

cookbooks/aws-parallelcluster-slurm/attributes/slurm_attributes.rb
```diff
@@ -0,0 +1,4 @@
+# Slurm attributes shared between install_slurm and configure_slurm_accounting
+default['cluster']['slurm']['commit'] = ''
```
eantonin, line 2: question, non-blocking, what's the difference between commit and sha? are both needed?

jdeamicis: The sha is the checksum of the tarball we download from GitHub, not the commit hash.

NSsirena: To add a few cent, the commit can be used to pick a very specific slurm version instead of an officially released one. Like for example to test a patch or a feature that is still in development.

eantonin (approved):
LGTM

PR #1760, opened 2023-02-11
[Fix] Multi NIC IP Alignment

Nbot07A:
### Description of changes
* set ip to NIC 0 Device 0 primary ip to ensure IP alignment with Multi NIC instances

### Tests
* Passed new unit test test_primary_ip

### Checklist


cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
```diff
@@ -96,7 +96,21 @@
...
+    uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/device-number")
+    device_number = get_metadata_with_token(token, uri)
+    uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/network-card")
+    network_card = get_metadata_with_token(token, uri)
+    next unless device_number == '0' && network_card == '0'
+
+    uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/local-ipv4s")
+    primary_ip = get_metadata_with_token(token, uri)
+    break
+  end
```
francesco-giordano: **Suggestion:** What about move this part in an helper function? In this way also `test_primary_ip.rb` can reuse it.

cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
```diff
@@ -96,7 +96,21 @@
...
+  pattern "^#{primary_ip}\s+"
```
francesco-giordano: **Thought:** what happens if the ip already present in the `/etc/hosts` is not what we identify as `primary_ip`. I think that we should clear each line where we find an IP associated with the current instance

cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
Nbot07A: I considered that while implementing but decided against it because `test_primary_ip.rb` helps to ensure that the code is correct by not reusing the functions within it.

cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
Nbot07A: I'm not sure that I completely understand your question. If the IP is already correct then there wouldn't be any issue. So, what do you mean by "if the ip is already present?"

cookbooks/aws-parallelcluster-test/recipes/test_primary_ip.rb
```diff
@@ -0,0 +1,46 @@
...
+        for word in $IP_HOSTS
+        do
+          if [[ ${word} == ${mac_ip} ]]
+          then
+            exit 0
+          fi
+        done
+      fi
+    done
+    echo Error: Route53 IP does not match host IP
```
francesco-giordano: **Nitpick (non-blocking):** In this case Route53 is not checked. I would change the line with something like
```
echo "Primary ip ${mac_ip} not in file /etc/hosts"
echo "cat /etc/hosts"
cat /etc/hosts
exit 1
```
So we can have more log if it fails

libraries/helpers.rb
```diff
@@ -535,6 +535,7 @@ def are_mount_or_unmount_required?
   change_set = JSON.load_file("#{node['cluster']['shared_dir']}/change-set.json")
   change_set["changeSet"].each do |change|
     next unless change["updatePolicy"] == "SHARED_STORAGE_UPDATE_POLICY"
+
```
francesco-giordano: Typo

cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
```diff
@@ -93,10 +93,18 @@
...
+  end
+end
+
 # Configure fqdn in /etc/hosts
-replace_or_add "set fqdn in the /etc/hosts" do
+append_if_no_line "Append primary ip to /etc/hosts" do
   path "/etc/hosts"
-  pattern "^#{node['ec2']['local_ipv4']}\s+"
-  line(lazy { "#{node['ec2']['local_ipv4']} #{node['cluster']['assigned_hostname'].chomp('.')} #{node['cluster']['assigned_short_hostname']}" })
+  primary_ip = get_primary_ip
```
francesco-giordano: **Suggestion**: Can you use the get_primary_ip directly in the line 108?

libraries/helpers.rb
```diff
@@ -575,3 +575,30 @@ def format_directory(dir)
...
+      device_number = get_metadata_with_token(token, uri)
+      uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/network-card")
+      network_card = get_metadata_with_token(token, uri)
+      next unless device_number == '0' && network_card == '0'
+
+      uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/local-ipv4s")
+      primary_ip = get_metadata_with_token(token, uri)
+      break
+    end
+  end
```
francesco-giordano, line 602: **Nitpick (non-blocking):** can you open a task to update this logic as `Delete all local ips in /etc/hosts` to not add calls to the metadata service.

francesco-giordano (approved):

libraries/helpers.rb
Nbot07A: Sure

libraries/helpers.rb
```diff
@@ -575,3 +575,30 @@ def format_directory(dir)
...
+  primary_ip = node['ec2']['local_ipv4'].to_s
+
+  token = get_metadata_token
+  macs = network_interface_macs(token)
+  if macs.length > 1
+    macs.each do |mac|
+      uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/device-number")
+      device_number = get_metadata_with_token(token, uri)
+      uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/network-card")
+      network_card = get_metadata_with_token(token, uri)
```
gmarciani, line 595: Minor: let's avoid URI duplication.duplications

cookbooks/aws-parallelcluster-test/recipes/test_primary_ip.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-test
+# Recipe:: test_primary_ip
+#
+# Copyright:: 2013-2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani, line 7: Need to change the copyright year

gmarciani:
Closed in favour of https://github.com/aws/aws-parallelcluster-cookbook/pull/1971

PR #1761, opened 2023-02-13
[Isolated Regions] Add support for US isolated regions: us-iso-* and us-isob-*.

gmarciani:
### Changes
Add support for US isolated regions: **us-iso-*** and **us-isob-***.

In particular:

1. Added support for the AWS domain in these regions.
1. Added script `/opt/parallelcluster/scripts/patch-iso-instance.sh` that is expected to be used at run-time by both head and compute nodes to configure the instance in these regions.
1. Added environment variable `AWS_CA_BUNDLE` to the environment loaded by supervisord in all the daemons when running in US isolated regions.
1. Added `/etc/sudoers.d/99-parallelcluster-env-keep` to preserve the environment variable `AWS_CA_BUNDLE` when switching user in non interactive sessions. This is done at config time only when running in US isolated regions.
1. Added `--verbose` option to the execution of cfn-hup with supervisord to make it easier to troubleshoot cfn-hup issues.
1. Added support for these regions to the scripts used to install Cinc.
1. Added a comment to `manage_fsx` recipe to explain why the file system domain must not be adapted to support these regions.

### Tests
Manual tests **(US ISO TESTS ARE STILL ONGOING)**:
1. AMI build (executed in Commercial)
2. Cluster creation (executed both in Commercial and US Isolated)
3. Cluster update with dynamic file system mounting (executed both in Commercial and US Isolated)
4. Cluster update with compute nodes min-count change (executed both in Commercial and US Isolated)
5. Submission of job that requires compute fleet autoscaling (executed both in Commercial and US Isolated)

A full regression tests will be done on the authoritative pipeline once the change will be merged.

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/base.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/base.rb
@@ -83,11 +83,16 @@ template '/etc/sudoers.d/99-parallelcluster-user-tty' do
 end
 
 # Install parallelcluster specific supervisord config
+region = node['cluster']['region']
 template '/etc/parallelcluster/parallelcluster_supervisord.conf' do
   source 'base/parallelcluster_supervisord.conf.erb'
   owner 'root'
   group 'root'
   mode '0644'
+  variables(
+    region: region,
+    aws_ca_bundle: region.start_with?('us-iso') ? "/etc/pki/#{region}/certs/ca-bundle.pem" : ''
+  )
 end
 
 # Mount EFS, FSx
--- a/cookbooks/aws-parallelcluster-config/templates/default/base/parallelcluster_supervisord.conf.erb
+++ b/cookbooks/aws-parallelcluster-config/templates/default/base/parallelcluster_supervisord.conf.erb
@@ -4,16 +4,19 @@
 <%# HeadNode -%>
 <% when 'HeadNode' -%>
 [program:cfn-hup]
-command = bash -c "[ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh; <%= node['cluster']['cfn_bootstrap_virtualenv_path'] %>/bin/cfn-hup"
+command = bash -c "[ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh; <%= node['cluster']['cfn_bootstrap_virtualenv_path'] %>/bin/cfn-hup --verbose"
 # The following are needed because cfn-hup starts as a daemon
 exitcodes = 0
 autorestart = unexpected
 startsecs = 0
+<% if @region.start_with?('us-iso') -%>
+environment = AWS_CA_BUNDLE="<%= @aws_ca_bundle %>"
+<% end -%>
 <% if node['cluster']['scheduler'] == 'slurm' -%>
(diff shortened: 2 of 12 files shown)


cookbooks/aws-parallelcluster-config/recipes/base.rb
```diff
@@ -83,11 +83,16 @@
...
 
 # Install parallelcluster specific supervisord config
+region = node['cluster']['region']
 template '/etc/parallelcluster/parallelcluster_supervisord.conf' do
   source 'base/parallelcluster_supervisord.conf.erb'
   owner 'root'
   group 'root'
   mode '0644'
+  variables(
+    region: region,
```
lukeseawalker, line 93: why setting region? it should be avail through node['cluster']['region'] inside the .erb file

cookbooks/aws-parallelcluster-config/templates/default/base/parallelcluster_supervisord.conf.erb
```diff
@@ -4,24 +4,27 @@
 <%# HeadNode -%>
 <% when 'HeadNode' -%>
 [program:cfn-hup]
-command = bash -c "[ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh; <%= node['cluster']['cfn_bootstrap_virtualenv_path'] %>/bin/cfn-hup"
+command = bash -c "[ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh; <%= node['cluster']['cfn_bootstrap_virtualenv_path'] %>/bin/cfn-hup --verbose"
 # The following are needed because cfn-hup starts as a daemon
 exitcodes = 0
 autorestart = unexpected
 startsecs = 0
+<% if @region.start_with?('us-iso') -%>
```
lukeseawalker, line 12: why not using `node['cluster']['region']` for `region`?

cookbooks/aws-parallelcluster-config/recipes/base.rb
gmarciani: This is true. I did this way to mainly to avoid duplication of node['cluster']['region'] within the ERB file and making it more readable. You think it's better to avoid this?

cookbooks/aws-parallelcluster-config/templates/default/base/parallelcluster_supervisord.conf.erb
gmarciani: I preferred that way to improve readability of the ERB file.

cookbooks/aws-parallelcluster-config/recipes/base.rb
gmarciani: As an alternative I think we may declare variables directly within the ERB file like this

```
<% @region = node['cluster']['region'] %>
```

gmarciani: You think it's preferable?

gmarciani: Talked with Luca about this. Both ways are ok.
Will keep the current approach.

lukeseawalker (approved):

PR #1762, opened 2023-02-13
Move efa resource into common package and add configure action 

enrico-usai:
### Description of changes
* Move efa resource into common package to be used by both install and configure recipes and add configure action
* Add test for configure action of efa resource

### Tests
* Tested efa resource to verify movement is working as expected, on all OSes on docker.
* Tested efa-configure on all OSes on EC2. It is skipped except for ubuntu:
``` 
  ✔  efa_system_settings_configured: Check system is correctly configured for EFA
     ✔  Kernel Parameter kernel.yama.ptrace_scope value is expected to eq 0
```


kitchen.resources.yml
```diff
@@ -119,3 +119,14 @@ suites:
       resource: efa
```
awslitvit: Nitpick. I would specify `setup` for the sake of clarity.

awslitvit (dismissed):

awslitvit (approved):

PR #1766, opened 2023-02-14
Add test for ephemeral drives setup

enrico-usai:
### Description of changes
The setup-ephemeral-drives.sh script requires the cfn_ephemeral_dir variable in the cfnconfig file, I added the cfnconfig_mock.rb to create the required file.

Ephemeral drives test verifies the /scratch directory (from cfnconfig) is correctly mounted in an instance type m5d.xlarge with 1 instance store.

lvm2 is already installed as part of the install_packages but it is required to have the ephemeral_drives.rb recipe self contained. The name of the package is the same in all the OSes.

### Tests
* Tested on EC2 on all the OSes.
* instance_type variable is skipped when testing on docker.

### DIFF
--- a/kitchen.recipes.yml
+++ b/kitchen.recipes.yml
@@ -328,27 +328,29 @@ suites:
             vpc_ipv4_cidr_blocks: |
               cidr1
               cidr2
-  - name: ephemeral_drives
-    #driver:
-    #  instance_type: m5d.xlarge  # instance type with ephemeral drives
+  - name: ephemeral_drives_mounted
+    driver:
+      instance_type: m5d.xlarge  # instance type with ephemeral drives
     run_list:
       - recipe[aws-parallelcluster::add_dependencies]
       - recipe[aws-parallelcluster-config::ephemeral_drives]
     verifier:
       controls:
-        - ephemeral_drives_service
+        - ephemeral_drives_service_running
+        - ephemeral_drives_configured
     attributes:
       dependencies:
         - recipe:aws-parallelcluster-install::ephemeral_drives
+        - recipe:aws-parallelcluster-test::cfnconfig_mock
       cluster:
         ebs_shared_dirs: test1,test2
         efs_shared_dirs: ''
         fsx_shared_dirs: ''
         raid_shared_dir: ''
-        ephemeral_dir: test3
+        ephemeral_dir: /scratch
   - name: ephemeral_drives_skipped
-    #driver:
-    #  instance_type: m5d.xlarge  # instance type with ephemeral drives
+    driver:
+      instance_type: m5d.xlarge  # instance type with ephemeral drives
     run_list:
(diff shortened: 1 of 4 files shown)


kitchen.recipes.yml
```diff
@@ -329,26 +329,28 @@ suites:
               cidr1
               cidr2
   - name: ephemeral_drives
```
awslitvit: Nitpick: would it make sense to call it `ephemeral_drives_mounted`, as the opposite of `ephemeral_drives_skipped`?

kitchen.recipes.yml
```diff
@@ -329,26 +329,28 @@ suites:
...
-    #driver:
-    #  instance_type: m5d.xlarge  # instance type with ephemeral drives
+    driver:
+      instance_type: m5d.xlarge  # instance type with ephemeral drives
     run_list:
       - recipe[aws-parallelcluster::add_dependencies]
       - recipe[aws-parallelcluster-config::ephemeral_drives]
     verifier:
       controls:
         - ephemeral_drives_service
```
awslitvit: Nitpick: should be a past participle, as the line below.

test/recipes/controls/aws_parallelcluster_config/ephemeral_drives_spec.rb
```diff
@@ -18,8 +18,19 @@
...
 
-  # TODO: add test to see drivers are mounted
+  only_if { !os_properties.virtualized? }
+
+  # This value is set by cfnconfig_mock.rb
+  describe directory('/scratch') do
+    it { should exist }
+    it { should be_writable }
+    it { should be_mounted }
+  end
```
awslitvit, line 33: Thought. Do you think we should try to write a file in this dir?

awslitvit (dismissed):

test/recipes/controls/aws_parallelcluster_config/ephemeral_drives_spec.rb
enrico-usai: I think we can avoid to create a file and rely on the `be_writable` flag.

kitchen.recipes.yml
enrico-usai: changed, thanks

kitchen.recipes.yml
enrico-usai: renamed, thanks

test/recipes/controls/aws_parallelcluster_config/ephemeral_drives_spec.rb
awslitvit: Not sure [it checks](https://app.asana.com/0/1203019340940586/1203391610890408/f) a file can be actually written in the directory. Looks rather like posix write permission check (even if it is not clear against which user).

awslitvit (approved):

PR #1770, opened 2023-02-15
Kitchen tests dokken platform support fixed to amd64

psacc:
### Description of changes
Execution of kitchen tests with dokken assumes that the only execution platform supported is currently amd64. This configuration should be specified in order to avoid incurring in dokken open issue https://github.com/test-kitchen/kitchen-dokken/issues/288

### Tests
Tested running locally with an arm64 macbook with apple silicon some sample recipes. Previously all were failing.

### References
dokken issue: https://github.com/test-kitchen/kitchen-dokken/issues/288


kitchen.docker.yml
```diff
@@ -1,6 +1,7 @@
 ---
 driver:
   name: dokken
+  platform: linux/amd64
```
ermanno, line 4: We shouldn't set this as default, as it's not the most common case. You can use the `<%= %>` syntax to substitute a value (architecture can be seen programmatically by `uname -m`). You can take a look at `kitchen.ec2.yml`.

ermanno (changes requested):

kitchen.docker.yml
psacc: no other platform can currently be supported with dokken, `chef_image: cincproject/cinc` only exists for `amd64` and is there already set an env that is `amd64` specific: `KERNEL_RELEASE=3.10.0-1160.42.2.el7.x86_64`

ermanno: Yes, sorry, of course you are correct, I had mistakenly read `arm64` there. I am okay with this change.

ermanno (approved):

PR #1775, opened 2023-02-16
Add IMDS test

ermanno:
### Description of changes
Add IMDS test.

### Tests
Run test on EC2 (not on Docker as it requires IMDS).

### Checklist


kitchen.recipes.yml
```diff
@@ -399,3 +399,20 @@ suites:
...
+      - recipe[aws-parallelcluster::add_dependencies]
+      - recipe[aws-parallelcluster-config::imds]
+    verifier:
+      controls:
+        - imds_configured
+    attributes:
+      dependencies:
+        - recipe:aws-parallelcluster-install::directories
+        - resource:package_repos
+        - resource:install_packages
```
enrico-usai, line 414: why do you need the entire install_packages repo? Do you know which are the required packages?

ermanno: Yes, the `iptables` package.

enrico-usai: if you need iptables, this has been added as part of IMDS feature so I think we can move outside of base_package: https://github.com/aws/aws-parallelcluster-cookbook/commit/17ee7fbd123c0e32e7b06904bc252a39cdb3ff90

ermanno: From offline discussion: we will leave it the `base_packages`, with a comment explaining that it is used for IMDS.

We can't install it in `imds.rb` because it would be installed at runtime when/where it is needed (i.e. on the _**head node**_ if we are not using `awsbatch`), but the node might be in a subnet that is not connected to the internet.

This is a problem we don't have, for example, when installing `dcv-gl` at runtime, because in that case the node has to be connected to the internet, as it is required to connect via the DCV web client.

cookbooks/aws-parallelcluster-config/recipes/imds.rb
```diff
@@ -15,7 +15,7 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-return if node['cluster']['scheduler'] == 'awsbatch'
+return if virtualized? || node['cluster']['scheduler'] == 'awsbatch'
```
enrico-usai, line 18: With this new check we're skipping the execution of this code in the systemtests too, instead before they were executed. This let me think we can mock something and execute it in docker too.

enrico-usai: I found the point. We were skipping it in the `init.rb`

```
# IMDS
include_recipe 'aws-parallelcluster-config::imds' unless virtualized?
```

We can remove the `unless` check from init.rb and keep this one you added.

enrico-usai (dismissed):

enrico-usai (approved):

PR #1772, opened 2023-02-15
Refactor efs recipe in a resource

francesco-giordano:
### Description of changes
* [Add env variables KITCHEN_INSTANCE_TYPE](https://github.com/aws/aws-parallelcluster-cookbook/commit/324d56082b4728eca434ba66d61710f1582ef550)
* [Refactor efs recipe in a resource](https://github.com/aws/aws-parallelcluster-cookbook/commit/fba9c4c4eaa81b351b0e21c2095b47ef1e1c854f)

### Tests
* Locally tested on EC2 and Docker

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1778, opened 2023-02-16
Refactor and test install_slurm recipe

awslitvit:
### Description of changes
Refactor and test install_slurm recipe

### Tests
Tested on docker and on EC2 (all supported OSes)

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/recipes/install_pmix.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/install_pmix.rb
@@ -26,15 +26,8 @@ remote_file pmix_tarball do
   mode '0644'
   retries 3
   retry_delay 5
-  not_if { ::File.exist?(pmix_tarball) }
-end
-
-ruby_block "Validate PMIx Tarball Checksum" do
-  block do
-    require 'digest'
-    checksum = Digest::SHA256.file(pmix_tarball).hexdigest
-    raise "Downloaded Tarball Checksum #{checksum} does not match expected checksum #{pmix_sha256}" if checksum != pmix_sha256
-  end
+  checksum pmix_sha256
+  action :create_if_missing
 end
 
 bash 'Install PMIx' do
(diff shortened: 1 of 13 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/install_pmix.rb
```diff
@@ -26,15 +26,8 @@
...
-  not_if { ::File.exist?(pmix_tarball) }
-end
-
-ruby_block "Validate PMIx Tarball Checksum" do
-  block do
-    require 'digest'
-    checksum = Digest::SHA256.file(pmix_tarball).hexdigest
-    raise "Downloaded Tarball Checksum #{checksum} does not match expected checksum #{pmix_sha256}" if checksum != pmix_sha256
-  end
+  checksum pmix_sha256
```
demartinofra, line 29: From [chef-docs](https://docs.chef.io/resources/remote_file/) it looks like checksum is used to prevent a file from being re-downloaded, when the local file matches the checksum.

All the checksums present in our cookbook have a different goal which is to verify the integrity of the downloaded artifact and prevent the potential download of compromised software. Does the new checksum mechanism achieves the same result?

NSsirena: According to this comment in the code it seems so

https://github.com/chef/chef/blob/main/lib/chef/resource/remote_file.rb#L45-L63

If the file doesn't exist it is downloaded and the checksum, validated.
If the file exists and the checksum match it is kept and a new download avoided.

awslitvit: Chef documentation is not clear: they don't say the checksum is used *only* to that end. In fact, the second example on the page you linked says "Download a file with a checksum to validate". This is why we tested the behavior [in a previous PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/1772#discussion_r1108657687) before changing the code.

PR #1781, opened 2023-02-17
Fix Configure step of SystemTests

enrico-usai:
### Add bump version to cookbook version used in system tests

The Configure step of the systemtests was failing by long time with an error like:
```
This AMI was created with aws-parallelcluster-cookbook-3.2.0,
but is trying to be used with aws-parallelcluster-cookbook-3.6.0.
Please either use an AMI created with aws-parallelcluster-cookbook-3.6.0
or change your ParallelCluster to aws-parallelcluster-cookbook-3.2.0
```
See for example: https://github.com/aws/aws-parallelcluster-cookbook/actions/runs/4198312110/jobs/7281766085

### Permit the Configure step of SystemTests to fail
Return the exit code of the docker in case of failure.
```
$ docker run debian bash -c "exit 1"; echo $?
1
$ docker run debian bash -c "exit 0"; echo $?
0
```
Verified this approach works with the run: https://github.com/aws/aws-parallelcluster-cookbook/actions/runs/4202791232/jobs/7291394251

### Skip slurm_accounting recipe execution on SystemTests

This recipe has never been executed, the Configure step of the system tests was silently failing with:
``` 
[2023-02-17T09:52:28+00:00] FATAL: NoMethodError: ruby_block[Configure Slurm Accounting] (aws-parallelcluster-slurm::config_head_node line 211) had an error: NoMethodError: undefined method `dig' for nil:NilClass
```

The `dig` command cannot be executed in the empty `node['cluster']['config']` attribute
and mock it means write a cluster configuration file, we can skip it for now.

### DIFF
--- a/.github/workflows/system-tests-centos7.yml
+++ b/.github/workflows/system-tests-centos7.yml
@@ -36,4 +36,4 @@ jobs:
       - name: Configure
         run: |
           sed -i 's/\(.*base_os":\).*/\1 "centos7",/' ${{ github.workspace }}/system_tests/dna.json
-          docker run -i -v ${{ github.workspace }}:/build -v ${{ github.workspace }}/system_tests/dna.json:/etc/chef/dna.json localhost:5000/pcluster/chef-base:centos7 /build/system_tests/systemd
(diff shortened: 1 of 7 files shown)


eantonin (dismissed):

eantonin (dismissed):
LGTM, the mock for the slurm_accounting config will be done in future PRs (as agreed offline)

eantonin (approved):
LGTM

PR #1794, opened 2023-02-21
[RedHat8] Add head_node_base tests

francesco-giordano:
### Description of changes
* [Change aws_domain to work in test without region specified](https://github.com/aws/aws-parallelcluster-cookbook/pull/1794/commits/549fd0a4a6c3c68f69843c82b28f17711d752317)
* [Add head_node_base test](https://github.com/aws/aws-parallelcluster-cookbook/pull/1794/commits/1ac0da8b742d412fc53c26aec34434879d77a598)

### Tests
* Manually tested on Ec2 and on Docker only in GitHub action since on the Mac the exportfs from the docker is not permitted.


cookbooks/aws-parallelcluster-config/recipes/head_node_base.rb
```diff
@@ -30,22 +30,22 @@
   network get_vpc_cidr_list
   writeable true
   options ['no_root_squash']
-end
+end unless virtualized?
```
awslitvit: Why did you need to exclude docker from it? How did system tests work without this exclusion?

libraries/helpers.rb
```diff
@@ -256,7 +256,7 @@ def platform_supports_dcv?
 def aws_domain
   # Set the aws domain name
   aws_domain = "amazonaws.com"
-  aws_domain = "#{aws_domain}.cn" if node['cluster']['region'].start_with?("cn-")
```
awslitvit, line 225: Why don't we set the region for tests instead of changing the code?
The region should always be set, correct?

test/recipes/controls/aws_parallelcluster_config/head_node_base_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+control 'head_node_base_configured' do
+  title 'Check the base headnode configuration'
+
+  describe command("sudo su -c 'ls /home/*/.ssh/authorized_keys_cluster'") do
```
awslitvit, line 17: Nitpick: since it is not immediate to understand what this instruction verifies, I would add a comment.

kitchen.docker.yml
```diff
@@ -28,27 +28,115 @@ platforms:
...
+            mac: 02:04:c9:66:6c:23
+            public_hostname: ec2-3-253-9-238.eu-west-1.compute.amazonaws.com
+            public_ipv4s: 3.253.9.238
+            security_group_ids: sg-0d6ed39436326b0c1
+            security_groups: Security Group kitchen
+            subnet_id: subnet-061d5d152f1ea6e1f
+            subnet_ipv4_cidr_block: 10.0.0.0/24
+            vpc_id: vpc-0f2e67ad1d48c13c6
+            vpc_ipv4_cidr_block: 10.0.0.0/16
+            vpc_ipv4_cidr_blocks: 10.0.0.0/16
```
awslitvit: Should we add this block at `head_node_base` unit test level?
We should probably need it also for the validation step, which I don't see in the PR.
I would not set it up for every test on docker, unless you prove me wrong.

cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
```diff
@@ -87,5 +87,5 @@
   path "/etc/hosts"
   pattern "^#{node['ec2']['local_ipv4']}\s+"
   line(lazy { "#{node['ec2']['local_ipv4']} #{node['cluster']['assigned_hostname'].chomp('.')} #{node['cluster']['assigned_short_hostname']}" })
-  notifies :reload, "ohai[reload_hostname]"
+  notifies :reload, "ohai[reload_hostname]", :immediately
```
awslitvit: Why did you need this?

kitchen.ec2.yml
```diff
@@ -31,6 +31,7 @@ driver:
     instance_metadata_tags: 'enabled'
   #iam_profile_name: chef-client
   instance_type: <%= ENV['KITCHEN_INSTANCE_TYPE'] || "t2.micro" %>
+  iam_profile_name: <%= ENV['AWS_IAM_PROFILE'] %>
```
awslitvit: 1. You should document what profile is required to run the test
2. This breaks the possibility to run tests without setting env variables, relying on defaults. Could we think of a default? Is there a way to leverage tags?

kitchen.ec2.yml
```diff
@@ -67,6 +68,10 @@ platforms:
       <% end %>
     transport:
       username: ec2-user
+    attributes:
+      cluster:
+        ddb_table: <%= ENV['DDB_TABLE'] %>
```
awslitvit: As far as I understand, this attribute is required for `head_node_base` test and not elsewhere. Can we add it to the specific test? Also, I suppose we don't use this table name but to resolve a variable in a template. It can literally be any value. And I would check the contents of `clusterstatusmgtd.conf` and verify there is a line:
```
dynamodb_table = WHATEVER_VALUE_YOU_SET
```

test/recipes/controls/aws_parallelcluster_config/head_node_base_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    its('owner') { should eq 'root' }
+    its('group') { should eq 'root' }
+    its('mode') { should cmp '0755' }
+  end
+
+  describe file('/etc/parallelcluster/clusterstatusmgtd.conf') do
+    it { should exist }
+    its('owner') { should eq 'root' }
+    its('group') { should eq 'root' }
+    its('mode') { should cmp '0644' }
```
awslitvit, line 39: Please check the contents of the file.

awslitvit (changes requested):

libraries/helpers.rb
francesco-giordano: Yes, but this was always a pain point when I had to test on an instance not in a cluster: example run instance from a ParallelCluster AMI.

cookbooks/aws-parallelcluster-slurm/recipes/init_dns.rb
francesco-giordano: Typo

cookbooks/aws-parallelcluster-config/recipes/head_node_base.rb
francesco-giordano: Export filesystem is difficult from a docker container. I do not see a specific value to invest in test it in the docker container. What do you think?

kitchen.docker.yml
francesco-giordano: These attributes are a mock of ec2 value. Every test in running on docker might leverage this configuration or override if needed. Do you prefer to keep this file cleaner moving the attributes under `head_node_base`?

I have tried to use the yaml references to not repeat the code but GITHUB does not liked it.

libraries/helpers.rb
awslitvit: I understand but I'm always reluctant to "weaken" production code for testing reasons, unless it is unavoidable.

awslitvit (approved):

PR #1797, opened 2023-02-22
First stab at ChefSpec unit tests

awslitvit:
### Description of changes
* First stab at ChefSpec unit tests

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,3 +85,17 @@ jobs:
         with:
           severity: error
           ignore_paths: third-party
+  rspec:
+    name: ChefSpec
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - uses: actionshub/chef-install@main
+        with:
+          version: 22.10.1013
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -85,3 +85,17 @@ jobs:
...
           severity: error
           ignore_paths: third-party
+  rspec:
+    name: ChefSpec
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - uses: actionshub/chef-install@main
+        with:
+          version: 22.10.1013
```
enrico-usai, line 95: Is it required to specify a version? or can we just use the latest?

enrico-usai:
Thanks for adding ChefSpec into the project!

.github/workflows/ci.yml
awslitvit: Unfortunately no, because it would use Chef 18, while we enforce Chef 17 in our code.

enrico-usai (approved):

PR #1800, opened 2023-02-23
[Isolated Regions] Add support for US isolated regions: us-iso-* and us-isob-*.

gmarciani:
### Changes
Add support for US isolated regions: **us-iso-*** and **us-isob-***.

In particular:

1. Added support for the AWS domain in these regions.
2. Added script ‘/opt/parallelcluster/scripts/patch-iso-instance.sh’ that is expected to be used at run-time
    by both head and compute nodes to configure the instance during their bootstrap phase in these regions.
3. Added environment variable ‘AWS_CA_BUNDLE’ to the environment loaded by supervisord in all the daemons when 
    running in US isolated regions.
5. Added the explicit setting of the CA bundle in `ec2_dev_2_volid.py` when running in US isolated regions.
6. Added ‘--verbose’ option to the execution of cfn-hup with supervisord to make it easier
    to troubleshoot cfn-hup issues.
7. Added support for these regions to the scripts used to install Cinc.
8. Added a comment to ‘manage_fsx’ recipe to explain why the file system domain must not be adapted
    to support these regions.

##### Notes
This PR is similar to the one already approved: https://github.com/aws/aws-parallelcluster-cookbook/pull/1761, but with the following changes:
1. point 5 above has been added: Added the explicit setting of the CA bundle in `ec2_dev_2_volid.py` when running in US isolated regions.
2. point 4 from previously approved PR has been removed because further tests demonstrated it was not necessary: Added /etc/sudoers.d/99-parallelcluster-env-keep to preserve the environment variable AWS_CA_BUNDLE when switching user in non interactive sessions. This is done at config time only when running in US isolated regions.

### Tests
Manual tests:
1. AMI build (executed in Commercial)
3. Cluster creation (executed both in Commercial and US Isolated)
4. Cluster update with dynamic file system mounting (executed both in Commercial and US Isolated)
5. Cluster update with compute nodes min-count change (executed both in Commercial and US Isolated)
6. Submission of job that requires compute fleet autoscaling (executed both in Commercial and US Isolated)

A full regression tests will be done on the authoritative pipeline once the change will be merged.

### DIFF
--- a/cookbooks/aws-parallelcluster-install/recipes/base_isolated.rb
+++ b/cookbooks/aws-parallelcluster-install/recipes/base_isolated.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+# Recipe:: base_isolated
+#
+# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+cookbook_file "#{node['cluster']['scripts_dir']}/patch-iso-instance.sh" do
+  source 'base/patch-iso-instance.sh'
+  owner 'root'
+  group 'root'
+  mode '0744'
+end
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -303,10 +303,17 @@ def platform_supports_dcv?
 end
 
 def aws_domain
-  # Set the aws domain name
-  aws_domain = "amazonaws.com"
-  aws_domain = "#{aws_domain}.cn" if node['cluster']['region'].start_with?("cn-")
-  aws_domain
+  # Get the aws domain name
+  region = node['cluster']['region']
+  if region.start_with?("cn-")
+    "amazonaws.com.cn"
+  elsif region.start_with?("us-iso-")
+    "c2s.ic.gov"
+  elsif region.start_with?("us-isob-")
+    "sc2s.sgov.gov"
+  else
+    "amazonaws.com"
+  end
 end
 
 def kernel_release
--- a/system_tests/install_cinc.sh
+++ b/system_tests/install_cinc.sh
@@ -31,13 +31,21 @@ elif [ `echo "${OS}" | grep -E '^ubuntu'` ]; then
   PLATFORM='DEBIAN'
 fi
 
-BUCKET="s3.amazonaws.com"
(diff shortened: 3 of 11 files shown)


cookbooks/aws-parallelcluster-install/recipes/base_isolated.rb
```diff
@@ -0,0 +1,23 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+cookbook_file "#{node['cluster']['scripts_dir']}/patch-iso-instance.sh" do
+  source 'base/patch-iso-instance.sh'
+  owner 'root'
+  group 'root'
+  mode '0754'
```
enrico-usai: question: why not 0744?

libraries/helpers.rb
```diff
@@ -303,9 +303,11 @@ def platform_supports_dcv?
 end
 
 def aws_domain
-  # Set the aws domain name
+  # Get the aws domain name
   aws_domain = "amazonaws.com"
-  aws_domain = "#{aws_domain}.cn" if node['cluster']['region'].start_with?("cn-")
+  aws_domain = "amazonaws.com.cn" if node['cluster']['region'].start_with?("cn-")
```
enrico-usai: nitpick: it's cleaner to have a switch case

system_tests/install_cinc.sh
```diff
@@ -31,12 +31,19 @@ elif [ `echo "${OS}" | grep -E '^ubuntu'` ]; then
...
+S3_ENDPOINT="s3.${AWS_Region}.${AWS_DOMAIN}"
+
+BUCKET="cloudformation-examples"
+[[ ${AWS_DOMAIN} != "amazonaws.com" ]] && BUCKET="${AWS_Region}-aws-parallelcluster/cloudformation-examples"
 if [[ ${OS} =~ ^(ubuntu2004)$ ]]; then
-  CfnBootstrapUrl="https://${BUCKET}/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz"
+  CfnBootstrapUrl="https://${S3_ENDPOINT}/${BUCKET}/aws-cfn-bootstrap-py3-latest.tar.gz"
 else
-  CfnBootstrapUrl="https://${BUCKET}/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz"
+  CfnBootstrapUrl="https://${S3_ENDPOINT}/${BUCKET}/aws-cfn-bootstrap-latest.tar.gz"
```
enrico-usai: nitpick: here we could set another variable with the package name and at the end have:
```
  CfnBootstrapUrl="https://${S3_ENDPOINT}/${BUCKET}/${PACKAGE_NAME}"
```
to avoid repetition

enrico-usai (dismissed):

cookbooks/aws-parallelcluster-install/recipes/base_isolated.rb
gmarciani: I agree will fix in a follow up PR.

libraries/helpers.rb
gmarciani: I agree will fix in a follow up PR.

system_tests/install_cinc.sh
gmarciani: I agree will fix in a follow up PR.

enrico-usai (approved):

PR #1803, opened 2023-02-23
Move attributes in the corresponding default files and disable unit tests

francesco-giordano:
### Description of changes
* Move attributes in the corresponding default files
* Disable unit tests. The tests can be started manually
* https://github.com/aws/aws-parallelcluster-cookbook/pull/1793

### Tests
* Manual tested


.github/workflows/dokken-validate.yml
```diff
@@ -13,8 +13,6 @@ jobs:
           - ubuntu18
           - ubuntu20
           - redhat8
-        suite:
-          - aws-parallelcluster-install
```
awslitvit: Why did you remove these lines? 
Even if we don't need them now, we can keep them for when we'll have more suites to run in parallel, can't we?

francesco-giordano: Since now it is not mandatory but we can keep it there like a place holder: I'm going to add the lines again.

awslitvit (approved):

PR #1805, opened 2023-02-27
[3.x] Add EFA error message on RH < 8.4

eantonin:
### Description of changes
* Add EFA error message on RH < 8.4
* To be able to add the RH 8.4 if I needed to change the `setup` action name, and introduce the setup action in all the subrecipes (haven't found a better way to do that yet)

### Tests
* ChefSpec
* EC2 tests
* Docker tests

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-common/resources/efa/efa_redhat8.rb
+++ b/cookbooks/aws-parallelcluster-common/resources/efa/efa_redhat8.rb
@@ -24,6 +24,18 @@ action :configure do
   # do nothing
 end
 
+action :check_efa_support do
+  if node['cluster']['platform_version'].to_f < 8.4
+    log "EFA is not supported in this RHEL version #{node['cluster']['platform_version']}, supported versions are >= 8.4" do
+      level :warn
+      action :write
+    end
+    node.override['cluster']['efa_supported'] = false
+  else
+    node.override['cluster']['efa_supported'] = true
+  end
+end
+
 action_class do
   def conflicting_packages
     %w(openmpi-devel openmpi)
--- a/cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
+++ b/cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
@@ -16,29 +16,32 @@
 efa_tarball = "#{node['cluster']['sources_dir']}/aws-efa-installer.tar.gz"
 
 action :setup do
-  if efa_installed? && !::File.exist?(efa_tarball)
-    Chef::Log.warn("Existing EFA version differs from the one shipped with ParallelCluster. Skipping ParallelCluster EFA installation and configuration.")
-    return
-  end
+  action_check_efa_support
+  if node['cluster']['efa_supported']
+    if efa_installed? && !::File.exist?(efa_tarball)
+      Chef::Log.warn("Existing EFA version differs from the one shipped with ParallelCluster. Skipping ParallelCluster EFA installation and configuration.")
+      return
+    end
 
-  # remove conflicting packages
-  # default openmpi installation conflicts with new install
-  # new one is installed in /opt/amazon/efa/bin/
-  package conflicting_packages do
-    action :remove
-    not_if { efa_installed? }
-  end
+    # remove conflicting packages
+    # default openmpi installation conflicts with new install
+    # new one is installed in /opt/amazon/efa/bin/
+    package conflicting_packages do
+      action :remove
+      not_if { efa_installed? }
+    end
 
-  # update repos and install prerequisite packages
-  package_repos 'update package repos' do
-    action :update
-  end
-  package %w(environment-modules) do
-    retries 3
-    retry_delay 5
-  end
+    # update repos and install prerequisite packages
+    package_repos 'update package repos' do
+      action :update
+    end
+    package %w(environment-modules) do
+      retries 3
+      retry_delay 5
+    end
 
-  action_download_and_install
+    action_download_and_install
+  end
 end
 
 action :download_and_install do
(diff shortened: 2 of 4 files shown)


cookbooks/aws-parallelcluster-install/attributes/default.rb
```diff
@@ -4,3 +4,5 @@
 default['cluster']['sysctl']['ipv4']['gc_thresh1'] = 0
 default['cluster']['sysctl']['ipv4']['gc_thresh2'] = 15_360
 default['cluster']['sysctl']['ipv4']['gc_thresh3'] = 16_384
+# Created this new variable to be able to mock it freely
+default['cluster']['platform_version'] = node['platform_version']
```
francesco-giordano: Question: is it intended to ne in the install cookbook? Shouldn't be in the common cookbook since it is used there?

cookbooks/aws-parallelcluster-common/resources/efa/efa_redhat8.rb
```diff
@@ -20,6 +20,17 @@
 
 use 'partial/_setup'
 
+action :setup do
+  if node['cluster']['platform_version'].to_f < 8.4
```
francesco-giordano, line 28: Question: why is `node['cluster']['platform_version']` used instead of `node['platform_version']`?

cookbooks/aws-parallelcluster-common/spec/unit/resources/efa_spec.rb
```diff
@@ -146,6 +147,25 @@ def mock_efa_supported(supported)
       end
     end
   end
+
+  context 'when rhel version is older than 8.4' do
+    let(:chef_run) do
+      ChefSpec::Runner.new(
+        # Create a runner for the given platform/version
+        platform: "redhat",
```
francesco-giordano, line 155: If it is possible to use the `node['platform_version']`in the code above, in the ChefSpec `version` should be able to override that value

cookbooks/aws-parallelcluster-install/attributes/default.rb
eantonin: You're right, will fix

cookbooks/aws-parallelcluster-common/resources/efa/efa_redhat8.rb
eantonin: `node['platform_version']` is kind of a reserved attribute, so it's not possible to override it (because it gets overridden by the actual test platform version) so I need to use a different parameter

cookbooks/aws-parallelcluster-common/spec/unit/resources/efa_spec.rb
eantonin: Nope, I tried that, it does not work because there is no platform version that can be used for the specific red hat version.

Error is `Could not find platform 'redhat/8.3' `

cookbooks/aws-parallelcluster-install/attributes/default.rb
eantonin: (I do not get how it worked with the file in the wrong folder)

cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
```diff
@@ -16,29 +16,31 @@
...
-    action :remove
-    not_if { efa_installed? }
-  end
+    # remove conflicting packages
+    # default openmpi installation conflicts with new install
+    # new one is installed in /opt/amazon/efa/bin/
+    package conflicting_packages do
+      action :remove
+      not_if { efa_installed? }
+    end
```
awslitvit, line 32: I don't understand this change: looks like you added a whitespace at the beginning of the line. Could you please check?

cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
```diff
@@ -16,29 +16,31 @@
...
-    retry_delay 5
-  end
+    # update repos and install prerequisite packages
+    package_repos 'update package repos' do
+      action :update
+    end
+    package %w(environment-modules) do
+      retries 3
+      retry_delay 5
+    end
```
awslitvit, line 40: Same as above.

cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
```diff
@@ -16,29 +16,31 @@
...
 
-  action_download_and_install
+    action_download_and_install
```
awslitvit, line 42: Same as above.

cookbooks/aws-parallelcluster-common/resources/efa/efa_redhat8.rb
```diff
@@ -24,6 +24,18 @@
   # do nothing
 end
 
+action :check_efa_support do
```
awslitvit, line 27: Nitpick. You use the action as a plain ruby function. In fact, you can't return a value from an actual action.
It would be cleaner if you define a function (i.e. private behavior) or if you keep using the action (i.e. public behavior) but set an attribute as the outcome. If you choose the former, you'll need to pass the node as argument.

cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
eantonin: This is the indentation due to the new if

cookbooks/aws-parallelcluster-common/resources/efa/efa_redhat8.rb
eantonin: Pushed a new version using the latter

francesco-giordano (dismissed):

cookbooks/aws-parallelcluster-common/resources/efa/partial/_setup.rb
awslitvit: Oh sorry, right!

awslitvit (approved):

PR #1808, opened 2023-02-27
Use start of string anchor regex for dir path ref

nyetsche:
Previous to this commit, the pattern sent to `delete_lines` is an (interpolated) string followed by a space, such as: `/nfs `, operating on /etc/exports. Without the start of string anchor regex (`^`), the string will match a different line that shares includes the string and whitespace. Example:

`/shared `

will match:

`/shared ` (good)
`/opt/parallelcluster/shared ` (bad!)

Including a `^` forces the string to match from the beginning of the line.

### Tests
* Manually tested by operating on a ParallelCluster v3.4.1 instance. Adding the lines:

```
SharedStorage:
  - MountDir: /shared
    Name: shared
    StorageType: Ebs
    EbsSettings:
      VolumeType: st1
      Size: 500
```

then then _removing_ those lines (`MountDir: /shared` being the critical point) results in `/etc/exports` missing both `/shared` and `/opt/parallelcluster/shared`. Further cluster operations (ex: spin-up a new instance) fail because the parallelcluster software is no longer exported over NFS.

I'm not familiar with Chef, so I didn't see an obvious place to add or manually run a test. I'm hoping the automatic ones in Github will cover me.

--- 


chenwany:
Thank you for reporting the issue and providing the PR!

nyetsche:
Is there anything I need to do to have the CI checks run? I received an email about runs that took place, but they don't show up here. `python (ubuntu-latest, Code Checks)` failed, but I think that's just because of the `develop` branch, not my PR.

chenwany (approved):
LGTM!

PR #1011, opened 2021-06-24
3.0: Drop support for CentOS8

enrico-usai:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1822, opened 2023-03-02
Use a custom NetworkManager.conf with the correct dhcp plugin in Red Hat 8

eantonin:
### Description of changes
* Use a custom NetworkManager.conf with the correct dhcp plugin in Red Hat 8

### Tests
* Created a test that was passing in all OS apart from RH8
* Fixed the code and now the test passed also in RH8

### References
* Solution inspired by https://github.com/aws/aws-parallelcluster-cookbook/pull/1011

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/files/redhat/dns_domain/NetworkManager.conf
+++ b/cookbooks/aws-parallelcluster-slurm/files/redhat/dns_domain/NetworkManager.conf
@@ -1,0 +1,52 @@
+# Configuration file for NetworkManager.
+#
+# See "man 5 NetworkManager.conf" for details.
+#
+# The directories /usr/lib/NetworkManager/conf.d/ and /run/NetworkManager/conf.d/
+# can contain additional .conf snippets installed by packages. These files are
+# read before NetworkManager.conf and have thus lowest priority.
+# The directory /etc/NetworkManager/conf.d/ can contain additional .conf
+# snippets. Those snippets are merged last and overwrite the settings from this main
+# file.
+#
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-slurm/files/redhat/dns_domain/NetworkManager.conf
```diff
@@ -0,0 +1,52 @@
...
+# by having a file NAME.conf in either /run/NetworkManager/conf.d/ or /etc/NetworkManager/conf.d/.
+# Likewise, snippets from /run can be prevented from loading by placing
+# a file with the same name in /etc/NetworkManager/conf.d/.
+#
+# If two files define the same key, the one that is read afterwards will overwrite
+# the previous one.
+
+[main]
+plugins = ifcfg-rh,
+dhcp = dhclient
```
enrico-usai, line 24: what about having a spec test to verify this file is created and the dhcp = dhclient line is there?

enrico-usai (approved):

cookbooks/aws-parallelcluster-slurm/files/redhat/dns_domain/NetworkManager.conf
eantonin: I'm conflicted on this. I see your point.

Anyway I'm not sure we want to add it as in the InSpec tests we should test the effect of actions, and the effect of adding this configuration just lets the resolv.conf file to be created correctly also in redhat8. So I won't add that test for the moment

PR #1824, opened 2023-03-02
[Multi User] Add support for AD password as SSM Parameter.

gmarciani:
### Description of changes
Add support for AD password as SSM Parameter.
With this extension we will be able to provide the multi-user feature in those regions where Secrets Manager is not supported, but SSM is.

#### Customer Experience
With this change the customer will be able to specify the AD password in [DirectoryService/PasswordSecretArn](https://docs.aws.amazon.com/parallelcluster/latest/ug/DirectoryService-v3.html#yaml-DirectoryService-PasswordSecretArn) both as a secret in Secrets Manager (current behaviour) and a parameter in SSM.

### Tests
* Cluster creation + AD user login with AD password specified as a secret in Secrets Manager.
* Cluster creation + AD user login with AD password specified as a parameter in SSM.
* AD Password sync with `update_directory_service_password.sh` using Secrets Manager.
* AD Password sync with `update_directory_service_password.sh` using SSM.

### DIFF
--- a/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
@@ -25,6 +25,7 @@ shared_directory_service_dir = "#{node['cluster']['shared_dir']}/directory_servi
 shared_sssd_conf_path = "#{shared_directory_service_dir}/sssd.conf"
 
 if node['cluster']['node_type'] == 'HeadNode'
+  region = node['cluster']['region']
   # DomainName
   # We can assume that DomainName can only be a FQDN or the domain section in a LDAP Distinguished Name.
   # We can assume it because the CLI is in charge of validating it.
@@ -49,6 +50,25 @@ if node['cluster']['node_type'] == 'HeadNode'
     end
   end
 
+  # Password
+  # The Active Directory Password can be:
+  #   1. A secret in Secrets Manager, e.g. arn:aws:secretsmanager:eu-west-1:12345678910:secret:PasswordName
+  #   2. A parameter in SSM, e.g. arn:aws:ssm:eu-west-1:12345678910:parameter/PasswordName
+  password_secret_arn_str = node['cluster']['directory_service']['password_secret_arn']
+  password_secret_arn = parse_arn(password_secret_arn_str)
+  password_secret_service = password_secret_arn[:service]
+  password_secret_resource = password_secret_arn[:resource]
+  ldap_password =
+    if password_secret_service == "secretsmanager" && password_secret_resource.start_with?("secret")
+      (secret_name = password_secret_resource.split(":")[1])
+      shell_out!("aws secretsmanager get-secret-value --secret-id #{secret_name} --region #{region} --query 'SecretString' --output text").stdout.strip
+    elsif password_secret_service == "ssm" && password_secret_resource.start_with?("parameter")
+      (parameter_name = password_secret_resource.split("/")[1])
+      shell_out!("aws ssm get-parameter --name #{parameter_name} --region #{region} --with-decryption --query 'Parameter.Value' --output text").stdout.strip
+    else
+      raise "The secret #{password_secret_arn_str} is not supported"
+    end
+
   # Head node writes the sssd.conf file and contacts the secret manager to retrieve the LDAP password.
   # Then the sssd.conf file is shared through shared_sssd_conf_path to compute nodes.
   # Only contacting the secret manager from head node avoids giving permission to compute nodes to contact the secret manager.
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -49,6 +50,22 @@
     end
   end
 
+  # Password
+  password_secret_arn_str = node['cluster']['directory_service']['password_secret_arn']
+  password_secret_arn = parse_arn(password_secret_arn_str)
+  password_secret_service = password_secret_arn[:service]
+  password_secret_resource = password_secret_arn[:resource]
+  ldap_password =
+    if password_secret_service == "secretsmanager" && password_secret_resource.start_with?("secret")
```
enrico-usai, line 62: it would be useful to have some examples of the format of the password_secret_arn in the two cases

cookbooks/aws-parallelcluster-config/templates/default/directory_service/update_directory_service_password.sh.erb
```diff
@@ -20,20 +20,34 @@ REGION="<%= @region %>"
...
-echo "Writing AWS Secrets Manager password to ${SSSD_CONFIG_FILE}"
-python3 -c "${PYTHON_CODE_READ_CONFIG}; config['${SSSD_SECTION}']['${SSSD_PROPERTY}']='${password_from_secrets_manager}'; config.write(open(file,'w'))"
-echo "Password updated in ${SSSD_CONFIG_FILE}"
-cp ${SSSD_CONFIG_FILE} ${SSSD_SHARED_CONFIG_FILE}
-echo "${SSSD_CONFIG_FILE} copied to ${SSSD_SHARED_CONFIG_FILE}"
-
-echo "Restarting services"
+echo "[INFO] Reading password from AWS: ${SECRET_ARN}"
+service=$(echo "${SECRET_ARN}" | cut -d ':' -f 3)
+resource=$(echo "${SECRET_ARN}" | cut -d ':' -f 6-)
```
enrico-usai, line 28: same here it would be useful to see fthe format of the parameter we're parsing to better understand the code in the two cases.

enrico-usai: Question: what does the `6-` mean?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
```diff
@@ -49,6 +50,22 @@
...
+      (secret_name = password_secret_resource.split(":")[1])
+      shell_out!("aws secretsmanager get-secret-value --secret-id #{secret_name} --region #{region} --query 'SecretString' --output text").stdout.strip
+    elsif password_secret_service == "ssm" && password_secret_resource.start_with?("parameter")
```
enrico-usai, line 65: This block seems a duplicate of the one in the bash script.

What about defining a bash(python) script to be called by both the places and leave the logic to decide if use secrets manager or ssm in that script?

libraries/helpers.rb
```diff
@@ -600,3 +600,16 @@ def format_directory(dir)
   format_dir = "/#{format_dir}" unless format_dir.start_with?('/')
   format_dir
 end
+
+def parse_arn(arn_string)
+  parts = arn_string.nil? ? [] : arn_string.split(':', 6)
```
enrico-usai, line 610: let's add an example of the format here as well. 
Question: Is this vaild for all the ARNs ?

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: Doing in next revision

cookbooks/aws-parallelcluster-config/templates/default/directory_service/update_directory_service_password.sh.erb
gmarciani: Examples of ARNs will be added in the next revision.
`6-` means that fields from the 6th onwards will be returned. This is required because an ARN is made of `arn:PARTITION:SERVICE:REGION:ACCOUNT:RESOURCE`
1. for a secret in Secrets Manager, the RESOURCE field must be `secret:PasswordName`
2. for a parameter in SSM, the RESOURCE field must be `parameter/PasswordName`

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
gmarciani: Yes the logic is duplicated in two places, but I do not think it's worth it to put this logic in a common script.
To do that we would put into the AMI a script with this common logic to make it callable by both the recipe and another AMI script. I don't think it's worth it for this logic. Do you know other ways?

libraries/helpers.rb
gmarciani: Adding comments on the format in next revision.
Yes, the vcode covers alla types of ARN. It's actually the same logic implemented by the AWS SDK for Ruby v3.

cookbooks/aws-parallelcluster-config/recipes/directory_service.rb
enrico-usai: Nope, I was thinking the same, a file to be used in both the places.

enrico-usai (approved):

PR #1863, opened 2023-03-09
3.x Publish JSON event logs to CloudWatch

davprat:
### Description of changes
* This change adds configuration to CloudWatch Agent to publish structured JSON event logs to CloudWatch log streams.
* This change also changes write_cloudwatch_agent_json.py so that it supports log streams without timestamps. This is required because CloudWatch won't parse a log event as JSON if it included anything before the JSON structure.

### Tests
* Manually deployed to personal account
* Verified log events were published to CloudWatch

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
+++ b/cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
@@ -4,7 +4,7 @@
     "default": "%Y-%m-%d %H:%M:%S,%f",
     "bracket_default": "[%Y-%m-%d %H:%M:%S]",
     "slurm": "%Y-%m-%dT%H:%M:%S.%f",
-    "metric": ""
+    "json": ""
   },
   "log_configs": [
     {
@@ -165,7 +165,7 @@
       "feature_conditions": []
     },
     {
-      "timestamp_format_key": "metric",
+      "timestamp_format_key": "json",
       "file_path": "/var/log/parallelcluster/clustermgtd.events",
       "log_stream_name": "clustermgtd_events",
       "schedulers": [
@@ -183,9 +183,9 @@
       "feature_conditions": []
     },
     {
-      "timestamp_format_key": "metric",
+      "timestamp_format_key": "json",
       "file_path": "/var/log/parallelcluster/slurm_resume.events",
-      "log_stream_name": "resume_events",
+      "log_stream_name": "slurm_resume_events",
       "schedulers": [
         "slurm"
       ],
--- a/cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
+++ b/cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
@@ -3,7 +3,8 @@
     "month_first": "%b %-d %H:%M:%S",
     "default": "%Y-%m-%d %H:%M:%S,%f",
     "bracket_default": "[%Y-%m-%d %H:%M:%S]",
-    "slurm": "%Y-%m-%dT%H:%M:%S.%f"
+    "slurm": "%Y-%m-%dT%H:%M:%S.%f",
+    "metric": ""
   },
   "log_configs": [
     {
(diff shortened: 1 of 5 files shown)


cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -168,14 +168,19 @@ def add_scheduler_plugin_log(config_data, cluster_config_path):
...
 
 
 def filter_output_fields(configs):
     """Remove fields that are not required by CloudWatch agent config file."""
     desired_keys = ["log_stream_name", "file_path", "timestamp_format", "log_group_name"]
-    return [{desired_key: config[desired_key] for desired_key in desired_keys} for config in configs]
+    return [
+        {desired_key: config[desired_key] for desired_key in desired_keys if desired_key in config}
+        for config in configs
+    ]
```
delongmeng, line 183: Is it worth changing to something like
```
desired_keys = {"log_stream_name", "file_path", "timestamp_format", "log_group_name"}
return [
  {key: config[key] for key in desired_keys & set(config.keys())}
  for config in configs
]
```
for readability?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
```diff
@@ -3,7 +3,8 @@
     "month_first": "%b %-d %H:%M:%S",
     "default": "%Y-%m-%d %H:%M:%S,%f",
     "bracket_default": "[%Y-%m-%d %H:%M:%S]",
-    "slurm": "%Y-%m-%dT%H:%M:%S.%f"
+    "slurm": "%Y-%m-%dT%H:%M:%S.%f",
+    "metric": ""
```
delongmeng: Minor question for the name "metric" here, does it make sense to name it as "empty", or "ignore", "none", "skip", etc., so that we can re-use it in the future for any situation (not limited to the metric use case) when we don't want a timestamp?

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
```diff
@@ -163,6 +164,42 @@
...
+      ],
+      "node_roles": [
+        "HeadNode"
+      ],
+      "feature_conditions": []
+    },
+    {
+      "timestamp_format_key": "metric",
+      "file_path": "/var/log/parallelcluster/slurm_resume.events",
+      "log_stream_name": "resume_events",
```
delongmeng: Do we usually keep the log stream name and the log name consistent? If that's the case, maybe `slurm_resume_events` for log stream?

davprat: Yes.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/cloudwatch_agent_config.json
davprat: I think you're right that "metric" is wrong. I think the correct name should be "json" since you should only not have a timestamp in the case where you want CW to parse JSON logs.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
davprat: Yeah it is.

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
```diff
@@ -168,14 +168,16 @@ def add_scheduler_plugin_log(config_data, cluster_config_path):
...
+        if timestamp_format:
+            config["timestamp_format"] = timestamp_format
     return configs
 
 
 def filter_output_fields(configs):
     """Remove fields that are not required by CloudWatch agent config file."""
     desired_keys = ["log_stream_name", "file_path", "timestamp_format", "log_group_name"]
-    return [{desired_key: config[desired_key] for desired_key in desired_keys} for config in configs]
+    return [{desired_key: config[desired_key] for desired_key in (desired_keys & config.keys())} for config in configs]
```
delongmeng: Very minor but maybe we can replace "desired_key" with just "key" here?
```
{key: config[key] for key in (desired_keys & config.keys())}
```
because the "key" now is any element from the intersection of two groups, instead of only the "desired_keys" group

chenwany: nit: This can also be
```
return [{desired_key: config[desired_key] for desired_key in desired_keys if desire_key in config } for config in configs]
```

chenwany (dismissed):
LGTM!

cookbooks/aws-parallelcluster-config/files/default/cloudwatch_agent/write_cloudwatch_agent_json.py
davprat: Updated.

chenwany (approved):
LGTM!

PR #1887, opened 2023-03-14
[3.x] Move installation of hostname package into dns_domain setup action, called from install cookbook

eantonin:
### Description of changes
* This is to avoid tests in no_internet subnet to fail
* Moved dns_domain resource in common to be called from both slurm and install packages
* Added dns_domain_setup dependency to init_dns recipe

### Tests
* Kitchen on docker and EC2 (in progress)

### Checklist


cookbooks/aws-parallelcluster-common/resources/dns_domain/dns_domain.rb
```diff
@@ -16,6 +16,13 @@
 use 'partial/_dns_search_domain_ubuntu'
 use 'partial/_dns_search_domain_redhat'
 
+action :setup do
```
enrico-usai, line 19: let's change the default_action to be `:setup`, in this way is consistent with most of the other resources (and be sure we're calling the righ action in the two places).

cookbooks/aws-parallelcluster-common/resources/dns_domain/dns_domain_redhat8.rb
```diff
@@ -17,6 +17,12 @@
 
 use 'partial/_dns_search_domain_redhat'
 
+action :setup do
```
enrico-usai, line 20: same here for teh default action

kitchen.resources.yml
```diff
@@ -172,6 +172,7 @@ suites:
     verifier:
       controls:
         - dns_domain_configured
+        - dns_domain_setup
```
enrico-usai: here, let's call the :configure explicitly.

eantonin: not needed in this particular point, but yes, I'll call explicitly the configure where needed after changing the default action

enrico-usai (approved):

PR #1894, opened 2023-03-14
Fix RedHat8 environment-modules config path and add test

francesco-giordano:
### Description of changes
* Add intel_mpi InSpec test 
* Fix RedHat8 environment-modules config path since the version of environment-modules is 4.x.x while in previous os was 3.x.x

### Tests
* Tested on docker and EC2

### References
* https://modules.readthedocs.io/en/v4.1.4/diff_v3_v4.html

### DIFF
--- a/cookbooks/aws-parallelcluster-install/recipes/intel_mpi.rb
+++ b/cookbooks/aws-parallelcluster-install/recipes/intel_mpi.rb
@@ -23,6 +23,12 @@ intelmpi_installer = "l_mpi_oneapi_p_#{node['cluster']['intelmpi']['full_version
 intelmpi_installer_path = "#{node['cluster']['sources_dir']}/#{intelmpi_installer}"
 intelmpi_installer_url = "https://#{node['cluster']['region']}-aws-parallelcluster.s3.#{node['cluster']['region']}.#{aws_domain}/archives/impi/#{intelmpi_installer}"
 
+# Prerequisite for module install
+package %w(environment-modules) do
+  retries 3
+  retry_delay 5
+end
+
 # fetch intelmpi installer script
 remote_file intelmpi_installer_path do
   source intelmpi_installer_url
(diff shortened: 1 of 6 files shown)


cookbooks/aws-parallelcluster-install/recipes/intel_mpi.rb
```diff
@@ -43,7 +49,7 @@
   creates intelmpi_installation_path.to_s
 end
 
-append_if_no_line "append intel modules file dir to modules conf" do
+append_if_no_line "append intel modules file dir to modules conf #{node['cluster']['modulepath_config_file']}" do
```
enrico-usai, line 52: The fact that the path changes in environment-modules 4.x let me think that if there is an update of this package in other OSes we'll be broken for that os as well.

I'm thinking at least in our kitchen spec tests we can have a  test to verify environment-modules version in every OS, to be sure to fail if the package is updated.

kitchen.validate.yml
```diff
@@ -40,7 +40,6 @@ suites:
         - write_common_udev_configuration_files
         - ec2blkdev_service_installation
         - debian_udevd_reload_configuration
-        - package_repos
```
enrico-usai: Why did you remove it? 

Is it needed only for intel_mpi? If this is not the case I prefer having it explicitly called.

enrico-usai:
Can you please add more information in the commit messages? 
It's important to clarify this difference between 3.x and 4.x version in the commit message.

cookbooks/aws-parallelcluster-install/recipes/intel_mpi.rb
francesco-giordano: I did not noticed but also ubuntu uses environment-modules version 4.x.x but they have solved the issue of compatibility adding directly in the package a link that point  `/etc/environment-modules/modulespath` to `/usr/share/modules/init/.modulespath`

There is already a test that verify that the intel_mpi module added in the cookbook is available

kitchen.validate.yml
francesco-giordano: No, I have removed it because that InSpec test has been remove in previous PR

enrico-usai (approved):

PR #1895, opened 2023-03-14
fetch_and_run in python supporting multiple custom actions

psacc:
### Description of changes
* fetch_and_run is now implemented in python in custom_actions_executor.py
* error messages and behavior is kept as much as possible as in the previous implementation in order to avoid impacting customer experience
* custom actions configuration is directly parsed from cluster configuration
* add support for executing multiple custom actions in sequence
* no longer depends on configuration provided via dna.json and cfnconfig, but still leaves cfnconfig sourceable in a backward compatible way (since custom actions scripts do not have a formally defined execution environment that can be relied upon)
* sorry the diff is big, couldn't find a way to make a meaningful smaller change, and also my python is a bit rusty 😅
* excluded security check, added comment in code to explain the reason

### Tests
* fetch_and_run python implementation custom_actions_executor has unit tests and coverage
* manual local integration test execution (will add some automation in a following change)
* local system test execution (> 1h per execution time)

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
+++ b/cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
@@ -1,0 +1,518 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import asyncio
+import logging
+import os
+import subprocess  # nosec B404
+import tempfile
+from builtins import RuntimeError
+from dataclasses import dataclass
+from enum import Enum
+from urllib.parse import urlparse
+
+import boto3
+import botocore
+import requests
+import yaml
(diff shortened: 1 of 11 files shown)


cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
```diff
@@ -0,0 +1,518 @@
...
+
+    stack_name: str
+    region_name: str
+    node_type: str
+    queue_name: str
+    event_name: str
+    legacy_event: LegacyEventName
+    can_execute: bool
+    dry_run: bool
+    script_sequence: list  # type list[ScriptDefinition]
```
ermanno, line 203: The situation for list types in Python is:
- Prior to 3.9 you can import capital case `List` from the `typing` module and specify the type of objects in the list with `List[ScriptDefinition]`.
- From 3.9 onward you can use `list[ScriptDefinition]` and the `List` form is deprecated, as the lower-case version supports subscripting (see [doc](https://docs.python.org/3/library/typing.html#typing.List))

ermanno (approved):

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
```diff
@@ -0,0 +1,518 @@
...
+            raise DownloadRunError(
+                f"Failed to run {self.event_name} script {exe_script.step_num} {exe_script.url} "
+                f"due to a failure in making the file executable, return code: {err.returncode}.",
+                f"Failed to run {self.event_name} script {exe_script.step_num} "
+                f"due to a failure in making the file executable, return code: {err.returncode}.",
+            ) from err
+
+        # execute script with it's args
+        try:
+            # arguments are provided by the user who has the privilege to create/update the cluster
```
psacc, line 139: security relevant exclusion: surfacing an issue already present in the bash implementation, see this comment in the code

PR #1896, opened 2023-03-14
Create volume resource

awslitvit:
### Description of changes
* Create volume resource to be used to handle all the volumes.

### Tests
* ChefSpec tests

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/resources/volume/volume.rb
+++ b/cookbooks/aws-parallelcluster-config/resources/volume/volume.rb
@@ -1,0 +1,130 @@
+provides :volume
+unified_mode true
+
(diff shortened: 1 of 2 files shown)


cookbooks/aws-parallelcluster-config/resources/volume/volume.rb
```diff
@@ -0,0 +1,130 @@
+provides :volume
+unified_mode true
+
+property :shared_dir, String, required: %i(mount export unmount)
```
francesco-giordano, line 4: Nitpick: since the resources seems general for every kind of volume, the name `shared_dir` can be more general

francesco-giordano (approved):

PR #1898, opened 2023-03-14
[develop] structured logging for compute node Chef recipe error

delongmeng:
### Description of changes
* Created a Chef handler (`WriteComputeFleetSlurmChefError`) for compute nodes of Slurm scheduler that can create a log file (`bootstrap_error_msg`) to store Chef error related information into json structure
* Added this log file into CloudWatch agent configuration so it can be pushed to CloudWatch (only for compute nodes)
* Make minor changes to the [previous Chef handler](https://github.com/aws/aws-parallelcluster-cookbook/pull/1596):
  * move the code from the global entry point to within the config cookbook
  * only conditionally register this handler for head node bootstrap 
  * rename to `WriteHeadNodeChefError`

### Tests
Manual test:
* Introduce some errors in Chef recipe compute fleet-specific part, create a cluster using Slurm scheduler so that Chef failure will be triggered in the compute node during cluster creation
  * see the error file log in CloudWatch with expected logstream and json structure content
  * confirm in the cloud-init-output log, only the slurm compute node handler was registered and executed
* Introduce some errors in Chef recipe head node-specific part, create a cluster so that Chef failure will be triggered in the head node during cluster creation
  * do not see the error file log in CloudWatch log group
  * confirm in the chef-client log, only the head node handler was registered and executed

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
+++ b/cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
@@ -14,6 +14,8 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'chef/handler'
+
 module WriteChefError
   # this class is used to handle chef errors and write the errors into a certain file if the file does not exist yet
   class WriteChefError < Chef::Handler
--- a/cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
+++ b/cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
@@ -25,11 +25,8 @@ module WriteChefError
       # if the error file already exists we don't take any additional action here
       unless File.exist?(error_file)
         message_error = 'Failed to run chef recipe.'
-        node_type_to_log = {
-          'HeadNode' => 'Please check /var/log/chef-client.log in the head node, or check the chef-client.log in CloudWatch logs.',
-          'ComputeFleet' => 'Please check the cloud-init-output.log in CloudWatch logs.',
-        }
-        message_logs_to_check = node_type_to_log[node['cluster']['node_type']]
+        message_logs_to_check = \
+          'Please check /var/log/chef-client.log in the head node, or check the chef-client.log in CloudWatch logs.'
         message_troubleshooting_link = 'Please refer to'\
           ' https://docs.aws.amazon.com/parallelcluster/latest/ug/troubleshooting-v3.html#troubleshooting-v3-get-logs'\
           ' for more details on ParallelCluster logs.'
--- a/cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
+++ b/cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-# Cookbook:: aws-parallelcluster
+# Cookbook:: aws-parallelcluster-config
 #
 # Copyright:: 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
@@ -18,7 +18,7 @@ require 'chef/handler'
 
 module WriteChefError
   # this class is used to handle chef errors and write the errors into a certain file if the file does not exist yet
-  class WriteChefError < Chef::Handler
+  class WriteHeadNodeChefError < Chef::Handler
     def report
       extend Chef::Mixin::ShellOut
       error_file = node['cluster']['bootstrap_error_path']
--- a/cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
@@ -15,7 +15,11 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-chef_handler 'WriteChefError::WriteChefError' do
-  type exception: true
-  action :enable
+if node["cluster"]["node_type"] == "HeadNode"
+  chef_handler 'WriteChefError::WriteChefError' do
+    type exception: true
+    action :enable
+  end
 end
+
+include_recipe "aws-parallelcluster-slurm::enable_chef_error_handler" if node["cluster"]["scheduler"] == "slurm"
--- a/cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
+++ b/cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-# Cookbook:: aws-parallelcluster
+# Cookbook:: aws-parallelcluster-config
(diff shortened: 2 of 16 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,21 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+chef_handler 'WriteChefError::SlurmChefHandler' do
```
chenwany: Question: It is not clear to me why do we only add the handler under slurm recipes, what if the failure happen outside Slurm recipes?

cookbooks/aws-parallelcluster-slurm/libraries/emit_chef_error_event.rb
```diff
@@ -0,0 +1,102 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+module WriteChefError
```
chenwany, line 17: What about combine it with  `\aws-parallelcluster-cookbook/libraries/write_chef_error_handler.rb` and move it out of slurm recipes? 
we can check `if node["cluster"]["node_type"]` in `def report` and have different implementation based on headnode and compute nodes.
Alternatively we can enable different handlers based on different node types

cookbooks/aws-parallelcluster-slurm/recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+# Recipe:: enable_chef_error_handler
+#
+# Copyright:: 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
chenwany: nit:2023?

cookbooks/aws-parallelcluster-slurm/recipes/enable_chef_error_handler.rb
delongmeng: This handler is designed to be slurm-specific, because the error json file contains a lot of slurm-specific information (such as slurm node name, etc). So it should only be registered when the scheduler is slurm. As long as the sheduler is slurm and this handler gets registered, it will handle any error during the run, regardless of where the failure comes from, Slurm recipes, or other recipes.

cookbooks/aws-parallelcluster-slurm/recipes/enable_chef_error_handler.rb
delongmeng: good catch. fixed

cookbooks/aws-parallelcluster-slurm/libraries/emit_chef_error_event.rb
delongmeng: After discussion, we decide to keep these two handlers separate, but use the same recipe to enable these handlers according to the node type and/or scheduler type. Have updated the code according to this logic.

cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
```diff
@@ -14,6 +14,8 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'chef/handler'
```
chenwany, line 17: nit: Cookbook name at line 4 need to be changed or remove

cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
```
chenwany: nit: change the cookbook name here

cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+if node["cluster"]["node_type"] == "HeadNode"
+  chef_handler 'WriteChefError::WriteChefError' do
+    type exception: true
+    action :enable
+  end
+end
+
+include_recipe "aws-parallelcluster-slurm::enable_chef_error_handler" if node["cluster"]["scheduler"] == "slurm"
```
chenwany, line 25: What about 
rename the WriteChefError to WriteHeadNodeChefError

same for emit_chef_error_events --> WriteHComputeFleetChefError

What about:
And remove if node["cluster"]["node_type"] == "ComputeFleet" in slurm recipe?
```
if node["cluster"]["node_type"] == "HeadNode"
  chef_handler 'WriteChefError::WriteHeadNodeChefError' do
    type exception: true
    action :enable
  end
elsif node["cluster"]["scheduler"] == "slurm" && node["cluster"]["node_type"] == "ComputeFleet"
  include_recipe "aws-parallelcluster-slurm::enable_chef_error_handler"
end
```

cookbooks/aws-parallelcluster-config/libraries/write_chef_error_handler.rb
delongmeng: good catch; will update accordingly

cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
delongmeng: good catch; will update accordingly

cookbooks/aws-parallelcluster-config/recipes/enable_chef_error_handler.rb
delongmeng: I agree with changing names of these handlers.
`WriteChefError::WriteChefError` -> `WriteChefError::WriteHeadNodeChefError`
`WriteChefError::SlurmChefHandler` -> `WriteChefError::WriteComputeFleetSlurmChefError`

In my opinion:
- Although now we only have one slurm handler (the compute node handler), maybe in the future we can have another slurm handler, regardless of node type
- In this case, instead of hard code it in config (checking the node type), we may want to let slurm to make the decision on when and which slurm handler(s) should be enabled 
- This `aws-parallelcluster-slurm::enable_chef_error_handler` recipe will do exactly this job, so we should have the condition check logic within this recipe

chenwany (approved):

PR #1899, opened 2023-03-15
Prepare kitchen configuration to be used by kitchen tests running on Jenkins

awslitvit:
### Description of changes
* Prepare kitchen configuration to be used by kitchen tests running on Jenkins

### Tests
* Tested locally on EC2

### Checklist

### DIFF
--- a/kitchen.ec2.sh
+++ b/kitchen.ec2.sh
@@ -11,11 +11,17 @@
 # KITCHEN_KEY_NAME:           KeyPair to use with EC2 instances
 #                             default kitchen
 #
-# KITCHEN_SUBNET_ID:          subnet-0c8a5dd6c753bec3f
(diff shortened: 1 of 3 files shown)


francesco-giordano (approved):

kitchen.ec2.sh
```diff
@@ -63,7 +69,7 @@ then
   source "${THIS_DIR}/.kitchen.env.sh"
 fi
 
-if [ "$1" != "list" ]; then
+if [ "$1" == "create" ] || [ "$1" == "converge" ] || [ "$1" == "verify" ] || [ "$1" == "destroy" ] || [ "$1" == "test" ]; then
```
francesco-giordano, line 72: nitpick: this does not work when a parameters are used as regex like `cr` instead of `create`

awslitvit: Correct, thanks for spotting it.

PR #1907, opened 2023-03-17
Refactor kitchen config file for Jenkins

awslitvit:
### Description of changes
Refactor kitchen config file for Jenkins to avoid repeating code.

### Tests
* Launched manually on EC2

### Checklist

### DIFF
--- a/kitchen.validate-config.yml
+++ b/kitchen.validate-config.yml
@@ -1,0 +1,88 @@
+# Validates config recipes
+---
+_common_cluster_attributes: &_common_cluster_attributes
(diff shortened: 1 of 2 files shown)


kitchen.validate-config.yml
psacc, line 1: this is way more DRY than before, missing the reason why it's in a new file, may be missing some previous PRs on this topic

psacc (approved):

kitchen.validate-config.yml
awslitvit: This new file name is consistent with naming conventions of other config files.
However, I couldn't just rename the old config file because still used by Jenkins.
I will delete it later.

PR #1908, opened 2023-03-17
Avoid interacting with S3 and DDB in kitchen tests

awslitvit:
### Description of changes
By default, when kitchen testing, `cluster-config.yml` and `instance-types-data.json` will be loaded from kitchen data path.
To force reading from S3, set `interact_with_s3 attribute` to `true`.

Also, by default writing on DynamoDB will be skipped when running Kitchen tests. To force writing to DynamoDB, set `interact_with_ddb attribute` to `true`.

### Tests
* ChefSpec test
* Manually tested on EC2

### Checklist


psacc (dismissed):
system_tests/cluster-config.yaml could be then removed adopting the same approach for system_tests

psacc (approved):

PR #1955, opened 2023-03-27
[develop] Refactor code for slurm configs to prepare it for custom slurm settings 

NSsirena:
### Description of changes
* Move some functions used to retrieve ComputeResource data from a given config
* Add a class responsible of rendering the ComputeResource config in different contexts 
* Add a class responsible of rendering the Queue config in different context

### Tests
* Unit tests successfully launched
* small set of integration tests successfully launched 

### Note
Given an inconsistency between unit and integ tests configurations changes that were meant to be in different files were moved back into a single file until the inconsistency is fixed.

### References
* [CLI Changes](https://github.com/aws/aws-parallelcluster/pull/5079)

### Checklist


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+                    f"Type={ self.gpu_type } File=/dev/nvidia[0-{ self.gpu_count - 1 }]\n"
+                )
+
+        return gres_render
+
+    def _definitions(self, dynamic=False):
+        definitions = f" CPUs={ self.vcpus() } RealMemory={ self.real_memory } State=CLOUD {self._features(dynamic)}"
+
+        if self.has_gpu and self.gpu_count > 0:
+            definitions += f" Gres=gpu:{ self.gpu_type }:{self.gpu_count}"
```
jdeamicis: Nitpick: the spacing between for the `{ }` is a bit inconsistent throughout the file (I'm surprised the linters are not complaining about it).

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+            self.vcpus_count
+            if not self.disable_simultaneous_multithreading
+            else (self.vcpus_count // self.threads_per_core)
+        )
+
+    def gpus(self) -> dict:
+        """Return the number of GPUs and type for the compute resource."""
+        return {"count": self.gpu_count, "type": self.gpu_type}
+
+    def render(self):
```
jdeamicis: Could we call the render functions in a similar manner? Maybe something like this?
- `render_as_node()`;
- `render_as_nodeset()`;
- `render_as_gres_conf()`;

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+
+class QueueRenderer:
+    """Renders a PCluster Queue Config as a Slurm partition config or as gres config."""
+
+    def __init__(self, queue_config, no_gpu, memory_ratio, instance_types_data, conf_type="partition", default=False):
+        self.name = queue_config["Name"]
+        self.is_default = default
+        self.conf_type = conf_type
+        self.capacity_type = queue_config["CapacityType"]
+        self.iam = queue_config["Iam"]
```
jdeamicis: I assume we can remove this, right?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+        """Launch the rendering process."""
+        config = ""
+        if self.static_nodes > 0:
+            config += f"NodeName={self.static_node_name()}{ self._definitions() }\n"
+        if self.dynamic_nodes > 0:
+            config += f"NodeName={self.dynamic_node_name()}{ self._definitions(dynamic=True) }\n"
+
+        return config
+
+    def as_node_set(self):
```
jdeamicis: Why don't we split the NodeSet as a trivial separate class? Maybe we don't need it now, but in the long run it may be cleaner.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+        self.capacity_type = queue_config["CapacityType"]
+        self.iam = queue_config["Iam"]
+        self.compute_renderers = [
+            ComputeResourceRenderer(self.name, compute_resource_config, no_gpu, memory_ratio, instance_types_data)
+            for compute_resource_config in queue_config["ComputeResources"]
+        ]
+
+    def render(self):
+        """Launch the rendering of the required configuration."""
+        rendered = f"{CONFIG_HEADER}\n"
```
jdeamicis: Nitpick: could we make the internal variable name consistent with the `render()` method of the `ComputeResourceRenderer` class?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+            gres_render += f"{renderer.as_gres()}"
+
+        return gres_render
+
+    def _render_node_set(self):
+        node_set = f"NodeSet={self.name}_nodes Nodes="
+        for renderer in self.compute_renderers:
+            node_set += f"{renderer.as_node_set()}"
+
+        return node_set[:-1]
```
jdeamicis: You can avoid the need for this `[:-1]` by implementing `ComputeResourceRenderer.as_node_set()` using the `",".join()` string method instead of using the `+=` operator to join two strings.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -17,18 +17,293 @@
...
+        return [compute_resource_config["InstanceType"]]
+
+
+def _get_efa_settings(compute_resource_config) -> Tuple[bool, bool]:
+    """Return a Tuple with EFA flags."""
+    if "Efa" in compute_resource_config:
+        efa = compute_resource_config["Efa"]
+        return efa["Enabled"], efa["GdrSupport"]
+    else:
+        return False, False
```
jdeamicis, line 54: Let's create an internal NTH task to add a unit test to cover the `else:` case. There are other occurrences where this is the case apparently (see below).

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -17,18 +17,293 @@
...
+
+    def _render_partition(self):
+        partition = f"PartitionName={self.name} Nodes={self.name}_nodes MaxTime=INFINITE State=UP"
+        if self.is_default:
+            partition += " Default=YES"
+
+        return partition
+
+
+# end of config_renderer.py
```
jdeamicis, line 306: I guess this line should be removed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -17,18 +17,293 @@
...
+
+    def render(self):
+        """Launch the rendering of the required configuration."""
+        rendered = f"{CONFIG_HEADER}\n"
+        if self.conf_type == "gres":
+            rendered += self._as_gres()
+        else:
+            rendered += self._as_partition()
+
+        return rendered
```
jdeamicis: OK, this is a bit confusing at first although the code is completely fine.

In `ComputeResourceRenderer`, the `render()` method renders the node definition (compute resource) line.

In `QueueRenderer` instead the `render()` method renders the whole Slurm include file (so all the node definitions, nodeset and partition).

Could we maybe call them in two different ways? Maybe call this more general method in `QueueRenderer` something like `render_include_file()`?

Also, maybe the `_as_partition()` and `_as_gres()` method could be renamed after the portion of the Slurm config they refer to (e.g. `_for_slurm_conf()` and `_for_gres_conf`)?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -17,18 +17,293 @@
...
+        return partition_render
+
+    def _as_gres(self):
+        gres_render = ""
+        for renderer in self.compute_renderers:
+            gres_render += f"{renderer.as_gres()}"
+
+        return gres_render
+
+    def _render_node_set(self):
```
jdeamicis: Nitpick: I believe we should use `nodeset` instead of `node_set` just because this way it's easier to find all info about nodesets with a single case-insensitive search (otherwise we would need a regex that may include the underscore).

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
jdeamicis: It might be a good idea to switch to `join()` in all places we are currently using `+=`.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def static_node_name(self):
+        """Render the NodeName section for static nodes."""
+        return self._node_name("st", self.static_nodes)
+
+    def dynamic_node_name(self):
+        """Render the NodeName section for dynamic nodes."""
+        return self._node_name("dy", self.dynamic_nodes)
+
+    def vcpus(self) -> int:
```
jdeamicis: Nitpick: I guess some of these methods should be prefixed with an underscore.

jdeamicis (commented):
I left some comments around mainly aimed at making the code as readable as possible.

Watch out, some of the comments I left are inlined in the [first commit](https://github.com/aws/aws-parallelcluster-cookbook/pull/1955/commits/a7d65f4bf6d225a7397a8940ddcbd697ded72926).

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: :) ..you're right .
I'll fix it.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: sure

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: The reasoning behind this decision is that it would require to duplicate the list of ComputeResource objects or at least to share the same references with the Queue class.
And the advantage would be to move a single method.
I think we aren't there yet.
If/when there will be a richer behavior than just formatting a string we will add it.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: sure

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: I added this case in the following PR to be sure it was covered.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: You mean the comment ?

I added it to make it clear which portion of the code should be moved into a dedicated file once the issue with the path resolution is fixed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: Yep. 
As discussed I'll try to rename these methods to make it more clear what they are doing and keep them consistent across both classes.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: Good catch.
I agree and I'll fix it

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Yes. 
I agree. I created when I wasn't yet really sure where/how they will be used. But they are used only internally, so I'll leave as public only the `render_` methods.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Done

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Done

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: I also think so.
I added them since they were read before and I expected to need them for some tests, but since the aren't yet used we can surely remove them.

NSsirena: Done

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Done

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Done

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: Should be more clear now

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Left public only the required methods

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
NSsirena: Done.

Not sure it really improves the code readability.

jdeamicis (approved):

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -17,18 +17,294 @@
...
+
+    def render_as_nodeset_element(self) -> List[str]:
+        """Alternative rendering for the NodeSet definition."""
+        node_set = []
+        if self.static_nodes > 0:
+            node_set.append(f"{self._static_node_name()}")
+        if self.dynamic_nodes > 0:
+            node_set.append(f"{self._dynamic_node_name()}")
+
+        return node_set
```
jdeamicis, line 180: Nitpick: `nodeset` instead of `node_set`.

PR #1958, opened 2023-03-27
Render CustomSlurmSettings in cookbook 

jdeamicis:
### Description of changes
* Render CustomSlurmSettings in cookbook
  * create include file infrastructure for user-controlled custom Slurm settings defined in the cluster config YAML;
  * add possibility to update the bulk `CustomSlurmSettings` in the cookbook;
  * create unit test for the renderization of the bulk custom Slurm settings;

### Tests
* Ran manual tests with:
  * created one cluster with `CustomSlurmSettings`; 
  * started with cluster with no CustomSlurmSettings (to validate that normal cluster creation is possible without the added functionality) and then:
    * updated the cluster with `CustomSlurmSettings`;
    * updated the cluster modifying `CustomSlurmSettings`;
    * updated the cluster removing `ClusterSlurmSettings`;
* Added unit tests to cover the renderization of the custom Slurm settings from the `CustomSlurmSettings` coming from the YAML file.
* Ran round of selected integration tests with the development branch used here.

### References
* PR from the CLI side: https://github.com/aws/aws-parallelcluster/pull/5103

### Checklist
- [ ] ~Check if documentation is impacted by this change.~ -> Pending later


libraries/helpers.rb
```diff
@@ -310,6 +310,14 @@ def are_queues_updated?
   config["Scheduling"] != previous_config["Scheduling"] or is_compute_node_bootstrap_timeout_updated?(previous_config, config)
 end
 
+# Verify if CustomSlurmSettings has been updated in the config
```
NSsirena, line 313: We should have tests for these functions.
They should be added as part of the review of the recipes, but let's try to add them opportunistically.

Let's open a nice-to-have task to add some of them before closing the story.

libraries/helpers.rb
```diff
@@ -310,6 +310,14 @@ def are_queues_updated?
...
+def are_custom_slurm_settings_updated?
```
NSsirena: Since this function is monitoring only the `bulk` custom settings, let's be explicit about this and rename the function `are_bulk_custom_slurm_settings_updated?`

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
NSsirena: 2023 ?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+
+def _generate_simple_parameter(param: Dict) -> str:
+    """
+    Generate string to represent a simple parameter in the Slurm configuration.
+
+    :param param: dictionary containing the parameter structure.
+    :return: string to be included in the Slurm configuration file.
+    """
+    key = list(param.keys())[0]
+    value = list(param.values())[0]
```
NSsirena: What if the dictionary has more than one item ?

We should at least check this condition and add a log line to track it.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    # Some complex parameters require a specific subparameter to be defined first in the line
+    # (e.g. the NodeName parameter in the node definition)
+    for key, value in param.items():
+        # Here we assume that the users do not pass more than one of these parameters in the same dictionary.
+        if key.lower() in ["nodename", "partitionname", "nodeset", "downnodes"]:
+            first_subparam = {key: value}
+            break
+
+    remaining_subparams = {key: value for key, value in param.items() if key not in first_subparam}
```
NSsirena, line 64: With the current code this will fail if the dictionary doesn't contains at least one of the keys listed at line `66` since `first_subparam` will not be initialized.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+    Generate string to represent a complex parameter in the Slurm configuration.
+
+    :param param: dictionary containing the parameter structure.
+    :return: string to be included in the Slurm configuration file.
+    """
+    # Some complex parameters require a specific subparameter to be defined first in the line
+    # (e.g. the NodeName parameter in the node definition)
+    for key, value in param.items():
+        # Here we assume that the users do not pass more than one of these parameters in the same dictionary.
+        if key.lower() in ["nodename", "partitionname", "nodeset", "downnodes"]:
```
NSsirena, line 56: We should check that no more than one of these keys is present or fail.

Also, unless you want to make this function more general and handle both cases you should also check that at least one is present.

On the other and, if you initialize `first_subparam` with `None` and then skip the first `for` loop below if `first_subparam is None` you should be able to handle both cases with this function and remove the previous function.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+        return yaml.load(input_file, Loader=yaml.SafeLoader)
+
+
+def _setup_logger():
+    logging.basicConfig(
+        level=logging.INFO, format="%(asctime)s - [%(name)s:%(funcName)s] - %(levelname)s - %(message)s"
+    )
+
+
+def _generate_simple_parameter(param: Dict) -> str:
```
NSsirena: While I understand from where it comes from, I really do not like this method. 
A method that assumes a dictionary has only a single key/value pair is confusing.

You should be able to handle this case within the boundaries of the following method when
* no `first_subparam` is passed
* only a key/value pair is present.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+
+    :param param: dictionary containing the parameter structure.
+    :return: string to be included in the Slurm configuration file.
+    """
+    key = list(param.keys())[0]
+    value = list(param.values())[0]
+    return f"{key}={value}"
+
+
+def _generate_complex_parameter(param: Dict) -> str:
```
NSsirena: probably a generic `_render_parameter` would be better (provided that both cases are handled with the same function.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+    """Generate Custom Slurm Settings include files for the Slurm configuration."""
+    # Make output directories
+    output_directory = path.abspath(output_directory)
+    pcluster_subdirectory = path.join(output_directory, "pcluster")
+    makedirs(pcluster_subdirectory, exist_ok=True)
+
+    # Load cluster configuration from YAML
+    cluster_config = _load_cluster_config(input_file)
+
+    for include_target in ["slurm"]:
```
NSsirena: Why this generalization ?

If we do not have an immediate need I'd keep it simple and use 

```
include_file_config = cluster_config["Scheduling"]["SlurmSettings"].get("CustomSlurmSettings", {} )
```

and also remove the `include_target` parameter.

Also since these files are `slurm` specific, do we expect to have other targets in the future ?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        if include_file_config:
+            _generate_include_file(
+                include_target,
+                include_file_config,
+                pcluster_subdirectory,
+                dryrun,
+            )
+
+    log.info("Finished.")
```
NSsirena: As is, it is not very informative :).

Let's add a more descriptive message like `Definition of Custom Slurm Settings completed.`

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Addressing in next commit.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: You mean having an explicit assertion during execution?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: OK, I see what you mean. Addressing in the next commit.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: > We should check that no more than one of these keys is present or fail.

I'm not sure I agree with that.

I mean, I see the problem, but that would be a sort of validation performed at cookbook time: we should either do it in the CLI at validator time, or not do it at all since a main assumption in the feature is that we are not going to validate the Slurm config coming from the user, other than checking against the deny-list. Let's discuss about this.

> Also, unless you want to make this function more general and handle both cases you should also check that at least one is present.

Yes, I'm going to handle both cases with the same method. Plus I wonder if there are other configuration parameters that may be composed of space-separated key=value pairs that are not covered by that list...

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Addressing in the next commit.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: I was thinking about future extensions where we allow setting `CustomSlurmDbdSettings`, `CustomSlurmCgroupSettings`, etc.

If we don't want to do this now I can remove it. Let's discuss about this.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: It's inherited from the script I used as a blueprint for this one :)

Consider that this would appear after the following line:
```python
log.info("Generating custom_slurm_settings_include_file_%s.conf", include_target)
```

libraries/helpers.rb
jdeamicis: I think this is part of some broader story the QA team is working on.

libraries/helpers.rb
jdeamicis: Addressing.

jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Removed this method.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Removed method.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Done.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: OK, we will put a warning if more than one of the special parameters in the list are provided.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: We decided to remove it.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    first_subparam = {}
+
+    # Some complex parameters require a specific subparameter to be defined first in the line
+    # (e.g. the NodeName parameter in the node definition)
+    first_subparam_found = False
+    for key, value in param.items():
+        # Here we assume that the users do not pass more than one of these parameters in the same dictionary.
+        if key.lower() in ["nodename", "partitionname", "nodeset", "downnodes"]:
+            if not first_subparam_found:
```
NSsirena: [Nitpick] Here we could write 

`if !first_subparam:` and avoid using a bool flag.

NSsirena (approved):

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_custom_slurm_settings_include_file_generator.py
jdeamicis: Done.

PR #1962, opened 2023-03-29
[develop] Add Slurm CustomSettings renderer for queue and ComputeResource

NSsirena:
### Description of changes
* Add logic for adding custom settings for queue and compute resources configs
* Add unit test for custom settings on queue and compute resources

### Tests
* Unit test executed locally.

### References
* [Code refactoring](https://github.com/aws/aws-parallelcluster-cookbook/pull/1955)

### Checklist


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -148,7 +148,8 @@ def __init__(self, queue_name: str, compute_resource_config: Dict, no_gpu, memor
         self.static_nodes = compute_resource_config["MinCount"]
         self.dynamic_nodes = compute_resource_config["MaxCount"] - self.static_nodes
         self.name = compute_resource_config["Name"]
-        self.disable_simultaneous_multithreading = compute_resource_config["DisableSimultaneousMultithreading"]
+        self.disable_multithreading = compute_resource_config.get("DisableSimultaneousMultithreading", False)
```
jdeamicis: If this was implemented this way, it most likely means that `DisableSimultaneousMultithreading` is always defined (and you can confirm it by checking the rendered config file on the cluster).

The parameters that require a default value are the ones that in the CLI get a default value of `None`.

By the way, I see from the CLI that in the end we kept the `CustomSlurmSettings` at compute resource and queue level with a default value of `{}` (see [here](https://github.com/aws/aws-parallelcluster/blob/e1bf25108cc298224c67776b62f81189cbe12313/cli/src/pcluster/config/cluster_config.py#L2186) and [here](https://github.com/aws/aws-parallelcluster/blob/e1bf25108cc298224c67776b62f81189cbe12313/cli/src/pcluster/config/cluster_config.py#L1885)). Was this intended? I thought we decided to keep it `None`.

I mean, it's not a big problem, but it means we may end up in the future with a slightly different logic for handling that parameter at validation time or during the execution of the recipes (for example, because of this the line below this one does not really need the default `{}` in `.get("CustomSlurmSettings", {}`).

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -163,21 +164,23 @@ def render_as_nodename(self):
         """Launch the rendering process."""
         config = ""
         if self.static_nodes > 0:
-            config += f"NodeName={self._static_node_name()}{self._definitions()}\n"
+            config += f"NodeName={self._static_node_name()}{self._definitions()}{self._custom_settings()}\n"
```
jdeamicis, line 164: I didn't notice it earlier: so the space that divides the entries on the line comes from the private method themselves?

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_input.yaml
```diff
@@ -0,0 +1,86 @@
+Scheduling:
```
jdeamicis, line 1: I would propose to make this file as minimal as possible in order to be able to test the rendering of the custom Slurm settings. Let's just define the queue names, the compute resources names and the custom Slurm settings (plus any parameter absolutely necessary for the rendering of the include files to succeed).

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_instance_types_data.json
```diff
@@ -0,0 +1 @@
...nstanceStorageInfo": {"TotalSizeInGB": 1800, "Disks": [{"SizeInGB": 900, "Count": 2, "Type": "ssd"}], "NvmeSupport": "required"}, "EbsInfo": {"EbsOptimizedSupport": "default", "EncryptionSupport": "supported", "EbsOptimizedInfo": {"BaselineBandwidthInMbps": 19000, "BaselineThroughputInMBps": 2375.0, "BaselineIops": 80000, "MaximumBandwidthInMbps": 19000, "MaximumThroughputInMBps": 2375.0, "MaximumIops": 80000}, "NvmeSupport": "required"}, "NetworkInfo": {"NetworkPerformance": "100 Gigabit", "MaximumNetworkInterfaces": 15, "MaximumNetworkCards": 1, "DefaultNetworkCardIndex": 0, "NetworkCards": [{"NetworkCardIndex": 0, "NetworkPerformance": "100 Gigabit", "MaximumNetworkInterfaces": 15}], "Ipv4AddressesPerInterface": 50, "Ipv6AddressesPerInterface": 50, "Ipv6Supported": true, "EnaSupport": "required", "EfaSupported": true}, "GpuInfo": {"Gpus": [{"Name": "V100", "Manufacturer": "NVIDIA", "Count": 8, "MemoryInfo": {"SizeInMiB": 32768}}], "TotalGpuMemoryInMiB": 262144}, "PlacementGroupInfo": {"SupportedStrategies": ["cluster", "partition", "spread"]}, "HibernationSupported": false, "BurstablePerformanceSupported": false, "DedicatedHostsSupported": false, "AutoRecoverySupported": false}}
```
jdeamicis, line 1: Let's keep here only the instance types we are going to keep in the minimal test yaml file.

Also watch out, this file does not end with a newline character.

jdeamicis (commented):
Overall great job! I just left some comments mainly to make the unit tests as simple as possible.

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_input.yaml
```diff
@@ -0,0 +1,86 @@
...
+            CustomEfaComputeParam: unique_efa_compute_value
+            OtherEfaComputeParam: second_efa_compute_value
+          DisableSimultaneousMultithreading: true
+          Efa:
+            Enabled: true
+            GdrSupport: false
+          InstanceType: c5n.18xlarge
+          MaxCount: 5
+          MinCount: 1
+
```
jdeamicis: Nitpick: there's a hole here.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: Yes.
Each field is responsible to provide its own piece of text, space included.
Otherwise you should have to add logic here to check for empty strings and add spaces only when needed.

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_input.yaml
NSsirena: While not strictly required for testing our changes, this portion of the config is required by the function `generate_slurm_config_files` that still process the config to generate the `slurmbdb` config.

Since this method is shared with other tests I prefer not to change it.

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_input.yaml
NSsirena: fixed

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_instance_types_data.json
NSsirena: I copied the file from another test so I assume it is valid.

We should reuse the same file across multiple test instead of copying it around, but this is out-of-scope for this PR.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
NSsirena: > we kept the CustomSlurmSettings at compute resource and queue level with a default value of {}

Since It was already in the code, replacing it with `None` wouldn't change the logic or add anything else, so I kept it as it was.

And also here: this `.get("CustomSlurmSettings", {})` this ensure we fall back to the same default.

test/unit/slurm/test_slurm_config_generator/test_generate_slurm_config_with_custom_settings/sample_input.yaml
jdeamicis: I was talking about the test input yaml, not about any logic in the Python code.

jdeamicis (approved):

PR #1979, opened 2023-04-05
[develop] Fix python path for slurm config unit test in cookbook

NSsirena:
### Description of changes
* Align tox python path for slurm unit tests to the one used by the recipes
* Extract common functions in `config_utils.py`
* Extract Queue and ComputResource Renderers in `config_renderer.py`
* Review logic to identify jinja template folder

### Tests
* All unit tests launched from the IDE and with tox are passing

### References
* [Code refactoring](https://github.com/aws/aws-parallelcluster-cookbook/pull/1955)
* [Custom Slurm Settings Renderer](https://github.com/aws/aws-parallelcluster-cookbook/pull/1962)

### Checklist


test/unit/slurm/test_slurm_config_generator.py
```diff
@@ -9,13 +9,19 @@
...
 import slurm
 from assertpy import assert_that
-from slurm.pcluster_slurm_config_generator import generate_slurm_config_files
+from pathlib import Path
+from pcluster_slurm_config_generator import generate_slurm_config_files
 
+def _get_template_folder() -> str:
+    module_file = inspect.getsourcefile(generate_slurm_config_files)
+    template_path = Path(module_file).parent / "templates"
+    return str(template_path)
```
jdeamicis: Is this a solid implementation to retrieve the templates folder? I'm asking because I honestly don't know.

Still, I guess it should be equivalent to the old `os.path.dirname(slurm.__file__) + "/templates"`

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_utils.py
```diff
@@ -0,0 +1,98 @@
+import math
+from typing import List, Tuple
+
+def _get_instance_types(compute_resource_config) -> List[str]:
```
jdeamicis: Nitpick (maybe addressed later?): these are now functions used by external scripts/classes, so let's drop the leading underscore.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_utils.py
```diff
@@ -1,6 +1,15 @@
+import inspect
 import math
+from pathlib import Path
 from typing import List, Tuple
 
+
+def get_template_folder() -> str:
```
jdeamicis, line 7: Maybe let's call this function something like `get_slurm_config_template_folder()`?

jdeamicis (commented):
Good job, I'm just leaving some comments to improve the readability of the code.

test/unit/slurm/test_slurm_config_generator.py
NSsirena: It is equivalent "by design" :)

It is also a solid implementation and it is more general since starts from the path of a module containing a given function. As of now since we only have one use I embedded the function name, and so the module.
However if other cookbooks need a similar function it could be moved in common and passing the function as parameter it would work also for other paths/modules/cookbooks if they are in the python_path of the code that is invoking the function.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_utils.py
NSsirena: (See comment above) I kept it general since the function can easily be generalized (by simply making the function name a parameter) and used in a more general context.
I didn't do this now, since we do not yet need it.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_utils.py
NSsirena: Good catch.
I fixed ` get_template_folder()` but I forgot the other ones since I moved them in bulk.

Fixing.

jdeamicis (approved):

PR #1990, opened 2023-04-07
fetch_and_run source /etc/parallelcluster/cfnconfig in a backward compatible way

psacc:
### Description of changes
introduces basic env variables management in fetch_and_run execution in order to support keeping cfnconfig defined variables available in scripts it executes with values compatible with previous versions of pcluster

### Tests
* added tests to cover env management

### References
*  https://github.com/aws/aws-parallelcluster-cookbook/pull/1895

### Checklist


cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
```diff
@@ -62,12 +62,115 @@ class ExecutableScript(ScriptDefinition):
...
+    ON_NODE_UPDATED = "postupdate"
+
+    def map_to_current_name(self):
+        """Return the current event name value as it's configured in the cluster config."""
+        if self == LegacyEventName.ON_NODE_START:
+            result = "OnNodeStart"
+        elif self == LegacyEventName.ON_NODE_CONFIGURED:
+            result = "OnNodeConfigured"
+        elif self == LegacyEventName.ON_NODE_UPDATED:
+            result = "OnNodeUpdated"
```
davprat: Would it make sense to convert this to a simple dictionary lookup? I'm honestly not sure which is cleaner though...

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
```diff
@@ -62,12 +62,115 @@ class ExecutableScript(ScriptDefinition):
...
+        return self._base_env.copy()
+
+
+class CfnConfigEnvEnricher(EnvEnricher):
+    """Provides cnfconfig backward compatible environment variables for the ExecutableScript."""
+
+    def __init__(self, conf: CustomActionsConfig):
+        super().__init__()
+        self.conf = conf
+        self._cfn_base_env = self._create_additional_cfnconfig_compatible_env()
```
davprat: I'm not sure, but I think I've seen a similar pattern get flagged by the linters (but I could be wrong). Would it make sense to make both of the below calls static and pass self.conf as a parameter. Just to make sure there are no order of operation issues?

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
```diff
@@ -164,7 +267,12 @@ async def _execute_script(self, exe_script: ExecutableScript):
         try:
             # arguments are provided by the user who has the privilege to create/update the cluster
             subprocess.run(
-                [exe_script.path] + (exe_script.args or []), check=True, stderr=subprocess.PIPE
+                [exe_script.path] + (exe_script.args or []),
+                check=True,
+                stderr=subprocess.PIPE,
+                stdout=stdout,
+                text=True,
+                env=self._env_enricher.build_env(exe_script),
```
davprat, line 283: Would it be possible to isolate the calls to subprocess.run to a single method so that we only need to #nosec in a single place? Though maybe this isn't a great suggestion as it could obfuscate the security risk, so I'm not sure.

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
```diff
@@ -514,6 +562,26 @@ def load_configuration(self, args) -> CustomActionsConfig:
...
 
+    def _deserialize_script_sequences(self, cluster_config, event_name, node_type, queue_name):
+        try:
+            if node_type == "HeadNode":
+                data = cluster_config["HeadNode"]["CustomActions"][event_name]
+            else:
+                data = next(
+                    (
+                        q
+                        for q in next(v for v in cluster_config["Scheduling"].values() if isinstance(v, list))
```
davprat: Not a big deal, but could q and v be renamed to something more descriptive?

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
psacc: I do not agree, the different calls have a different of parameters and also the security risk changes, only this one accepts external inputs while putting the two together will make this less obvious IMHO

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
psacc: You are right and I don't like it either. Let's see what I can do about it.

cookbooks/aws-parallelcluster-config/files/default/custom_action_executor/custom_action_executor.py
psacc: It was like this because it came from the bash port... in python I agree it better fits with dict lookup.

davprat (approved):
lgtm!

PR #1995, opened 2023-04-11
[Fix] Make supervisord manage the cfn-hup program in no-daemon mode

gmarciani:
### Description of changes
Make supervisord manage the cfn-hup program in no-daemon mode to avoid systemic failures when the program gets restarted.

**Note** the 120 second period is the interval that we currently use when running cfn-hup in daemon mode, as per [this configuration](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/templates/cluster_stack.py#L1162-L1162)

### Tests
* Integration Tests: `test_update_slurm`
* Verified that cfn-hup is correctly restarted by supervisord when killed
* Verified that cfn-hup is correctly terminated when supervisord is stopped
* Verified that cfn-hup is correctly restarted when supervisord is restarted
* Verified that only one cfn-hup process is running at any given time
* Verified that no cfn-hup process is running when supervisord is stopped.


CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 **BUG FIXES**
 - Fix IP association on instances with multiple network cards.
 
+**BUG FIXES**
+- Fix issue causing cfn-hup daemon to fail when it gets restarted.
```
enrico-usai, line 34: Which kind of failures are we seeing? Is there a bug in cfn-hup?

cookbooks/aws-parallelcluster-install/templates/default/base/cfn-hup-runner.sh.erb
```diff
@@ -0,0 +1,14 @@
...
+
+# This script is used to run cfn-hup is a way that is suitable to be managed by supervisord.
+# In particular, cfn-hup is executed in no-daemon mode.
+#
+# Usage:   ./cfn-hup-runner.sh
+
+[ -f /etc/profile.d/proxy.sh ] && . /etc/profile.d/proxy.sh
+
+while true; do
+  <%= node['cluster']['cfn_bootstrap_virtualenv_path'] %>/bin/cfn-hup --no-daemon --verbose
```
enrico-usai, line 12: If I'm not wrong cfn-hup log messages are going to cfn* log files. Can you check if cfn logs are correctly populated (and uploaded on CloudWatch) after the restart and update of the cluster?

CHANGELOG.md
gmarciani: If cfn-hup is run in daemon mode and it is then killed, the following daemons restarted by supervisord systematically fail after ~min due to a dangling lock on the PID file. So if the cfn-hup daemon is killed, then the cluster will never be able to perform a cluster update.

I'll open an issue to the CloudFormation team on this.

cookbooks/aws-parallelcluster-install/templates/default/base/cfn-hup-runner.sh.erb
gmarciani: Confirmed: cfn logs are populated and sent to CloudWatch. 
Just to be precise: cfn-hup logs are currently not sent to CW, but only cloud-init ones

enrico-usai (approved):

PR #2017, opened 2023-04-18
Fix lustre client for RedHat 8.7

francesco-giordano:
### Description of changes
* Fix lustre client for RedHat 8.7 due to broken kABI compat in the kernel version - https://github.com/openzfs/zfs/issues/14724

### Tests
* Run first stage in personal Jenkins
* Run locally kitchen tests on ec2

### DIFF
--- a/cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
+++ b/cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
@@ -29,18 +29,30 @@ action :setup do
     log "FSx for Lustre is not supported in this RHEL version #{version}, supported versions are >= 8.2" do
       level :warn
     end
-  elsif node['cluster']['kernel_release'].include? "4.18.0-425.3.1.el8"
-    log "FSx for Lustre is not supported in kernel version 4.18.0-425.3.1.el8 of RHEL, please update the kernel version" do
-      level :warn
-    end
   else
     action_install_lustre
   end
 end
 
+def find_os_minor_version
+  os_minor_version = ''
+  kernel_patch_version = find_kernel_patch_version
+
+  # kernel patch version less than 193 is equivalent to a RHEL 8.2
+  os_minor_version = '2' if kernel_patch_version >= '193'
+  os_minor_version = '3' if kernel_patch_version >= '240'
+  os_minor_version = '4' if kernel_patch_version >= '305'
+  os_minor_version = '5' if kernel_patch_version >= '348'
+  os_minor_version = '6' if kernel_patch_version >= '372'
+  os_minor_version = '7' if kernel_patch_version >= '425'
+
+  os_minor_version
+end
+
 action_class do
   def base_url
-    "https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/$basearch"
+    # https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html#lustre-client-rhel
+    "https://fsx-lustre-client-repo.s3.amazonaws.com/el/8.#{find_os_minor_version}/$basearch"
   end
 
   def public_key
--- a/cookbooks/aws-parallelcluster-common/spec/unit/resources/lustre_setup_spec.rb
+++ b/cookbooks/aws-parallelcluster-common/spec/unit/resources/lustre_setup_spec.rb
@@ -160,6 +160,7 @@ describe 'lustre:setup' do
         step_into: ['lustre']
       ) do |node|
         node.automatic['platform_version'] = "8.1"
+        node.override['cluster']['kernel_release'] = "4.18.0-147.9.1.el8"
       end
       Lustre.setup(runner)
     end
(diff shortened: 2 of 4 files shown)


cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
```diff
@@ -39,8 +39,14 @@
...
 action_class do
+  def kernel_minor_patch
+    node['cluster']['kernel_release'].match(/^\d+\.\d+\.\d+-(\d+)\..*$/)[1]
+  end
+
   def base_url
-    "https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/$basearch"
+    # Special cases for kernel of redhat 8.7 - https://access.redhat.com/articles/3078#RHEL8
+    # specific repositories for 8.0, 8.1, ecc.. does not exist
+    "https://fsx-lustre-client-repo.s3.amazonaws.com/el/8#{kernel_minor_patch.to_i >= 425 ? '.7' : ''}/$basearch"
```
awslitvit: Would it make sense to add a ChefSpec test for this code?

cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
```diff
@@ -39,8 +39,14 @@
 end
 
 action_class do
+  def kernel_minor_patch
+    node['cluster']['kernel_release'].match(/^\d+\.\d+\.\d+-(\d+)\..*$/)[1]
```
lukeseawalker: why not using same approach done for centos7?
This change:
* requires further modification once new minor will be out, while the centos7 code only requires a new line to be added
* doesn't cover the case of minor version not supported, e.g. 8.3

francesco-giordano: > requires further modification once new minor will be out, while the centos7 code only requires a new line to be added
> doesn't cover the case of minor version not supported, e.g. 8.3

Since we are not pinning any minor version, the recommended approach from Fsx is to point to the plain 8 repo, but at the moment the correct package is only in the 8.7 repo and the other minor version repo does not exist.
Also with centos7 we would need to add a condition if a new minor is released. In any case, however, if a new minor not compatible with the 8.7 is released, I would expect the installation to fail.

francesco-giordano: Refactored and followed the step in the [public doc](https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html#lustre-client-rhel)
Thanks

cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
```diff
@@ -29,18 +29,32 @@
...
 end
 
+def find_centos_minor_version
+  os_minor_version = ''
+
+  # kernel release is in the form 4.18.0-425.3.1.el8
+  kernel_patch_version = node['cluster']['kernel_release'].match(/^\d+\.\d+\.\d+-(\d+)\..*$/)
+  raise "Unable to retrieve the kernel patch version from #{node['cluster']['kernel_release']}." unless kernel_patch_version
+
+  os_minor_version = '2' if kernel_patch_version[1] >= '193'
```
enrico-usai: We should raise an exception if os_minor version is 0 or 1 like we were doing for centos7

cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
```diff
@@ -29,18 +29,32 @@
...
-  elsif node['cluster']['kernel_release'].include? "4.18.0-425.3.1.el8"
-    log "FSx for Lustre is not supported in kernel version 4.18.0-425.3.1.el8 of RHEL, please update the kernel version" do
-      level :warn
-    end
   else
     action_install_lustre
   end
 end
 
+def find_centos_minor_version
```
enrico-usai: let's rename it to find_kernel_minor_version

awslitvit: Should it be `find_redhat8_minor_version` (or maybe `find_os_minor_version`)?

cookbooks/aws-parallelcluster-common/spec/unit/resources/lustre_setup_spec.rb
```diff
@@ -170,51 +170,35 @@ def self.setup(chef_run)
...
-  context "on redhat supporting lustre" do
-    cached(:chef_run) do
-      runner = ChefSpec::Runner.new(
-        platform: 'redhat', version: '8',
-        step_into: ['lustre']
-      ) do |node|
-        node.automatic['platform_version'] = "8.2"
-        node.override['cluster']['kernel_release'] = "supported"
+  [%w(193 2), %w(240 3), %w(305 4), %w(348 5), %w(372 6), %w(425 7)].each do |kernel_patch, minor_version|
+    context "on redhat with kernel greater equal than 4.18.0-#{kernel_patch}.3.1.el8 supporting lustre" do
```
awslitvit: Nitpick: maybe `since` or `from` instead of `greater equal than` ?

cookbooks/aws-parallelcluster-common/spec/unit/resources/lustre_setup_spec.rb
```diff
@@ -170,51 +170,35 @@ def self.setup(chef_run)
...
+      cached(:chef_run) do
+        runner = ChefSpec::Runner.new(
+          platform: 'redhat', version: '8',
+          step_into: ['lustre']
+        ) do |node|
+          node.automatic['platform_version'] = "8.2"
```
awslitvit: Is there any reason why it is `8.2` instead of `8.7`?

awslitvit (changes requested):

cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
francesco-giordano: In the previous implementation we were only printing a warning message letting the installation continue. I agree that can be better to raise an exception. Changing that the behaviour.

cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
francesco-giordano: Done

cookbooks/aws-parallelcluster-common/spec/unit/resources/lustre_setup_spec.rb
francesco-giordano: No need of the platform_version anymore, we only relay on the kernel version from https://access.redhat.com/articles/3078#RHEL8

enrico-usai (dismissed):
LGTM but I didn't look at ChefSpec

awslitvit (dismissed):

cookbooks/aws-parallelcluster-common/resources/lustre/lustre_redhat8.rb
```diff
@@ -29,18 +29,32 @@
...
-  elsif node['cluster']['kernel_release'].include? "4.18.0-425.3.1.el8"
-    log "FSx for Lustre is not supported in kernel version 4.18.0-425.3.1.el8 of RHEL, please update the kernel version" do
-      level :warn
-    end
   else
     action_install_lustre
   end
 end
 
+def find_os_minor_version
```
lukeseawalker, line 37: can we put this function in common for both rhel and centos? the versioning of the kernel is actual the same

enrico-usai (approved):

PR #2021, opened 2023-04-20
3.x Add publishing of GPU failure metrics

davprat:
### Description of changes
* This change adds generating structured log messages for GPU health check results.
* Default configuration only generates events if the health check fails.
* This change updates the CloudWatch Configuration to publish the health check event log to CloudWatch.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/files/default/config_slurm/scripts/health_check_manager.py
+++ b/cookbooks/aws-parallelcluster-slurm/files/default/config_slurm/scripts/health_check_manager.py
@@ -28,11 +28,18 @@ from logging.config import fileConfig
 from typing import List
 
 import yaml
+from event_utils import (
+    EventPublisher,
+    get_node_spec_from_file,
+    publish_health_check_exception,
+    publish_health_check_result,
+)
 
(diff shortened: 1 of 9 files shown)


cookbooks/aws-parallelcluster-slurm/files/default/config_slurm/scripts/health_check_manager.py
```diff
@@ -252,6 +261,9 @@ def _execute_health_checks(health_check_manager_config: HealthCheckManagerConfig
                 else:
                     output = " empty"
                 log.info("Output of Health Check '%s' execution is%s", health_check.name, output)
+                publish_health_check_result(
```
lukeseawalker, line 264: are you handling "publish_health_check_result" exceptions?
We need to be sure that a failure while pushing the metric doesn't generate a failure on the prolog, hence a node termination/replacement

davprat: I was protecting the inner base publish_event call, but not the outer call. I've updated the code so the outer call is guarded as well.

lukeseawalker (dismissed):

delongmeng (approved):

PR #2059, opened 2023-05-04
Check MySQL package version

awslitvit:
### Description of changes
AMI builder validation step verifies mysql packages version for every OS but Ubuntu.

Adding version check to InSpec control in order to replace current AMI validation with it.

### Tests
* Tested on EC2 for all OSes and architectures.

### Checklist

### DIFF
--- a/test/resources/controls/aws_parallelcluster_install/mysql_client_spec.rb
+++ b/test/resources/controls/aws_parallelcluster_install/mysql_client_spec.rb
@@ -16,20 +16,22 @@ control 'mysql_client_installed' do
   if os.redhat?
     mysql_packages.concat %w(mysql-community-client-plugins mysql-community-common
        mysql-community-devel mysql-community-libs mysql-community-libs-compat)
-  elsif os.debian?
-    if os.release == '18.04'
-      mysql_packages.concat %w(libmysqlclient-dev libmysqlclient20)
-    else
-      mysql_packages.concat %w(libmysqlclient-dev libmysqlclient21)
-    end
+  elsif os_properties.ubuntu1804?
+    mysql_packages.concat %w(libmysqlclient-dev libmysqlclient20)
+  elsif os_properties.ubuntu2004?
+    mysql_packages.concat %w(libmysqlclient-dev libmysqlclient21)
   else
     describe "unsupported OS" do
       pending "support for #{os.name}-#{os.release} needs to be implemented"
     end
   end
+
+  ubuntu = os_properties.ubuntu?
+
   mysql_packages.each do |pkg|
     describe package(pkg) do
       it { should be_installed }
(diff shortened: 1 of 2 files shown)


test/resources/controls/aws_parallelcluster_install/mysql_client_spec.rb
```diff
@@ -16,20 +16,22 @@
...
       pending "support for #{os.name}-#{os.release} needs to be implemented"
     end
   end
+
+  ubuntu = os_properties.ubuntu?
+
   mysql_packages.each do |pkg|
     describe package(pkg) do
       it { should be_installed }
+      its('version') { should match /^8.0.31-/ } unless ubuntu
```
ermanno, line 34: I think that this could be inlined:

```suggestion
      its('version') { should match /^8.0.31-/ } unless os_properties.ubuntu?
```

and as an aside, maybe we could even use the existing check on Debian family which we typically use to discriminate changes that apply to the ubuntu OSs:

```suggestion
      its('version') { should match /^8.0.31-/ } unless os_properties.debian_family?
```

ermanno (approved):

test/resources/controls/aws_parallelcluster_install/mysql_client_spec.rb
```diff
@@ -16,20 +16,22 @@
   if os.redhat?
```
jdeamicis, line 16: Nitpick: if we are using `os_properties.ubuntu?` below, maybe we should use `os_properties.redhat?` here too?

jdeamicis (approved):

test/resources/controls/aws_parallelcluster_install/mysql_client_spec.rb
awslitvit: Unfortunately, it can't: `os_properties` is not available in that context. This is why I used a variable.

I could use `os.debian?` instead of `os_properties.ubuntu?` since right now they are equivalent, however I preferred a narrower condition, which we can broaden in the future.

test/resources/controls/aws_parallelcluster_install/mysql_client_spec.rb
jdeamicis: No, that would be wrong: `os.redhat?` actually checks if the OS is a RHEL derivative, and is evaluated to true even for CentOS and Amazon Linux 2. So better to leave it as it is.

awslitvit: `os` is a native resource and it doesn't have `ubuntu?` [helper](https://docs.chef.io/inspec/resources/os/#osfamily-helpers), thus the difference. See the comment above wrt the choice between `ubuntu?` and `debian?`.

PR #2078, opened 2023-05-08
Skip kitchen test for root volume space on Centos7 x86

enrico-usai:
This case fails because we're installing IntelHPC stuff at runtime.


eantonin (approved):
LGTM

PR #2099, opened 2023-05-10
First draft Ubuntu 22 resources

awslitvit:
### Description of changes
First draft Ubuntu 22 resources.
We decided to avoid code duplication and to use one file to define Ubuntu flavor of a resource, when the code is the same for all Ubuntu versions.

### Tests
* Kitchen tests on Ubuntu22
* Kitchen test for `arm-pl` on Ubuntu20, since the code was changes

### Checklist


cookbooks/aws-parallelcluster-install/attributes/default_ubuntu22.rb
```diff
@@ -0,0 +1,24 @@
+# Ubuntu 20 default attributes for aws-parallelcluster-install
```
gmarciani, line 1: Minor: should be Ubuntu 22

gmarciani (approved):

PR #2115, opened 2023-05-11
Enable saving image after Kitchen test

awslitvit:
### Description of changes
Enable saving image after Kitchen test

### Tests
Tested both on Docker and on EC2

### Checklist

### DIFF
--- a/kitchen.docker.yml
+++ b/kitchen.docker.yml
@@ -23,45 +23,52 @@ transport:
 verifier:
   name: inspec
 
+lifecycle:
+  post_verify:
+  - local: |
+      [[ ${KITCHEN_SAVE_IMAGE} == 'true' ]] || exit 0
+      docker_id=$(docker ps -a | grep ${KITCHEN_INSTANCE_NAME} | awk '{ print $1 }')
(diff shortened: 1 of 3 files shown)


kitchen.docker.yml
```diff
@@ -23,45 +23,52 @@ transport:
 verifier:
   name: inspec
 
+lifecycle:
+  post_verify:
+  - local: |
+      [[ ${KITCHEN_SAVE_IMAGE} == 'true' ]] || exit 0
+      docker_id=$(docker ps -a | grep ${KITCHEN_INSTANCE_NAME} | awk '{ print $1 }')
+      docker commit ${docker_id} pcluster-${KITCHEN_PHASE}/${KITCHEN_INSTANCE_NAME}:latest
```
enrico-usai, line 31: q: is KITCHEN_PHASE a known env variable?

enrico-usai (approved):

kitchen.docker.yml
awslitvit: Yes, we set it in our helper script and we use it to understand which AMI to use (a bare one or a PCluster one).
This allows us to use a similar mechanism with any AMI we want to build on the fly.

PR #2130, opened 2023-05-15
Move openssh from install to platform cookbook

awslitvit:
### Description of changes
Move openssh from install to platform cookbook

### Tests
* Kitchen-tested on Docker and on EC2
* AMI successfully built with this code

### Checklist


cookbooks/aws-parallelcluster-tests/recipes/docker_mock.rb
```diff
@@ -0,0 +1,85 @@
+return unless on_docker?
+
+file '/bin/systemctl' do
+  content %{
+#!/bin/bash
+
+if [ "$1" == "--system" ]; then
```
gmarciani, line 7: What is the effect of this option?
I would have expected the setting of a variable to change some behaviour...

gmarciani (approved):

cookbooks/aws-parallelcluster-tests/recipes/docker_mock.rb
awslitvit: I guess this part is not used anymore. I will create a task to cleanup this script which is a mere copy/paste from another cookbook.

PR #2149, opened 2023-05-16
Add network service kitchen tests

dreambeyondorange:
The network service has two actions, restart and reload.  This commit adds a control to check if the service is running with a switch for different operating systems.  There are two additional controls that verify the service is running after the two different actions defined for the service.

### Tests
* Ran kitchen tests for all OSes

### Checklist


kitchen.resources-install.yml
```diff
@@ -296,3 +296,27 @@ suites:
...
+  - name: network_service_restart
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-common::test_resource]
+    verifier:
+      controls:
+        - network_service_running
+    attributes:
+      dependencies:
+        - resource:node_attributes
```
awslitvit, line 344: Nitpick: this is redundant

cookbooks/aws-parallelcluster-common/resources/network_service/partial/_network_service_debian.rb
```diff
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 
 action :reload do
+  return if virtualized?
```
awslitvit, line 17: Nitpick: please prefer `unless on_docker?` appended to the instruction you want to skip (in this case, to the  `end` of the execute command).

kitchen.resources-install.yml
```diff
@@ -296,3 +296,27 @@ suites:
...
+  - name: network_service_reload
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-common::test_resource]
+    verifier:
+      controls:
+        - network_service_running
+    attributes:
+      dependencies:
+        - resource:node_attributes
```
awslitvit, line 356: Nitpick: this is redundant.

awslitvit (approved):

cookbooks/aws-parallelcluster-common/resources/network_service/partial/_network_service_debian.rb
dreambeyondorange: I have to skip here because netplan does not exist on the ubuntu docker images and installing it gives other errors.  Or do you mean to use unless on_docker? for line 18?  I have seen return if virtualized used fairly frequently, so that is why I adopted this and haven't noticed unless on_docker outside of the test code.  Is there some documentation somewhere detailing design decisions like this that I can refer to?  Otherwise, there is really no way to know which way to do it and why

kitchen.resources-install.yml
dreambeyondorange: Ah right, I had this before I switched to the tests::setup.  I was following other tests I'd seen.  Overall this process has been quite confusing deciding which resources and run lists to use, since there seems to be alot of duplication across the tests

PR #2180, opened 2023-05-19
Move openssh from config to platform cookbook

himani2411:
### Description of changes
* Move openssh from config to platform cookbook

### Tests
* ```./kitchen.docker.sh platform-config verify ssh-target-checker -c 6```
* ```./kitchen.ec2.sh platform-config verify ssh-target-checker -c 6```

### Checklist


cookbooks/aws-parallelcluster-platform/metadata.rb
```diff
@@ -18,3 +18,5 @@
 depends 'selinux', '~> 6.0.5'
 
 depends 'aws-parallelcluster-shared', '~> 3.7.0'
+# TOFIX: remove this when all the common has been moved to shared
+depends 'aws-parallelcluster-common', '~> 3.7.0'
```
awslitvit: The dependency on common cookbook should not be added.

You should rather try and move attributes and/or functions you need away from common cookbook, as close as you can to the code part using it. It might be the recipe or resource itself if it's the only code part using the attribute or the function, or the cookbook if the usage is limited to it. In (hopefully, rare) cases when multiple cookbooks use that code, it should be moved to the shared cookbook. In your case, `get_vpc_cidr_list` function is used or going to be used both by platform and environment code, so it can be moved to shared cookbook.

cookbooks/aws-parallelcluster-platform/spec/unit/recipes/openssh_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      cached(:chef_run) do
+        runner(platform: platform, version: version).converge(described_recipe)
+      end
+      cached(:node) { chef_run.node }
+
+      before do
+        mock_get_vpc_cidr_list(vpc_cidr_list)
+      end
+
+      it 'installs the ssh target checker with the correct attributes' do
```
awslitvit, line 15: I would also check the file contents (see the documentation [here](https://github.com/chefspec/chefspec#render_file)). You can find some examples in ParallelCluster code as well (for instance, you might try to take a look at `isolated_install_spec.rb`).

cookbooks/aws-parallelcluster-shared/spec/spec_helper.rb
```diff
@@ -76,4 +76,8 @@ def expect_to_include_recipe_from_resource(recipe, cookbook = 'any')
   expect_any_instance_of(Chef::RunContext).to receive(:include_recipe).with(recipe, { current_cookbook: cookbook })
 end
 
+def mock_get_vpc_cidr_list(cidr)
+  allow_any_instance_of(Object).to receive(:get_vpc_cidr_list).and_return(cidr)
+end
```
awslitvit: Please avoid defining common functions when not strictly necessary: it might be confusing in the long run. In this case it's just one line of code, you can use it directly in your test instead of defining a function in the most shared code we have for ChefSpec tests.

kitchen.recipes-config.yml
```diff
@@ -77,10 +77,10 @@ suites:
         - networking_configured
   - name: ssh_target_checker
```
awslitvit: Please remove this test from here, as you already moved it to `platform-config` yaml.

cookbooks/aws-parallelcluster-platform/kitchen.platform-config.yml
```diff
@@ -16,3 +16,17 @@ suites:
       resource: chrony:enable
       dependencies:
       - resource:chrony:setup
+  - name: ssh_target_checker
+    run_list:
+      - recipe[aws-parallelcluster-platform::openssh]
+    verifier:
+      controls:
+        - /ssh_target_checker_script_created/
```
awslitvit: Nitpick: you don't need to use the regexp here,  as you need to run exactly that specific control. You can therefore remove the slashes around the control name.

awslitvit (changes requested):

cookbooks/aws-parallelcluster-platform/kitchen.platform-config.yml
himani2411: I have added one control for Docker and another for EC2 which is why I have used a Regex.

kitchen.recipes-config.yml
himani2411: Will do thanks.

cookbooks/aws-parallelcluster-shared/spec/spec_helper.rb
himani2411: I added this in Shared as this will later on be used by manage_raid_export_unexport_spec.rb and volumne_spec.rb. It can be reused by those tests too. I could change those here as well but wanted to keep the PR separate.

cookbooks/aws-parallelcluster-platform/spec/unit/recipes/openssh_spec.rb
himani2411: Sure. Will add this. Thanks for the file.

cookbooks/aws-parallelcluster-platform/test/controls/ssh_spec.rb
```diff
@@ -21,6 +22,18 @@
   end
 end
 
+control 'ssh_target_checker_script_created_ec2' do
```
awslitvit: Could you remove duplicated code and just add the condition "if on docker" to the line which works only on docker? Something like:

```
on_docker = os_properties.on_docker?
...
its('content') { should match /VPC_CIDR_LIST=\(cidr1 cidr2\)/ } if on_docker
```

cookbooks/aws-parallelcluster-platform/spec/unit/recipes/openssh_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          mode:  '0755',
+          variables: {
+            vpc_cidr_list: get_vpc_cidr_list,
+          }
+        )
+      end
+
+      it 'has the correct content' do
+        is_expected.to render_file("/usr/bin/ssh_target_checker.sh")
+          .with_content("VPC_CIDR_LIST=(#{get_vpc_cidr_list.join(' ')})")
```
awslitvit: It is preferable to avoid calling the function you mock to verify the test: you can just put the expected value here, or at least `vpc_cidr_list.join(' ')`.

Moreover, you can even avoid mocking the function if you override node attributes `ec2.mac` and `ec2.$mac_value. vpc_ipv4_cidr_blocks`. Since we don't have a separate test for the function, at least it would be tested here as well.

cookbooks/aws-parallelcluster-platform/metadata.rb
himani2411: Made changes w.r.t moving `get_vpc_cidr_list` to shared cookbook

cookbooks/aws-parallelcluster-platform/spec/unit/recipes/openssh_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      it 'installs the ssh target checker with the correct attributes' do
+        is_expected.to create_template('/usr/bin/ssh_target_checker.sh').with(
+          source: "openssh/ssh_target_checker.sh.erb",
+          owner: 'root',
+          group: 'root',
+          mode:  '0755',
+          variables: {
+            vpc_cidr_list: get_vpc_cidr_list,
```
awslitvit, line 22: Please use the variable instead of function.

cookbooks/aws-parallelcluster-shared/spec/spec_helper.rb
himani2411: Removed the function as I no longer need to mock.

cookbooks/aws-parallelcluster-platform/spec/unit/recipes/openssh_spec.rb
himani2411: Made changes. Thanks for the Help in resolving the errors!

dreambeyondorange (approved):

awslitvit (approved):

PR #2182, opened 2023-05-23
[develop] Automate startup of Colima when running docker

enrico-usai:
If Colima software is available, start it and export DOCKER_HOST variable

### Tests

Tested with:
* colima stopped
* colima running
* with another command in place of colima (to simulate colima not found case)

### References

* https://github.com/abiosoft/colima

### Checklist

### DIFF
--- a/kitchen.docker.sh
+++ b/kitchen.docker.sh
@@ -7,6 +7,12 @@ export KITCHEN_YAML="${THIS_DIR}/kitchen.docker.yml"
 export KITCHEN_GLOBAL_YAML="${THIS_DIR}/kitchen.global.yml"
 export KITCHEN_DRIVER=dokken
 
+# Export DOCKER_HOST variable when using Colima
+if [[ $(command -v colima) ]]; then
+  colima status || colima start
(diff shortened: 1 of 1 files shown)


kitchen.docker.sh
```diff
@@ -7,6 +7,12 @@ export KITCHEN_YAML="${THIS_DIR}/kitchen.docker.yml"
 export KITCHEN_GLOBAL_YAML="${THIS_DIR}/kitchen.global.yml"
 export KITCHEN_DRIVER=dokken
 
+# Export DOCKER_HOST variable when using Colima
+if [[ $(command -v colima) ]]; then
+  colima status || colima start
+  export DOCKER_HOST=unix:///Users/$(whoami)/.colima/default/docker.sock
```
lukeseawalker: What about using $HOME, so that it works cross platform?

lukeseawalker (approved):

kitchen.docker.sh
enrico-usai: Changed

PR #2185, opened 2023-05-24
Update the NFS resource for Ubuntu 22

dreambeyondorange:
Split the ubuntu NFS resource into three, one for each version of Ubuntu we support, 18, 20, and 22.  
Since the difference is with 20 and 18 will soon be removed. 

Ubuntu 22 uses a new config file for NFS see https://ubuntu.com/server/docs/service-nfs for details.  

Upgrading the nfs third party resource to 5.0.1+ resulted in failures similar to the known issue https://github.com/sous-chefs/nfs/issues/93.  

Setting the server template in the :setup action will preserve the default nfs.conf file and allow for us to override any values we need without removing the history of the config for users.  
Also created more useful names for the two different types of config files used for NFS.

### Description of changes
* Added a new template for Ubuntu 22 configs
* Split the ubuntu resource into 3 -18, 20, and 22+
* converted the options in the existing config to the new one
 * some are not used

### Tests
* Ran kitchen tests on ubuntu 20 and 22 

### References
* https://ubuntu.com/server/docs/service-nfs
* https://github.com/sous-chefs/nfs/issues/93

### Checklist


cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_alinux2.rb
```diff
@@ -29,7 +29,7 @@
 action_class do
   def override_server_template
     edit_resource(:template, node['nfs']['config']['server_template']) do
-      source 'nfs/nfs.conf.erb'
+      source 'nfs/nfs.erb'
```
enrico-usai: can we use two different names like: `nfs/nfs.ubuntu.erb` and `nfs/nfs.ubuntu22+.erb` to be able to distinguish them in a easy way

cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_ubuntu22+.rb
```diff
@@ -0,0 +1,42 @@
...
+use 'partial/_install_nfs4_and_disable'
+use 'partial/_configure'
+
+default_action :setup
+
+action :setup do
+  action_install_nfs
+  action_install_nfs4
+  action_disable_start_at_boot
+  node.default['nfs']['config']['server_template'] = '/etc/nfs.conf.d/nfs.conf'
```
enrico-usai: You should not do this, you should instead upgrade NFS third party cookbook: https://supermarket.chef.io/cookbooks/nfs#changelog

Support for ubuntu has been added on 5.0.1 and we have 5.0.0 

Maybe with this change we don't need to do any customization at all.

cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_alinux2.rb
dreambeyondorange: sure, the `nfs.erb` is used for all the operating systems, I can create some more reader friendly names

cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_ubuntu22+.rb
dreambeyondorange: 5.0.1 fails due to this known issue and our workaround isn't helping in Ubuntu 22 for some reason https://github.com/sous-chefs/nfs/issues/93

himani2411 (approved):

PR #2191, opened 2023-05-25
Restructure CloudWatch Inspec test to avoid failures on Ubuntu20

enrico-usai:
### Description of changes

Inspec test executed in the validate phase after the build of the image is failing on the gpg --list-keys command because the exit code is 2. Anyway the cloudwatch keys are in the output.

This started to fail when we added the tag:install to the cloudwatch_installation_files control but it does not fail when executed as a singular inspec test.

For this reason I'm currently restructuring the Inspec code to avoid to check the exit code in the Ubuntu20 case.

### Tests
* Tested executing Inspec tests for all OSes on both docker and EC2

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/test/controls/cloudwatch_spec.rb
+++ b/cookbooks/aws-parallelcluster-environment/test/controls/cloudwatch_spec.rb
@@ -15,10 +15,12 @@ control 'tag:install_cloudwatch_installation_files' do
 
   describe 'Check the presence of the cloudwatch package gpg key'
   # In Ubuntu 20.04 due to environment variable the keyring is placed under home of the user ubuntu with the permission of root
+  ubuntu2004 = os_properties.ubuntu2004?
   keyring = os_properties.ubuntu2004? && !os_properties.virtualized? ? '--keyring /home/ubuntu/.gnupg/pubring.kbx' : ''
   sudo = os_properties.redhat_ubi? ? '' : 'sudo'
   describe bash("#{sudo} gpg --list-keys #{keyring}") do
-    its('exit_status') { should eq 0 }
+    # Don't check exit status for Ubuntu20 because it returns 2 when executed in the validate phase of a created AMI
+    its('exit_status') { should eq 0 } unless ubuntu2004
     its('stdout') { should match /3B789C72/ }
     its('stdout') { should match /Amazon CloudWatch Agent/ }
   end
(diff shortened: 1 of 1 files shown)


cookbooks/aws-parallelcluster-environment/test/controls/cloudwatch_spec.rb
```diff
@@ -14,11 +14,18 @@
...
+  if os_properties.ubuntu2004?
+    # In Ubuntu 20.04 due to environment variable the keyring is placed under home of the user ubuntu with the permission of root
+    # We don't check exit status for Ubuntu20 because it returns 2 when executed in the validate phase of a created AMI
+    keyring = !os_properties.on_docker? ? '--keyring /home/ubuntu/.gnupg/pubring.kbx' : ''
+  else
+    describe bash("#{sudo} gpg --list-keys #{keyring}") do
+      its('exit_status') { should eq 0 }
+    end
+  end
   describe bash("#{sudo} gpg --list-keys #{keyring}") do
```
dreambeyondorange, line 21: It might be better to put an if clause inside the describe to avoid calling gpg list keys twice?
```
its('exit_status') { should eq 0 } unless os_properties.ubuntu2004?
```
Is this valid syntax?

dreambeyondorange (dismissed):

cookbooks/aws-parallelcluster-environment/test/controls/cloudwatch_spec.rb
enrico-usai: I tried this and was failing saying that unless cannot be used into the describe, but I'll give another try.

awslitvit: You need to declare a variable outside and use it inside.

Outside: `ubuntu2004 = os_properties.ubuntu2004?`.

Inside: `unless ubuntu2004`.

enrico-usai: Thanks @awslitvit this solved this issue I was facing, that was:

``` 
>>>>>>     Failed to complete #verify action: [`os_properties` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).] on cloudwatch-ubuntu2004
```

cookbooks/aws-parallelcluster-environment/test/controls/cloudwatch_spec.rb
```diff
@@ -15,10 +15,12 @@
 
   describe 'Check the presence of the cloudwatch package gpg key'
   # In Ubuntu 20.04 due to environment variable the keyring is placed under home of the user ubuntu with the permission of root
+  ubuntu2004 = os_properties.ubuntu2004?
   keyring = os_properties.ubuntu2004? && !os_properties.virtualized? ? '--keyring /home/ubuntu/.gnupg/pubring.kbx' : ''
```
dreambeyondorange, line 19: might as well use the new variable here?

dreambeyondorange (approved):

PR #2192, opened 2023-05-25
Add job info daemon to supervisord conf, add configs and log files

chenwany:
### Description of changes
* Add job info daemon to supervisord conf when job info collection is enabled, 
* Set the loop time to be 180 seconds
* Add job info daemon config

### Tests
* Manually test with job info collection is enabled:
```
[ec2-user@ip-192-168-14-181 log]$ sudo cat /var/log/parallelcluster/jobinfomgtd
2023-05-25 03:08:36,544 - [jobinfomgtd:main] - INFO - Jobinfomgtd Startup
2023-05-25 03:08:36,544 - [jobinfomgtd:_get_config] - INFO - Reading /etc/parallelcluster/parallelcluster_jobinfomgtd.conf
2023-05-25 03:08:36,544 - [jobinfomgtd:set_config] - INFO - Applying new jobinfomgtd config: JobinfomgtdConfig(_config=<configparser.ConfigParser object at 0x7f3e69614400>, region='us-west-1', cluster_name='gpu-cluster11', logging_config='/opt/parallelcluster/scripts/jobinfomgtd_logging.conf', loop_time=180, cpu_gpu_cost_ratio=1, custom_discount_rate=0, _boto3_retry=5, _boto3_config={'retries': {'max_attempts': 5, 'mode': 'standard'}}, boto3_config=<botocore.config.Config object at 0x7f3e69620640>)
2023-05-25 03:08:36,544 - [jobinfomgtd:_get_config] - INFO - Reading /etc/parallelcluster/parallelcluster_jobinfomgtd.conf

```

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/templates/log_rotation/parallelcluster_log_rotation.erb
+++ b/cookbooks/aws-parallelcluster-platform/templates/log_rotation/parallelcluster_log_rotation.erb
@@ -70,6 +70,32 @@
   nodateext
 }
 <% end -%>
+<% if node['cluster']['scheduler'] == 'slurm' && node['cluster']['cluster_job_info_enabled'] -%>
+<%# job info log %>
+/var/log/parallelcluster/clusterjobinfomgtd {
+  missingok
+  notifempty
+  nodateext
+  size 50M
+  rotate 1
+  su root root
+  postrotate
+    pkill --signal=SIGUSR2 supervisord
+    echo 0
+  endscript
+  create 0644 root root
+}
+<%# jobinfo events log %>
+/var/log/parallelcluster/clusterjobinfomgtd.events {
+  rotate 1
+  size 10M
+  missingok
+  notifempty
+  copytruncate
+  su pcluster-admin pcluster-admin
+  nodateext
+}
+<% end -%>
 <% if node['cluster']["directory_service"]["generate_ssh_keys_for_users"] == 'true' -%>
 <%# pam_ssh_key_generator log %>
 /var/log/parallelcluster/pam_ssh_key_generator.log {
--- a/cookbooks/aws-parallelcluster-platform/templates/supervisord/parallelcluster_supervisord.conf.erb
+++ b/cookbooks/aws-parallelcluster-platform/templates/supervisord/parallelcluster_supervisord.conf.erb
@@ -39,7 +39,15 @@ command = <%= @dcv_auth_virtualenv_path %>/bin/python <%= @dcv_auth_user_home %>
 user = <%= @dcv_auth_user %>
 environment = HOME="<%= @dcv_auth_user_home %>",USER="<%= @dcv_auth_user %>"<% if @region.start_with?('us-iso') -%>,AWS_CA_BUNDLE="<%= @aws_ca_bundle %>"<% end -%>
 <% end -%>
-
+<% if node['cluster']['scheduler'] == 'slurm' && node['cluster']['cluster_job_info_enabled'] -%>
+[program:clusterjobinfomgtd]
+command = <%= node_virtualenv_path %>/bin/clusterjobinfomgtd
+user = <%= node['cluster']['cluster_admin_user'] %>
+environment = HOME="/home/<%= node['cluster']['cluster_admin_user'] %>",USER="<%= node['cluster']['cluster_admin_user'] %>"<% if @region.start_with?('us-iso') -%>,AWS_CA_BUNDLE="<%= @aws_ca_bundle %>"<% end -%>
+redirect_stderr = true
+stdout_logfile = /var/log/parallelcluster/clusterjobinfomgtd
+stdout_logfile_maxbytes = 0
+<% end -%>
 <%# ComputeFleet -%>
 <% when 'ComputeFleet' -%>
 <% if node['cluster']['scheduler'] == 'slurm' -%>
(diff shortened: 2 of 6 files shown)


cookbooks/aws-parallelcluster-computefleet/attributes/computefleet.rb
davprat, line 1: Is computefleet the right cookbook to use to configure this daemon? I thought that aws-parallelcluster-slurm would be where this should go since that is where head node stuff gets configured?

cookbooks/aws-parallelcluster-computefleet/templates/jobinfomgtd/parallelcluster_jobinfomgtd.conf.erb
```diff
@@ -0,0 +1,7 @@
+[jobinfomgtd]
+cluster_name = <%= node['cluster']['stack_name'] %>
```
davprat: I think 'cluster_name' is preferred to 'stack_name', but both are in use for the head node, so either way is fine.

cookbooks/aws-parallelcluster-computefleet/recipes/config/jobinfomgtd_config.rb
davprat, line 1: I wonder if we shouldn't be using the 'cluster...' naming convention here since 'jobinfo' might be too generic - e.g. if I were an admin and looking at the process list, would I know that jobinfomgtd was something to do with the cluster or not?  What about something like 'clusterjobinfod' since we're also not really managing anything either?

cookbooks/aws-parallelcluster-platform/templates/base/parallelcluster_supervisord.conf.erb
```diff
@@ -39,7 +39,15 @@ command = <%= node['cluster']['dcv']['authenticator']['virtualenv_path'] %>/bin/
 user = <%= node['cluster']['dcv']['authenticator']['user'] %>
 environment = HOME="<%= node['cluster']['dcv']['authenticator']['user_home'] %>",USER="<%= node['cluster']['dcv']['authenticator']['user'] %>"<% if @region.start_with?('us-iso') -%>,AWS_CA_BUNDLE="<%= @aws_ca_bundle %>"<% end -%>
 <% end -%>
-
+<% if node['cluster']['scheduler'] == 'slurm' && node['cluster']['job_info_enabled'] -%>
```
davprat: Are we going to pass this setting as part of the dna.json or are we going to get it from the cluster config yaml? I ask because, if we get it from the yaml, then I think you might need to use a lazy evaluator to correctly resolve the value.

cookbooks/aws-parallelcluster-platform/templates/log_rotation/parallelcluster_log_rotation.erb
```diff
@@ -70,6 +70,32 @@
   nodateext
 }
 <% end -%>
+<% if node['cluster']['job_info_enabled'] -%>
```
davprat: Should this also test for 'slurm' being the scheduler?

davprat (commented):
Looks good - I just want to understand if the `aws-parallelcluster-computefleet` is the right recipe for this to live in? I'm not saying it's not, I just want to understand.

cookbooks/aws-parallelcluster-computefleet/attributes/computefleet.rb
chenwany: Thank you! I read through the aws-parallelcluster-slurm cookbook, and feel like it only contains the installation and configuration of slurm and its dependencies. Under its attribute I only see slurm and its version related attributes.
`aws-parallelcluster-computefleet` has clusterstatusmgtd and get compute fleet status, so I was putting it there. I think it also make sense to move to slurm cookbook considering clustermgtd is also added there, changed

cookbooks/aws-parallelcluster-platform/templates/base/parallelcluster_supervisord.conf.erb
chenwany: The plan is to pass it from  dna json instead of the config. The reason is that we only set it to true in dna json on headnode when it read true from the config.

cookbooks/aws-parallelcluster-platform/templates/log_rotation/parallelcluster_log_rotation.erb
chenwany: yep, currently the default is false, it will only be set to True when enabled it through cluster configuration, which should only support with slurm scheduler.

cookbooks/aws-parallelcluster-computefleet/templates/jobinfomgtd/parallelcluster_jobinfomgtd.conf.erb
chenwany: sounds good, change to cluster_name

cookbooks/aws-parallelcluster-computefleet/recipes/config/jobinfomgtd_config.rb
chenwany: good idea, rename

cookbooks/aws-parallelcluster-computefleet/attributes/computefleet.rb
```diff
@@ -1,4 +1 @@
 default['cluster']['computefleet_status_path'] = "#{node['cluster']['shared_dir']}/computefleet-status.json"
-
-# Compute nodes bootstrap timeout
-default['cluster']['compute_node_bootstrap_timeout'] = 1800
```
davprat: Was this from a merge?

davprat (approved):
👍

PR #2202, opened 2023-05-27
Move gdrcopy resource to platform cookbook

awslitvit:
### Description of changes
* [Allow setting properties on resource dependencies.](https://github.com/aws/aws-parallelcluster-cookbook/commit/d9b29293749abcb588985b1de0d37c26b5d0d248) 
* [Move gdrcopy resource to platform cookbook](https://github.com/aws/aws-parallelcluster-cookbook/commit/a00cc518cc953a9497e675b02504f94b9d1880a0)

### Tests
* `bash kitchen.ec2.sh platform-install verify nvidia-gdrcopy -c 6`: OK (except on Centos7, this is a known issue)
* `bash kitchen.ec2.sh platform-config verify nvidia-gdrcopy -c 6`: OK (except on Centos7, this is a known issue)
* converge not working on Docker because of missing libraries
  * no way to easily skip not working parts, as dkms dependency can not be installed on Docker
  * system tests will not fail because they don't enable nvidia
  * decided to keep code as is and add a warning about tests not working on Docker

### Checklist


cookbooks/aws-parallelcluster-platform/kitchen.platform-install.yml
```diff
@@ -209,3 +209,26 @@ suites:
...
+        - /tag:install_.*gdrcopy/
+    driver:
+      # nvidia_driver can be executed only on a graphic EC2 instance example: g5.xlarge(x86_86) or g5g.xlarge(aarm64)
+      instance_type: g4dn.2xlarge
+    attributes:
+      resource: gdrcopy
+      dependencies:
+        - resource:package_repos
+        - resource:install_packages:install_kernel_source
+        - 'resource:package { "package_name": "dkms" }'
```
enrico-usai, line 228: q: dkms package is already included on gdrcopy deps, why do we need to explicitly add it here?
Are we missing it as deps of nvidia?

cookbooks/aws-parallelcluster-platform/resources/gdrcopy/partial/_gdrcopy_common.rb
```diff
@@ -12,25 +12,37 @@
...
+property :gdrcopy_checksum, String, default: 'b85d15901889aa42de6c4a9233792af40dd94543e82abe0439e544c87fd79475'
+
+unified_mode true
+default_action :setup
+
+action :setup do
+  return unless gdrcopy_enabled?
+
+  # Save gdrcopy version for InSpec tests
+  node.default['cluster']['nvidia']['gdrcopy']['version'] = new_resource.gdrcopy_version
```
enrico-usai, line 25: When there are this kind of variables that are required by both code and inspec wouldn't be better to avoid defining them as properties of the resource?

The code would be cleaner and we would have the possibility to override them from chef node attributes at build image time.

enrico-usai (approved):

cookbooks/aws-parallelcluster-platform/kitchen.platform-install.yml
awslitvit: This is just a copy/paste from other nvidia resources. I guess we can remove this line.

cookbooks/aws-parallelcluster-platform/resources/gdrcopy/partial/_gdrcopy_common.rb
awslitvit: My reasoning is that this way we explicitly surface the dependencies.

PR #2228, opened 2023-05-31
Add graphic_instance condition to dcv_gpu_accel_supported function

enrico-usai:
Having nvidia_installed? is not enough.

### Tests
* Executed ChefSpec.


awslitvit (approved):

enrico-usai:
Skipped system tests because they passed in previous run before the rebase.

PR #2236, opened 2023-05-31
Fix DCV InSpec test for ARM case

enrico-usai:
### Tests:

* `bash kitchen.ec2.sh platform-install test dcv-alinux2` passing on x86
* the same test is not passing on arm because we are missing to enable the amazon-linux-extras epel topic as dependency of the test (TODO).

### DIFF
--- a/kitchen.ec2.sh
+++ b/kitchen.ec2.sh
@@ -8,6 +8,9 @@
 # KITCHEN_ARCHITECTURE:       [x86_64|arm64];
 #                             default x86_64
 #
+# KITCHEN_INSTANCE_TYPE:      instance type to use
+#                             default t2.micro
+#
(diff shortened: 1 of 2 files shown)


awslitvit (dismissed):

kitchen.ec2.sh
```diff
@@ -125,6 +128,7 @@ export KITCHEN_KEY_NAME
 export KITCHEN_SSH_KEY_PATH
 export KITCHEN_AVAILABILITY_ZONE
 export KITCHEN_ARCHITECTURE
+export KITCHEN_INSTANCE_TYPE
```
awslitvit, line 131: Nit: this should be redundant.

awslitvit (approved):

PR #2241, opened 2023-06-01
Move directory_service.rb to environment cookbook

enrico-usai:
* Moved inspec tests
* Moved templates
* Moved suite definition
* Moved parse_arn function to the recipe where it is used
* Checked attributes used by the recipe, they have been already moved to environment cookbook
* Added on_docker conditions to recipe and inspec tests to permit to run them on Docker

### Tests

* `bash kitchen.ec2.sh environment-config test directory-service -c 5` (all oses except U22) -> OK
* `bash kitchen.docker.sh environment-config test directory-service -c 5` (all oses except U22) -> OK


cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
```diff
@@ -33,6 +33,33 @@ suites:
...
         cw_logging_enabled: "true"
         log_group_name: test
+  - name: directory_service
+    # FIXME: breaks on Docker
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-environment::directory_service]
+    verifier:
+      controls:
+        - /tag:config_sssd_configured_correctly/
```
awslitvit: The control can not be tagged with `tag:config` because it relies on attributes we set in this specific test. The control will fail in scope of `validate_config` test.

cookbooks/aws-parallelcluster-environment/recipes/config/directory_service.rb
```diff
@@ -24,6 +24,24 @@
 shared_directory_service_dir = "#{node['cluster']['shared_dir']}/directory_service"
```
awslitvit, line 24: This recipe should really be ChefSpec-tested, it has complex logic.

awslitvit (changes requested):

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
enrico-usai: you're right! thanks

awslitvit (approved):

enrico-usai:
Skipping system tests because passed before the final rebase.

PR #2245, opened 2023-06-02
POC for kitchen lifecycle hooks.

awslitvit:
### Description of changes

`kitchen.global.yml` configures lifecycle hooks to run potentially pre and post every operation. 
`kitchen.run-hook.sh` verifies if the hook script exists under `<cookbook>>/test/hooks/<install|config>/<suite>`. 
If the script exists, it is called.

`kitchen.global.yml` uses kitchen environment, which is configured in `test/environments/kitchen.rb`.

In order to share values with kitchen recipes running remotely, attribute values must be set in this file.

For every value to pass and for every OS, add a line to this file:

```
'<variable>/<platform>' => 'placeholder'
```

For instance: `'ebs_volume_id_ebs_mount/alinux2' => 'placeholder'`.

In your hook scripts, use sed to replace the placeholder with the actual value to share with the remote host. Also, read the value from the same file to retrieve it locally, for instance to be able to delete a resource during the destroy phase.

The example in this POC uses platform-install hooks recipe and test (created ad-hoc for the POC). It creates an EBS volume using post_create hook, shares it with hooks recipe which logs the value, and destroys it in pre_destroy hook.

### Tests
* `./kitchen.ec2.sh platform-install verify hooks -c 5`
* `./kitchen.ec2.sh platform-install destroy hooks -c 5`

One volume has been created for each platform, logged by hooks recipe and then deleted.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/kitchen.platform-install.yml
+++ b/cookbooks/aws-parallelcluster-platform/kitchen.platform-install.yml
@@ -324,3 +324,10 @@ suites:
         is_official_ami_build: true
     driver:
       instance_type: t3.medium
+  - name: hooks
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-platform::hooks]
+    verifier:
+      controls:
+        - /nothing/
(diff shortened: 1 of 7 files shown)


enrico-usai:
We can close this patch since this code has been used for 
* https://github.com/aws/aws-parallelcluster-cookbook/pull/2258
* https://github.com/aws/aws-parallelcluster-cookbook/pull/2266

PR #2249, opened 2023-06-05
Define node['cluster']['etc_dir'] attribute

enrico-usai:


cookbooks/aws-parallelcluster-config/templates/default/init/fetch_and_run.erb
```diff
@@ -16,7 +16,7 @@ set -euo pipefail
 # Disclaimer: this env setup is for *internal use only*, custom actions scripts SHOULD NOT rely on it
 
 # this is kept solely for backward compatibility, variables defined by it *SHOULD NOT* be used
-source /etc/parallelcluster/cfnconfig
+source <%= node['cluster']['etc_dir'] %>/cfnconfig
```
awslitvit, line 19: Would it be possible to pass it as variable, in order to decouple templates from attribute structure which was one of our goals?

enrico-usai: In the same file there are a lot of other attributes:
```
PCLUSTER_SCRIPTS_DIR="<%= node['cluster']['scripts_dir'] %>"
PCLUSTER_CONFIG_PATH="<%= node['cluster']['cluster_config_path'] %>"
PCLUSTER_NODE_TYPE="<%= node['cluster']['node_type'] %>"
PCLUSTER_SCHEDULER_QUEUE_NAME="<%= node['cluster']['scheduler_queue_name'] %>"
PCLUSTER_STACK_NAME="<%= node['cluster']['stack_name'] %>"
PCLUSTER_REGION="<%= node['cluster']['region'] %>"
```

so we can address this change to use parameters rather than node attributes for the full file in future.

awslitvit (approved):

PR #2259, opened 2023-06-07
Fix dcv_configured condition usage 

enrico-usai:
The `node['conditions']['dcv_supported']` is set in the dcv resource, so it's not always available in the environment.
By using the DCV file existence we're sure that DCV is installed and it can happen only on supported OSes.

Within this patch we're replacing `dcv_supported` node attribute with a helper function.
`dcv_supported` condition was only used on the kitchen tests and it's equivalent to use File.exist condition: `dcv.conf` file is present on supported AMIs only.

Note: the `node['cluster']['dcv_enabled'] == "head_node"` is required, we added it as part of: https://github.com/aws/aws-parallelcluster-cookbook/pull/2246

### Tests

Updated and extended dcv_spec tests

### DIFF
--- a/cookbooks/aws-parallelcluster-config/spec/unit/recipes/sudo_config_spec.rb
+++ b/cookbooks/aws-parallelcluster-config/spec/unit/recipes/sudo_config_spec.rb
@@ -28,7 +28,7 @@ describe 'aws-parallelcluster-config::sudo' do
           runner = ChefSpec::Runner.new(platform: platform, version: version) do |node|
             node.override['cluster']['node_type'] = 'HeadNode'
             node.override['cluster']['dcv_enabled'] = 'head_node'
-            allow(File).to receive(:exist?).with('/etc/dcv/dcv.conf').and_return(true)
+            allow_any_instance_of(Object).to receive(:dcv_installed?).and_return(true)
           end
           runner.converge(described_recipe)
         end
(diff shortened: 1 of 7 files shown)


awslitvit (dismissed):

gmarciani (dismissed):

cookbooks/aws-parallelcluster-config/spec/unit/recipes/sudo_config_spec.rb
```diff
@@ -28,7 +28,7 @@
           runner = ChefSpec::Runner.new(platform: platform, version: version) do |node|
             node.override['cluster']['node_type'] = 'HeadNode'
             node.override['cluster']['dcv_enabled'] = 'head_node'
-            node.override['conditions']['dcv_supported'] = true
+            allow(File).to receive(:exist?).with('/etc/dcv/dcv.conf').and_return(true)
```
awslitvit: You should mock the function, not the file existence.

cookbooks/aws-parallelcluster-platform/test/controls/dcv_spec.rb
```diff
@@ -166,7 +166,7 @@
 end
 
 control 'tag:config_dcv_external_authenticator_user_and_group_correctly_defined' do
-  only_if { node['conditions']['dcv_supported'] && !os_properties.redhat_ubi? }
+  only_if { dcv_installed? && !os_properties.redhat_ubi? }
```
awslitvit, line 169: Was it already defined?

awslitvit (changes requested):

cookbooks/aws-parallelcluster-config/spec/unit/recipes/sudo_config_spec.rb
enrico-usai: done. I missed to update it when I added the function definition in the latest commit.

cookbooks/aws-parallelcluster-platform/test/controls/dcv_spec.rb
enrico-usai: added, thanks for the hint!

enrico-usai:
closed in favour of: https://github.com/aws/aws-parallelcluster-cookbook/pull/2261

PR #2261, opened 2023-06-07
[develop] Fix dcv_configured condition usage

enrico-usai:
The `node['conditions']['dcv_supported']` is set in the dcv resource, so it's not always available in the environment.
By using the DCV file existence we're sure that DCV is installed and it can happen only on supported OSes.

Within this patch we're replacing `dcv_supported` node attribute with a helper function.
`dcv_supported` condition was only used on the kitchen tests and it's equivalent to use File.exist condition: `dcv.conf` file is present on supported AMIs only.

Note: the `node['cluster']['dcv_enabled'] == "head_node"` is required, we added it as part of: https://github.com/aws/aws-parallelcluster-cookbook/pull/2246

### Tests

* Updated and extended dcv_spec tests
* `bash kitchen.ec2.sh platform-install test dcv-alinux2`

### DIFF
--- a/cookbooks/aws-parallelcluster-common/test/common/libraries/instance.rb
+++ b/cookbooks/aws-parallelcluster-common/test/common/libraries/instance.rb
@@ -31,6 +31,10 @@ class Instance < Inspec.resource(1)
     !inspec.node['ec2']['instance_type'].start_with?(*unsupported_gpu_accel_list)
   end
 
+  def dcv_installed?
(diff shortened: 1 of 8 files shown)


awslitvit (approved):

enrico-usai:
Closed in favour of https://github.com/aws/aws-parallelcluster-cookbook/pull/2262 (GitHub was having issues).

PR #2280, opened 2023-06-12
Move log_rotation to platform cookbook

enrico-usai:
* Move template and default attribute
* Log rotation template depends by scheduler, node_type, dcv and directory_service parameters

### Tests

* Write ChefSpec to test the content of the template in the different cases


cookbooks/aws-parallelcluster-platform/spec/unit/recipes/log_rotation_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        cached(:chef_run) do
+          runner = runner(platform: platform, version: version) do |node|
+            node.override['cluster']['node_type'] = "HeadNode"
+            node.override['cluster']['log_rotation_enabled'] = 'true'
+            node.override['cluster']['dcv_enabled'] = "head_node"
+            node.override['cluster']["directory_service"]["generate_ssh_keys_for_users"] = 'true'
+            node.override['cluster']["scheduler"] = 'slurm'
+            allow_any_instance_of(Object).to receive(:dcv_installed?).and_return(true)
+          end
+          runner.converge(described_recipe)
```
awslitvit, line 18: As far as I know, this is converging 2 times: one time because if line 8 and the other one here.
You should converge in the most nested code snippet.

awslitvit (changes requested):

cookbooks/aws-parallelcluster-platform/spec/unit/recipes/log_rotation_spec.rb
enrico-usai: You're right, I missed to remove the first converge (from a cut and paste).

awslitvit (approved):

PR #2282, opened 2023-06-12
Move update_computefleet_status to computefleet cookbook

enrico-usai:
Define update_computefleet_status entrypoint and use it to update cluster status from clusterstatusmgtd.py.

Removed check for awsbatch scheduler from entrypoint code since it's present in the recipe too.

I checked in the CLI code and this entrypoint is not used.


awslitvit (approved):

enrico-usai:
Skipping system tests since they passed before the final rebase.

PR #2283, opened 2023-06-12
Fix dcv condition used in Inspec Tests

enrico-usai:
We removed dcv_supported condition in a previous commit in favour of the dcv_installed function that checks if dcv is installed.

DCV is installed on supported OSes/archs only.

### References
* Issue introduced with: https://github.com/aws/aws-parallelcluster-cookbook/pull/2262 on which we removed the condition but missed to update this test.


dreambeyondorange (approved):

enrico-usai:
Skipping system tests since they passed before the final rebase.

PR #2290, opened 2023-06-13
Refactoring of config main recipe

enrico-usai:
Preliminary commits:
* Move base.rb code into config.rb
* Reorder recipe execution to be consistent in head node and compute fleet
* Be sure the fs_update recipe is executed on head node only
* Be sure the fleet_status recipe is executed on head node only
  * Rename recipe from head_node_fleet_status.rb to fleet_status.
* Unify head node and compute fleet flow
  * The guards to skip the execution or execute different behaviours according to the node type are now into the recipes
* Move scheduler guards in the config recipes of the schedulers cookbooks 

And then there is the final commit to reorder config steps. This is the order we're following:
- enable_chef_error_handler must be executed as first to be able to handle errors during chef execution.
- Then there is the only shared recipe to setup the environment
- we have most of the platform recipes to setup system packages, networking, nvidia, dcv, intel as first
- EFA
- nfs, ephemeral drives and all the shared storage mount/export steps,
  - fs_update generates the shared storages mapping file so must be executed before shared storages recipes
- we fetch cluster config then execute compute fleet and scheduler related steps
- at the end will call log_rotation since it depends by all the configured features (dcv, scheduler, directory_service).


cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  template "#{node['cluster']['etc_dir']}/clusterstatusmgtd.conf" do
+    source 'clusterstatusmgtd/clusterstatusmgtd.conf.erb'
+    owner 'root'
+    group 'root'
+    mode '0644'
+  end
+
+  unless virtualized? || kitchen_test? && !node['interact_with_ddb']
```
awslitvit: `virtualized?` should be removed

cookbooks/aws-parallelcluster-environment/recipes/config.rb
```diff
@@ -0,0 +1,26 @@
...
+
+efa 'Configure system for EFA' do
+  action :configure
+end
+nfs "Configure NFS" do
+  action :configure
+end
+include_recipe 'aws-parallelcluster-environment::ephemeral_drives'
+# fs_update generates the shared storages mapping file so must be executed before shared storages recipes
+include_recipe 'aws-parallelcluster-environment::fs_update'
```
awslitvit, line 22: Is `fs_update` name is correct? I didn't expect to find it on the config path.

cookbooks/aws-parallelcluster-platform/recipes/config.rb
```diff
@@ -0,0 +1,27 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+include_recipe 'aws-parallelcluster-platform::openssh'
```
awslitvit, line 14: Nitpick: maybe we can work on the naming, putting verbs instead of nouns? For instance:
- it's not clear what `openssh` actually does
- `cluster_user` is actually `create_cluster_user`

cookbooks/aws-parallelcluster-platform/recipes/config.rb
```diff
@@ -0,0 +1,27 @@
...
+include_recipe "aws-parallelcluster-platform::sudo_config"
+include_recipe 'aws-parallelcluster-platform::cluster_user'
+include_recipe 'aws-parallelcluster-platform::networking'
+include_recipe "aws-parallelcluster-platform::nvidia_config"
+sticky_bits "setup sticky bits"
+chrony 'enable Amazon Time Sync' do
+  action :enable
+end
+include_recipe 'aws-parallelcluster-platform::dcv'
```
awslitvit, line 23: Shouldn't it be moved to `dcv` resource?

cookbooks/aws-parallelcluster-platform/recipes/config.rb
```diff
@@ -0,0 +1,27 @@
...
+intel_hpc 'Configure Intel HPC' do
+  action :configure
+end
+fetch_config 'Fetch and load cluster configs'
```
awslitvit, line 27: So no code above depends on the cluster config?

awslitvit (dismissed):

cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
enrico-usai: fixed, thanks!

cookbooks/aws-parallelcluster-platform/recipes/config.rb
enrico-usai: correct, you can see that when I unified base.rb with config.rb the fetch_config was already in this position: https://github.com/aws/aws-parallelcluster-cookbook/pull/2290/commits/e7faebcc2ca9c26536e6cda20f391a5d7d281584

cookbooks/aws-parallelcluster-environment/recipes/config.rb
enrico-usai: this is updating/creating the shared_storage_mapping file, we can find a better naming for this recipe.

cookbooks/aws-parallelcluster-platform/recipes/config.rb
enrico-usai: I'm not sure on having the verb before the nouns, especially if there are multiple "actions" for the same recipe. 

I'm probably in favour of having: `cluster_user_setup` or `cluster_user_create`.

Anyway I agree we can find a better naming for these recipes, but we can do it in a separate patch.

awslitvit: Usually it's recommended to use verbs.

awslitvit (approved):

PR #2292, opened 2023-06-14
Remove aws-parallelcluster-config cookbook

enrico-usai:
This patch is composed by multiple commits and the final result is the removal of the aws-parallelcluster-config cookbook.

I'd suggest to look at them one by one.


cookbooks/aws-parallelcluster-computefleet/kitchen.computefleet-config.yml
```diff
@@ -4,6 +4,10 @@ verifier:
   inspec_tests:
     - cookbooks/aws-parallelcluster-computefleet/test
 
+  <%
+  aws_region = ENV['KITCHEN_AWS_REGION'] || ENV['AWS_DEFAULT_REGION'] || 'eu-west-1'
```
awslitvit: This should be done upstream, otherwise hooks won't work.

awslitvit (approved):

PR #2293, opened 2023-06-14
Remove aws-parallelcluster-test cookbook

enrico-usai:
This patch is composed by multiple commits and the final result is the removal of the aws-parallelcluster-config cookbook.

I'd suggest to look at them one by one.

The only remaining file in the cookbook was test_primary_ip.rb that is not called from anywhere.


cookbooks/aws-parallelcluster-scheduler-plugin/recipes/test/tests_mock.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
```
awslitvit, line 1: Do we use it at all?

awslitvit (approved):

cookbooks/aws-parallelcluster-scheduler-plugin/recipes/test/tests_mock.rb
enrico-usai: We introduced the usage of it because it permits a minor optimization: https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/

PR #2287, opened 2023-06-13
Remove aws-parallelcluster-common cookbook

enrico-usai:
### Description of changes
common/libraries used by tests (instance.rb, node.rb, os_properties.rb) have been moved to shared

PR #2295, opened 2023-06-15
Fix shared Inspec profile structure

enrico-usai:
The Inspec profiles of functional cookbooks was pointing to the common folder of the shared cookbook but that folder didn't contain an Inspec profile.

Now the profile structure of the shared cookbook is correct. It contains both controls and libraries folders.

### Tests
* `bash kitchen.ec2.sh awsbatch-install test awsbatch-virtualenv-alinux2`

### References
* Issue introduced with: https://github.com/aws/aws-parallelcluster-cookbook/pull/2287
* https://docs.chef.io/inspec/profiles/


awslitvit (approved):

davprat (approved):
lgtm

PR #2301, opened 2023-06-16
Execute fetch_config after the mount of shared folders

enrico-usai:
The cluster config is saved in one of the shared folder in this way the compute nodes do not need to retrieve them from s3.

Anyway this movement, introduced as part of:
https://github.com/aws/aws-parallelcluster-cookbook/commit/1b00de8aefe09ae9ddd863539e3b982396607f1c permitted to discover that compute nodes were unable to download the config from s3, fetch config could not work on compute nodes, because the cluster/cluster_s3_bucket node attribute was not exported in the compute node dna.json.

### References
* Related to: https://github.com/aws/aws-parallelcluster/pull/5408


lukeseawalker (approved):

enrico-usai:
Skipping system tests since it's just a reorder on recipe execution

PR #2286, opened 2023-06-13
Use volume resource for ebs recipe and cleanup of all common libraries

enrico-usai:
### Description of changes
* Use volume resource for ebs recipe
  * I had to move format_directory to environment libraries because explicitly used by the ebs recipe.
* Replace all usage of virtualized with on_docker
* Delete unused graphic_instance function
  * This is defined in platform libraries too and used by nvidia, dcv, slurm recipes.

### Tests
* Moved tests for format_directory method.

PR #2303, opened 2023-06-16
Fix volume mount

enrico-usai:
index is not defined. Issue introduced when changing from nfs mount to volume resource: https://github.com/aws/aws-parallelcluster-cookbook/pull/2286

### Tests

Covered case with test.


awslitvit (dismissed):

enrico-usai:
Skipping system tests since they are failing for a separate reason and this patch is important to be merged ASAP.

lukeseawalker (approved):

PR #2304, opened 2023-06-16
Execute inspec tests accordingly to node type and scheduler

enrico-usai:
* Execute scheduler tests according to the scheduler
* Do not execute slurmd checks in the head node. slurmd service is available on the compute nodes only.


cookbooks/aws-parallelcluster-slurm/test/controls/config_slurmd_systemd_service_spec.rb
```diff
@@ -11,7 +11,7 @@
 
 control 'tag:config_systemd_slurmd_service' do
   title 'Check the basic configuration of the systemd slurmd service'
-  only_if { !os_properties.on_docker? }
+  only_if { !os_properties.on_docker? && instance.compute_node? }
```
awslitvit: Nitpicks:
1. Should we also change the control name to reflect it's only for compute nodes?
2. I find it more readable using multiple `only_if` statements. If you agree, could you try and confirm it actually works?

cookbooks/aws-parallelcluster-slurm/test/controls/config_slurmd_systemd_service_spec.rb
```diff
@@ -11,7 +11,7 @@
 
 control 'tag:config_systemd_slurmd_service' do
   title 'Check the basic configuration of the systemd slurmd service'
-  only_if { !os_properties.on_docker? }
+  only_if { !os_properties.on_docker? && instance.compute_node? && node['cluster']['scheduler'] == 'slurm' }
```
awslitvit, line 14: Nitpicks:
1. Should we also change the control name to reflect it's only for slurm compute nodes?
2. I find it more readable using multiple `only_if` statements. If you agree, could you try and confirm it actually works?

awslitvit (approved):

enrico-usai:
Skip system tests since they passed on previous run before the final rebase.

PR #2308, opened 2023-06-16
Skip mount_shared recipe tests on head node

enrico-usai:
These steps are executed in the compute nodes only.

Remove `tag:config` since this test requires some folders to be exported from the head node.

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
+++ b/cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
@@ -418,13 +418,14 @@ suites:
       - recipe[aws-parallelcluster-environment::mount_shared]
     verifier:
       controls:
-        - /tag:config_mount_shared/
+        - mount_shared
     attributes:
       dependencies:
         - recipe:aws-parallelcluster-platform::directories
         - resource:nfs
         - recipe:aws-parallelcluster-environment::mock_export_directories
       cluster:
+        node_type: 'ComputeFleet'
         head_node_private_ip: '127.0.0.1'
         head_node_home_path: '/fake_headnode_home'
         shared_dir_head: '/fake_headnode_shared'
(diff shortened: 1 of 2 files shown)


cookbooks/aws-parallelcluster-environment/test/controls/mount_shared_spec.rb
```diff
@@ -9,10 +9,10 @@
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 
-control 'tag:config_mount_shared_configured' do
+control 'mount_shared_configured' do
```
jdeamicis: [Nitpick] I would call this control "mount_shared_mounted" or "headnode_shared_mounted" instead of "mount_shared_configured" because the control is not testing that the mounts are configured, but that they are actually mounted.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
```diff
@@ -418,7 +418,7 @@ suites:
       - recipe[aws-parallelcluster-environment::mount_shared]
     verifier:
       controls:
-        - /tag:config_mount_shared/
+        - /mount_shared/
```
jdeamicis: [Nitpick] this always confuses me: can we try to use the full names of controls **unless** we specifically have a list of controls with the same initial pattern that we can use to specify multiple controls at once?

jdeamicis (dismissed):

cookbooks/aws-parallelcluster-environment/test/controls/mount_shared_spec.rb
enrico-usai: Sure, I didn't modify the tests I was just moving them from root folder to functional specific cookbooks. Fixed.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
enrico-usai: The idea is that if we use a regex in future we can avoid to touch this config file if we add more controls for the same recipe. Anyway fixed for now.

jdeamicis (approved):

PR #2311, opened 2023-06-19
Push instance-to-node mapping to cw log, configure new log group

Jennifer603:
### Description of changes
* Push instance-to-node mapping to cw log
* configured new log group nodemap

### Tests
* Manually tested by creating cluster and check log event nodemap

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/files/cloudwatch/write_cloudwatch_agent_json.py
+++ b/cookbooks/aws-parallelcluster-environment/files/cloudwatch/write_cloudwatch_agent_json.py
@@ -30,6 +30,17 @@ def parse_args():
         help="Role this node plays in the cluster " "(i.e., is it a compute node or the head node?)",
     )
     parser.add_argument("--scheduler", required=True, choices=["slurm", "awsbatch"], help="Scheduler")
+    parser.add_argument(
+        "--additional-log-groups",
+        required=False,
+        help="format: <key-1>=<log-group-name-1>,<key-2>=<log-group-name-2>,<key-3>=<log-group-name-3>,...",
+    )
+    parser.add_argument(
+        "--output-path",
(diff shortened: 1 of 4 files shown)


cookbooks/aws-parallelcluster-environment/files/cloudwatch/write_cloudwatch_agent_json.py
```diff
@@ -45,15 +56,19 @@ def gethostname():
...
+    with open(output_file, "w+", encoding="utf-8") as output_config_file:
         json.dump(config, output_config_file, indent=4)
 
 
-def add_log_group_name_params(log_group_name, configs):
+def add_log_group_name_params(default_log_group_name, log_group_name_map, configs):
     """Add a "log_group_name": log_group_name to every config."""
     for config in configs:
+        log_group_name = default_log_group_name
+        group_key = config.get("log_group_key")
```
davprat: I think you need to provide a default here: `.get("log_group_key", None)` otherwise this call will throw an error if "log_group_key" is not present in `config`.

cookbooks/aws-parallelcluster-environment/files/cloudwatch/write_cloudwatch_agent_json.py
```diff
@@ -45,15 +56,19 @@ def gethostname():
...
+        if group_key:
+            log_group_name = log_group_name_map.get(group_key)
```
davprat: Same here: better to provide a default value rather than raise an exception.

davprat (changes requested):
Looks good - I left a couple comments to address, plus we need to figure out why the checks are failing and fix those issues as well.

davprat (approved):
lgtm

PR #2315, opened 2023-06-20
Move dcv authenticator virtualenv path in platform attributes

enrico-usai:
### Description of changes
* Move sudoers and supervisord templates in the right folders
* Minor renaming of dcv parameter
* Move dcv authenticator virtualenv path in platform attributes. In this way we are sure that the variable is available for supervisord_config.rb


himani2411 (approved):

enrico-usai:
Skipping system tests since they passed before the final update.

PR #2323, opened 2023-06-21
Create JSON file with mapping between PC-managed Slurm partitions and PC-managed Slurm nodelists

jdeamicis:
### Description of changes
* Create JSON file with mapping between PC-managed Slurm partitions and PC-managed Slurm nodelists.
* Set `--autofix` and `--indent 4` in the pretty-format-json pre-commit hook.

### Tests
* Add unit test to verify that the mapping file is created and loaded correctly.

### References
* https://github.com/aws/aws-parallelcluster/pull/5442
* https://github.com/aws/aws-parallelcluster-node/pull/539

### Checklist
- [ ] ~Check if documentation is impacted by this change.~

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
+++ b/cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
@@ -144,6 +144,7 @@ class QueueRenderer:
             for compute_resource_config in queue_config["ComputeResources"]
         ]
         self.job_exclusive_allocation = queue_config.get("JobExclusiveAllocation")
+        self.nodelist = None
 
     def render_config(self):
         """Launch the rendering of the required configuration."""
@@ -154,6 +155,15 @@ class QueueRenderer:
             config += self._render_as_partition_config()
 
         return config
+
+    def get_queue_nodelist(self):
+        """Getter method that returns the nodelist used in the QueueRenderer object."""
+        if not self.nodelist:
+            nodes = []
+            for renderer in self.compute_renderers:
+                nodes.extend(renderer.render_as_nodeset_element())
+            self.nodelist = ",".join(nodes)
+        return self.nodelist
 
     def _render_as_partition_config(self):
         partition_config = "\n"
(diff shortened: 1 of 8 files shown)


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -171,12 +190,8 @@ def _render_as_gres_config(self):
...
     def _render_nodeset(self):
         nodeset = f"NodeSet={self.name}_nodes Nodes="
-        nodes = []
-        for renderer in self.compute_renderers:
-            nodes.extend(renderer.render_as_nodeset_element())
-
-        nodeset += ",".join(nodes)
-
+        nodelist = self._get_queue_nodelist()
+        nodeset += nodelist
```
NSsirena: This can be shortened to `return f"NodeSet={self.name}_nodes Nodes={self._get_queue_nodelist()}"`

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -167,6 +170,25 @@ def _generate_queue_config(
...
+        filename = path.join(output_dir, f"parallelcluster_{queue['Name']}_partition_nodelist_mapping.json")
+        with open(filename, "r", encoding="utf-8") as partition_mapping_file:
+            partition_mapping = json.load(partition_mapping_file)
+            full_mapping[queue["Name"]] = partition_mapping[queue["Name"]]
+        remove(filename)
+
+    # Dump the full mapping to a file
+    filename = path.join(output_dir, "parallelcluster_partition_nodelist_mapping.json")
+    with open(filename, "w", encoding="utf-8") as mapping_file:
+        json.dump(full_mapping, mapping_file, indent=4)
```
NSsirena: [Nitpick] Please add a comment (or TODO) to recall us that this method should be refactored.

Above (line 81), most of the parameters we pass to `_generate_queue_config` are used to create an instance of QueueRenderer inside the function.
We can create this object outside the function and pass the object as parameter instead.

Doing so we can ask the object to return us (without creating a temporary file) the mapping between nodelist and partitions and add this partial mapping to the `full_mapping`.

This will make this for loop redundant and the only piece of code we need is the part that writes the final file, that could be moved up and replace the call to this method (that will be redundant).

Not sure it's clear, let's sync up offline if you prefer.

jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
jdeamicis: Addressed.

NSsirena (dismissed):

hanwen-pcluste (approved):

PR #2328, opened 2023-06-26
Add utility scripts to simplify testing of a new OS

enrico-usai:
## Add new utility os-resources.py to simplify creation of OS specific resources

The new utility `os-resources.py` script is able to copy the `resources/<name>/<name>_{src_platform}{src_version}`
into a `resources/<name>/<name>_{dst_platform}{dst_version}`, replacing the following patterns:
* `platform: '{src_platform}'`
* `platform?('{src_platform}')`
* `to_i == {src_version}`
* `platform_version: '{src_version}'`

This utility permits to automatically generate all the required files,
then they must be checked one by one to see if there are OS specific checks on them.
Templates files must be manually copied.

## Add test-os.sh utility to simplify run of kitchen tests for an OS

The script is able to search for all the test suites defined
into the functional cookbook and run them sequentially.

It must be executed from the root folder of the cookbook repo:
`bash util/test-os.sh ec2 rhel8`

The output will be saved into a `/tmp/kitchen-<os>-<date>` folder.

To run config tests is required to set a `KITCHEN_${os}_AMI` environment variable (e.g. `KITCHEN_RHEL8_AMI`)
or have a ParallelCluster AMI in the account.

### DIFF
--- a/util/os-resources.py
+++ b/util/os-resources.py
@@ -1,0 +1,115 @@
+# !/usr/bin/python3
+import os
+import shutil
+
+import click
+
+
+@click.group()
+def cli():
+    """Utility script to list and generate cookbook resources files for a given OS.
+
+    \b
+    The script copies the `resources/<name>/<name>_{src_platform}{src_version}` cookbook resources
+    into new created files: `named: resources/<name>/<name>_{dst_platform}{dst_version}`,
+    replacing the platform and version patterns.
+
+    This utility permits to automatically generate all the required files,
+    then they must be checked one by one to see if there are OS specific checks on them.
+    Templates and other specific files must be manually copied.
+
+    \b
+    Usage:
+    python3 util/os-resources.py list-resources --cookbooks-path cookbooks --platform redhat --version 8
+    python3 util/os-resources.py generate --src-platform redhat --src-version 8 --dst-platform rocky --dst-version 8 --cookbooks-path cookbooks
+    """
+    pass
+
+
+@cli.command()
+@click.option("--cookbooks-path", help="Directory containing the cookbooks.", required=True, type=click.Path())
+@click.option(
+    "--src-platform",
+    help="Platform of the os to copy from (e.g. redhat)",
+    type=click.Choice(["amazon", "centos", "redhat", "ubuntu"]),
+    required=True,
+)
+@click.option(
+    "--src-version",
+    help="Version of the os to copy from (e.g. 8)",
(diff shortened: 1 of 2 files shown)


util/os-resources.py
```diff
@@ -0,0 +1,84 @@
...
+import shutil
+import os
+
+@click.group()
+def cli():
+    pass
+
+@cli.command()
+@click.option("--cookbooks-path", help="Directory containing the cookbooks.", required=True, type=click.Path())
+@click.option("--src-platform", help="Platform of the os to copy from (e.g. redhat)", required=True)
```
gmarciani: minor: we know the available source os, so we could specify the valid values, for src- options.

util/os-resources.py
```diff
@@ -0,0 +1,84 @@
...
+        filedata = filedata.replace(f"platform: '{src_platform}'", f"platform: '{dst_platform}'")
+        filedata = filedata.replace(f"platform?('{src_platform}')", f"platform?('{dst_platform}')")
+        if f"{dst_platform}" not in filedata:
+            print(f"ERROR: Unable to replace platform in file {dst_path}")
+
+        # Replace platform in the dst file. There are 2 alternatives ways to find version
+        filedata = filedata.replace(f"to_i == {src_version}", f"to_i == {dst_version}")
+        filedata = filedata.replace(f"platform_version: '{src_version}'", f"platform_version: '{dst_version}'")
+
+        if f"to_i == {dst_version}" not in filedata and f"platform_version: '{dst_version}'" not in filedata:
```
gmarciani, line 83: I think it's safer to have an or condition rather than an and condition because we want  both to be replaced.

util/os-resources.py
```diff
@@ -0,0 +1,84 @@
+# !/usr/bin/python3
```
gmarciani, line 1: Minor: what about documenting an example of usage? It could be in REAMDE, in a comment on top of this file or, even better, included in the script helper

gmarciani (dismissed):

util/os-resources.py
enrico-usai: This check is for "version" only, that can be replaced with 2 different patterns, but only one of them is present in the files.

util/os-resources.py
enrico-usai: added, thanks for the suggestion!

util/os-resources.py
enrico-usai: added.

gmarciani (approved):

PR #2342, opened 2023-06-29
Create new mechanism to execute OS specific hooks

enrico-usai:
### Description of changes

The kitchen tests will automatically execute hooks from `kitchen/hooks` folder before executing the hooks for the specific test.

For example we can have a `kitchen/hooks/alinux2/pre_converge.sh` with the following content,
to automatically install some packages in the instance right after the startup, before Chef installation:

```
#!/bin/bash

if [ -n "${KITCHEN_INSTANCE_HOSTNAME}" ]; then
    export KITCHEN_EC2_USER='ec2-user'
    echo "Install some-package package by using SSH key: ${KITCHEN_SSH_KEY_PATH}"

    KITCHEN_EC2_INSTANCE_ID=$(ssh -o StrictHostKeyChecking=no -i "${KITCHEN_SSH_KEY_PATH}" \
      "${KITCHEN_EC2_USER}@${KITCHEN_INSTANCE_HOSTNAME}" 'sudo yum install -y some-package')
fi
```

This is useful for example when Chef cannot be installed in the AMI because due to a missing or
conflicting packages.

### Tests
* Tested it with an hook for Amazon Linux 2023, on which I wasn't able to install Chef.

### References
* TODO add ALinux2023 patch.

### Checklist

### DIFF
--- a/kitchen/kitchen.run-hook.sh
+++ b/kitchen/kitchen.run-hook.sh
@@ -3,27 +3,21 @@
 export KITCHEN_HOOK=$1
 KITCHEN_ROOT_DIR=$(readlink -f $2)
 export KITCHEN_ROOT_DIR
-
-echo "*** Run ${KITCHEN_HOOK} for cookbook ${KITCHEN_COOKBOOK_PATH} suite ${KITCHEN_SUITE_NAME} on host ${KITCHEN_INSTANCE_HOSTNAME}"
-
 THIS_DIR=$(dirname "$0")
 SCRIPT="${THIS_DIR}/../${KITCHEN_COOKBOOK_PATH}/test/hooks/${KITCHEN_PHASE}/${KITCHEN_SUITE_NAME}/${KITCHEN_HOOK}.sh"
-echo "${SCRIPT}"
-
 if [ -e "${SCRIPT}" ]; then
-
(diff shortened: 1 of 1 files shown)


dreambeyondorange (approved):

kitchen/kitchen.run-hook.sh
```diff
@@ -3,27 +3,33 @@
...
 export KITCHEN_ROOT_DIR
-
-echo "*** Run ${KITCHEN_HOOK} for cookbook ${KITCHEN_COOKBOOK_PATH} suite ${KITCHEN_SUITE_NAME} on host ${KITCHEN_INSTANCE_HOSTNAME}"
-
 THIS_DIR=$(dirname "$0")
-SCRIPT="${THIS_DIR}/../${KITCHEN_COOKBOOK_PATH}/test/hooks/${KITCHEN_PHASE}/${KITCHEN_SUITE_NAME}/${KITCHEN_HOOK}.sh"
-echo "${SCRIPT}"
 
+# Run OS specific hooks from kitchen/hooks folder
+SCRIPT="${THIS_DIR}/hooks/${KITCHEN_PLATFORM_NAME}/${KITCHEN_HOOK}.sh"
```
dreambeyondorange, line 9: I wonder if the OS specific hooks could be also test specific?

dreambeyondorange (approved):

kitchen/kitchen.run-hook.sh
enrico-usai: if it's test specific we can use the previous mechanism already in place in the same script.

PR #2345, opened 2023-07-03
[develop] Adapt base recipes to configure LoginNodes

NSsirena:
### Description of changes
Each one of the following has a dedicated commit to make it easier to understand the PR

* Small preliminary fixes
    * Define helper method to identify LoginNodes 
    * Remove _netdev check from raid volumes test
    * Add LoginNode role in CloudWatch agent configuration 

* Cluster User
    * Configure cluster user in LoginNodes
    * Test cluster user configuration in LoginNodes 

* EBS volumes
    * Mount EBS volumes in LoginNodes
    * Test ebs volume configuration in LoginNodes 

* raid volumes
    * Mount raid volumes in LoginNodes
    * Test raid volume configuration in LoginNodes 

* dedicated shared folder for login nodes
    * Create shared_login_nodes directory 
    * Export shared_login_nodes directory 
    * Test /opt/intel folder configuration in LoginNodes 

* mounting of home and shared folders
    * Mount home and shared_login_nodes directories in login nodes 
    * Extract mount /home as subrecipe 
    * Test shared volumes configuration in LoginNodes 

* Slurm related configs
    * Setup munge and mount /opt/slurm in login nodes 
    * Validate that /opt/slurm is mounted in Compute and Login nodes 

Note the above test should be extended to validate also the munge configuration recipe. I did an attempt to create a mock environment for the test but the munge service still wasn't starting. Since the recipe is working (as my manual tests showed) I considered out-of-scope for this PR spending further time on this mock environment.

* Fixes in the logic used to retrieve the cluster config
    * Define current and previous config path for LoginNodes 
    * Ensure cluster config is retrieved only by the HeadNode 
    * LoadClusterConfig take the config path as parameter 

### Tests
* Most of the changes were successfully tested with unit tests
* I also manually launched a cluster with these changes in the recipes plus others required in the parallelcluster package and validated that I was able to log into a login node and have basic interactions with slurm (querying the cluster config, submitting a job, querying job status and job result)

### References
* [Changes in parallelcluster package](https://github.com/aws/aws-parallelcluster/pull/5446).

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.

PR #2355, opened 2023-07-10
[develop] Export /opt/intel only when it exists.

NSsirena:
On some OS/architecture is not available, so we need a guard.

### Description of changes
* Export /opt/intel only when it exists.
  * On some OS/architecture is not available, so we need a guard.

This was changed in the linked PR since we expected this folder to be always created, but it wasn't.

### Tests
* N/A

### References
* [LoginNodes recipes](https://github.com/aws/aws-parallelcluster-cookbook/pull/2345)

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/recipes/config/shared_storages.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/config/shared_storages.rb
@@ -33,14 +33,15 @@ when 'HeadNode'
     action :export
   end
 
-  # Export /opt/intel
+  # Export /opt/intel only if exists
   volume "export /opt/intel" do
     shared_dir "/opt/intel"
+    only_if { ::File.directory?("/opt/intel") }
     action :export
   end
 
 when 'ComputeFleet', 'LoginNode'
-  # Mount /opt/intel over NFS
+  # Mount /opt/intel over NFS only if it exists
   exported_intel_dir = format_directory('/opt/intel')
   volume "mount /opt/intel" do
     action :mount
@@ -50,6 +51,7 @@ when 'ComputeFleet', 'LoginNode'
     options node['cluster']['nfs']['hard_mount_options']
     retries 10
     retry_delay 6
+    only_if { ::File.directory?("/opt/intel") }
   end
 
 else


cookbooks/aws-parallelcluster-environment/recipes/config/shared_storages.rb
```diff
@@ -50,6 +51,7 @@
     options node['cluster']['nfs']['hard_mount_options']
     retries 10
     retry_delay 6
+    only_if { ::File.directory?("/opt/intel") }
```
jdeamicis, line 54: I know this is not your fault, but this condition looks quite silly: we mount `/opt/intel` only if the `/opt/intel` folder already exists. Personally I would be happier if we put `only_if { !arm_instance? }` here.

Also, I just noticed the way the `volume` resource was implemented imposes the exported dir and the mount point to be the exact same path, which will annoy us very soon I believe.

cookbooks/aws-parallelcluster-environment/recipes/config/shared_storages.rb
```diff
@@ -33,14 +33,15 @@
...
+  # Export /opt/intel only if exists
   volume "export /opt/intel" do
     shared_dir "/opt/intel"
+    only_if { ::File.directory?("/opt/intel") }
     action :export
   end
 
 when 'ComputeFleet', 'LoginNode'
-  # Mount /opt/intel over NFS
+  # Mount /opt/intel over NFS only if it exists
```
jdeamicis, line 44: See below.

jdeamicis (approved):

cookbooks/aws-parallelcluster-environment/recipes/config/shared_storages.rb
NSsirena: This is the previous code guard and I restored this on purpose since it worked.

It may be that `only_if { !arm_instance? }` works, but it will cover only a corner case and it may easily break if tomorrow we add another one.
And unless we have a centralized point where we clearly express the conditions when this folder is created (that we can also use as documentation) I prefer this implementation.

As for the `volume` resource implementation .. I only used it. I'm still not sure it is really useful since it is nearly a simple wrapper of an existing resource ..but let's give it a try...

PR #2366, opened 2023-07-17
Add login nodes daemon service in cookbook

hehe7318:
### Description of changes
* Create login_nodes_on_termination_script.sh.erb in slurm template
  * prevent further login/SSH attempts by updating ssh config
  * broadcast a message to all logged in users using the wall command
  * sleep for the gracetime period
* Create login_nodes_daemon_script.sh.erb in slurm template
  * check instance target state every 60 seconds, when target state is terminated, run on termination script
* Create login_nodes_damon_config.json.erb in slurm template
  * contains termination_script_path and termination_message
* Create login_nodes_damon_service.rb in slurm receipt
  * create configuration file in shared directory
  * create on_termination_script.sh in shared directory
  * create daemon_script.sh in shared directory
  * create supervisord configuration for the daemon

### Checklist


cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -281,3 +281,5 @@
   retries 5
   retry_delay 2
 end unless redhat_on_docker?
+
+include_recipe 'aws-parallelcluster-slurm::login_nodes_daemon_service'
```
gmarciani: The daemon should be created in login nodes, not in head node.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
+require 'yaml'
+config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
```
gmarciani: The file may not exist yet at the time this function gets called.
A way to overcome this issue is to lazily reference cluster config, this way from within a resource block.

```
lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :GracetimePeriod) }
```


See example: https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb#L39-L39

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+GRACETIME_PERIOD = config["LoginNodes"]["Pools"][0]["GRACETIME_PERIOD"]
```
gmarciani: Typo: the gracetime period property is not uppercase in the config file, but should rather be camelcase

gmarciani: Furthermore, what if the property is not specified in the cluster config because it's meant to be the default value?
Would this reference raise a KayError?

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+  group 'root'
+  mode '0755'
+end
+
+# Create daemon_script.sh
+template "#{node['cluster']['shared_dir']}/login_nodes_daemon_script.sh" do
+  source 'login_nodes_daemon_service/login_nodes_daemon_script.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0755'
```
gmarciani: Here and in every script defined in this PR that is meant to be executed by root.

`0755` is too permissive as it would grant execution access to everyone.
You should rather use `0744` so that root will have read-write-execute access and the others would have only read access.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_config.json.erb
```diff
@@ -0,0 +1,4 @@
+{
+  "termination_script_path": "<%= node['cluster']['shared_dir'] %>/login_nodes_on_termination_script.sh",
+  "termination_message": "System will be terminated after <%= GRACETIME_PERIOD %> minutes."
```
gmarciani: I think that specifying the gracetime period in the termination message is misleading because the user would not now when exactly the system will shut down.
We have two options here:
1. Leave the message generic: "The system will be terminated in less than ${GRACETIME_PERIOD} minutes"
2. Be more specific: "The system will be terminated on ${TERMINATION_DATE_TIME}", where *TERMINATION_DATE_TIME* is computed at runtime when the termination is scheduled.

Since we are close to the deadline, if you find hard to inject the termination date and time, I suggest to go with the generic approach and refine the solution in a post-beta follow up PR.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
gmarciani: Minor: We can shorten the file name as 'login_nodes_on_termination.sh.erb', omitting the term 'script' because it's implicit in the file extension that this file is a script.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
```diff
@@ -0,0 +1,33 @@
...
+
+if [[ $OS == "amzn" ]]; then
+    DEFAULT_USER='ec2-user'
+elif [[ $OS == "centos" ]]; then
+    DEFAULT_USER='centos'
+elif [[ $OS == "ubuntu" ]]; then
+    DEFAULT_USER='ubuntu'
+elif [[ $OS == "rhel" ]]; then
+    DEFAULT_USER='ec2-user'
+fi
```
gmarciani: You can omit this logic and inject the cluster user name from the node attribute `node['cluster']['cluster_user']`

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
```
gmarciani: Minor: it's a good practice to put an introductory comment that explains what this script does and howe it should be used

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
```diff
@@ -0,0 +1,33 @@
...
+# Prevent further login/SSH attempts by updating ssh config
+echo "AllowUsers $DEFAULT_USER" >> /etc/ssh/sshd_config
+
+# Reload the ssh configuration
+systemctl reload sshd
+
+# Broadcast a message to all logged in users using the wall command
+wall "$TERMINATION_MESSAGE"
+
+# Sleep for the gracetime period
```
gmarciani: *Minor*
It's a good practice to put comments on code section explaining what the code does.
Nevertheless, comments are expected to add informative value over the code they refer to; so if the code is self-explanatory, you can also avoid putting comments on every line of code.

In this specific case, for example, you can describe what the script is expected to do ion the introductory comment for the whole script and remove the comments on every pother line.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
```
gmarciani: *Minor*
It's a good practice to document the scope and usage of a script with an introductory comment

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
...
+# Load configuration file
+CONFIG_PATH="<%= node['cluster']['shared_dir'] %>/login_nodes_daemon_config.json"
+CONFIG_JSON=$(cat $CONFIG_PATH)
+TERMINATION_SCRIPT_PATH=$(echo $CONFIG_JSON | jq -r .termination_script_path)
+TERMINATION_MESSAGE=$(echo $CONFIG_JSON | jq -r .termination_message)
+GRACETIME_PERIOD=$1
+
+IMDS_TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
```
gmarciani: This token is going to expire in 6h after being retrieved and according to the code below it is expected to be reused for the whole life of the daemon. In this way the daemon is going to fail the IMDs call after 6h.

You should retrieve the IMDS token with a shorter expiration time from within the loop right before you're going to use it.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
...
+
+while true; do
+  INSTANCE_LIFECYCLE_STATE=$(curl -H "X-aws-ec2-metadata-token: $IMDS_TOKEN" -v http://169.254.169.254/latest/meta-data/instance-life-cycle)
+  if [[ "$INSTANCE_LIFECYCLE_STATE" == "Terminated" ]]; then
+    break
+  fi
+  sleep 60
```
gmarciani: *Minor*
It's a good practice to avoid magic numbers. This is not only for better readability but also to improve maintainability.
Let's avoid putting 60 here and define a dedicated constant with a self explanatory name, e.g. `DAEMON_SLEEP_SECONDS`

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Load configuration file
+CONFIG_PATH="<%= node['cluster']['shared_dir'] %>/login_nodes_daemon_config.json"
+CONFIG_JSON=$(cat $CONFIG_PATH)
+TERMINATION_SCRIPT_PATH=$(echo $CONFIG_JSON | jq -r .termination_script_path)
+TERMINATION_MESSAGE=$(echo $CONFIG_JSON | jq -r .termination_message)
+GRACETIME_PERIOD=$1
```
gmarciani: Why the gracetime period is expected to be injected as a script argument while the other parameters are passed through a config file?
what about passing everything using the config file?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
```diff
@@ -0,0 +1,33 @@
...
+sleep $((GRACETIME_PERIOD*60))
+
+# Broadcast a final message
+wall "System is now going down for termination!"
+
+exit 0
```
gmarciani: `exit 0` is redundant. A script always returns with exit code 0 unless differently specified.
What about removing it?

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
+require 'yaml'
+config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
+GRACETIME_PERIOD = config["LoginNodes"]["Pools"][0]["GRACETIME_PERIOD"]
+
+# Create configuration file
+template "#{node['cluster']['shared_dir']}/login_nodes_daemon_config.json" do
```
gmarciani: Every resource (script, config file, ..) that is expected to be shared among head and login nodes should be placed in `node['cluster']['shared_dir_login_nodes']` because , otherwise, you would share dlogin nodes specific resource with compute nodes.

This is not a functional issue, but we improve the security posture of our product by limiting privileges and access.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+template "#{node['cluster']['shared_dir']}/login_nodes_daemon_script.sh" do
+  source 'login_nodes_daemon_service/login_nodes_daemon_script.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0755'
+end
+
+# Create supervisord configuration for the daemon
+file "/etc/supervisor/conf.d/login_nodes_daemon_service.conf" do
+  content <<-EOF
```
gmarciani: It would be better to define this configuration within a templated file and simply call the template resource here.
This approach would improve maintainability and readability.

gmarciani:
The PR is currently missing unit tests. We should add them. You can refer to `cookbooks/aws-parallelcluster-slurm/test/controls/slurm_config_spec.rb` to have an example.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
+require 'yaml'
+config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
+GRACETIME_PERIOD = config["LoginNodes"]["Pools"][0]["GRACETIME_PERIOD"]
+
+# Create configuration file
+template "#{node['cluster']['shared_dir']}/login_nodes_daemon_config.json" do
+  source 'login_nodes_daemon_service/login_nodes_daemon_config.json.erb'
```
gmarciani: Here and below. The source is wrong because the template path must be relative to the folder `templates/default`. So it should be `slurm/login_nodes_daemon_service/...`

gmarciani (changes requested):

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_config.json.erb
```diff
@@ -0,0 +1,4 @@
+{
```
gmarciani: I suggest to keep the directory structure consistent with the rest of the project.
I suggest to replace `templates/default/slurm/login_nodes_daemon_service` with `templates/default/slurm/login`

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+  group 'root'
+  mode '0755'
+end
+
+# Create supervisord configuration for the daemon
+file "/etc/supervisor/conf.d/login_nodes_daemon_service.conf" do
+  content <<-EOF
+[program:login_nodes_daemon_service]
+command=bash #{node['cluster']['shared_dir']}/login_nodes_daemon_script.sh #{GRACETIME_PERIOD}
+autostart=true
```
gmarciani: Better to also specify the following properties:

```
user = <%= node['cluster']['cluster_admin_user'] %>
redirect_stderr = true
stdout_logfile = /var/log/parallelcluster/login_nodes_daemon
```

gmarciani (changes requested):

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
+GRACETIME_PERIOD = config["LoginNodes"]["Pools"][0]["GRACETIME_PERIOD"]
+
+# Create configuration file
+template "#{node['cluster']['shared_dir']}/login_nodes_daemon_config.json" do
+  source 'login_nodes_daemon_service/login_nodes_daemon_config.json.erb'
+  owner 'root'
+  group 'root'
+  mode '0644'
+  variables(
```
gmarciani, line 7: variables accepts key arguments, not a dictionary. So, this should be:

```
variables(
  GRACETIME_PERIOD: ....
)
```

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_config.json.erb
gmarciani: Injected variables must be references with `@`, i.e. `@GRACETIME_PERIOD`.
Furthermore, our convention is to use lowercase variable names for this type of injection.

gmarciani (changes requested):

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+# Create daemon_script.sh
+template "#{node['cluster']['shared_dir']}/login_nodes_daemon_script.sh" do
+  source 'login_nodes_daemon_service/login_nodes_daemon_script.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0755'
+end
+
+# Create supervisord configuration for the daemon
+file "/etc/supervisor/conf.d/login_nodes_daemon_service.conf" do
```
gmarciani: We have a central supervisord configuration in `cookbooks/aws-parallelcluster-platform/templates/supervisord/parallelcluster_supervisord.conf.erb`.

I like your approach, but:
1. That way of configuration is not documented, see http://supervisord.org/configuration.html
2. That way of configuration seems not portable, see last answer in https://stackoverflow.com/questions/12226113/supervisor-conf-default-location
3. Even if it was supported, it would be better to be consistent with the approach we adopted so far; in case we disagree with the current approach, better to be consistent and open a follow up activity to revise it given we are close to the deadline.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
```diff
@@ -0,0 +1,54 @@
...
+autostart=true
+autorestart=true
+stopasgroup=true
+killasgroup=true
+  EOF
+end
+
+# Update supervisor configuration
+execute 'supervisorctl reread' do
+  command 'supervisorctl reread'
```
gmarciani: Herer and below.

`supervisorctl` is not installed system-wide, but within the cookbook poython virtual environmet.

So you should rather do:

```
execute 'start login_nodes_daemon_service' do
  command "#{cookbook_virtualenv_path}/bin/supervisorctl start login_nodes_daemon_service"
end
```

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
gmarciani: This daemon and the script it calls does not echo any information about its execution. Let's think about useful info to be logged and we will make supervisord log into a dedicated log file

gmarciani: I suggest to execute this script with errexit and xtrace mode, i.e. adding `set -ex` right after the shebang.
With the errexit mode the script is gonna exit immeditelyt with failure if an inner command (pipelined or compiund) fails.

With the xtrace mode is like debugging mode for bash scripts: the script will log the executed commands together with their outputs.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
gmarciani: The autorestart policy strictly depends on the way the program has been implemented. 
In this specific case, the program should not be auto-restarted if it exits with exit code 0, otherwise it would send multiple notifications.

So we have two options here:
1. leave the daemon logic as it is, but make supervisord restart the process only if it exits with exti code != 0
2. leave supervisord always restarting the process as it currently does, but change the daemon logic in a way that it avoids sending multiple notifications

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
...
+CONFIG_PATH="<%= node['cluster']['shared_dir'] %>/login_nodes_daemon_config.json"
+CONFIG_JSON=$(cat $CONFIG_PATH)
+TERMINATION_SCRIPT_PATH=$(echo $CONFIG_JSON | jq -r .termination_script_path)
+TERMINATION_MESSAGE=$(echo $CONFIG_JSON | jq -r .termination_message)
+GRACETIME_PERIOD=$1
+
+IMDS_TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+
+while true; do
+  INSTANCE_LIFECYCLE_STATE=$(curl -H "X-aws-ec2-metadata-token: $IMDS_TOKEN" -v http://169.254.169.254/latest/meta-data/instance-life-cycle)
```
gmarciani: This is the wrong IMDS metadata path.
You should rather use: `http://169.254.169.254/latest/meta-data/autoscaling/target-lifecycle-state`

See doc: https://docs.aws.amazon.com/autoscaling/ec2/userguide/retrieving-target-lifecycle-state-through-imds.html

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
```diff
@@ -0,0 +1,20 @@
...
+  if [[ "$INSTANCE_LIFECYCLE_STATE" == "Terminated" ]]; then
+    break
+  fi
+  sleep 60
+done
+
+bash $TERMINATION_SCRIPT_PATH $TERMINATION_MESSAGE $GRACETIME_PERIOD
```
gmarciani: You should enclose bash variables within double quotes to prevent unexpected behaviour on argument parsing.
This is valid in general. In this specific case it's strictly required for argument TERMINATION_MESSAGE because the message is a string containing blank spaces

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Done! Thank you!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: I don't think so, because I saw the cluster-config.yaml file in shared directory in instance. It's like:
```
LoginNodes:
  Pools:
  - Count: 2
    GracetimePeriod: 6
    Iam:
      AdditionalIamPolicies: []
      InstanceProfile: null
      InstanceRole: null
    Image: null
    InstanceType: t2.micro
    Name: test
    Networking:
      AdditionalSecurityGroups: null
      Proxy: null
      SecurityGroups: null
      SubnetIds:
      - subnet-0c45c7238162199db
    Ssh:
      KeyName: xuanqi-us-east-2
```
So it will show the entire config yaml and get the default value of 60 minutes.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Have fixed it! Thank you.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_config.json.erb
hehe7318: That makes sense. I use the generic approach for now.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
hehe7318: Done!

hehe7318: Also done.-> shorten the `login_nodes_daemon_script.sh.erb` as `login_nodes_daemon.sh.erb`

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
hehe7318: Thank you for your reminding. Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
hehe7318: Done for daemon and on_termination scripts.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
hehe7318: Thanks for reminding. Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Done.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_on_termination_script.sh.erb
hehe7318: Makes sense. Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Thank you! I have corrected it.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_config.json.erb
hehe7318: Okay. Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Thank you, I have corrected it.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_config.json.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Thanks for reminding, I have corrected it.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: Have modified it, thank you for reminding!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Have added it.

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
hehe7318: I use the first option to do the change:
leave the daemon logic as it is, but make supervisord restart the process only if it exits with exti code != 0

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/login_nodes_daemon_service/login_nodes_daemon_script.sh.erb
hehe7318: Done! Thank you!

cookbooks/aws-parallelcluster-slurm/recipes/config/login_nodes_daemon_service.rb
gmarciani: We need the following configurations:

```
exitcodes = 0
autostart = true
autorestart = unexpected
```

With autorestart = unexpected, supervisord will restart the daemon only if it exits with unexpected exit codes, i.e. in this case code != 0.

gmarciani:
Closed in favour of https://github.com/aws/aws-parallelcluster-cookbook/pull/2368

PR #2391, opened 2023-08-03
Move the NFS config file override from the setup action to configure via override_server_template

dreambeyondorange:
The override_server_template function runs during bootstrapping to set the parameter needed for the OS specific config file. The setup action works locally since that action runs during local tests, but does not work during bootstrapping when the node attributes from the setup action are not present.

### Description of changes
* Describe *what* you're changing and *why* you're doing these changes.

### Tests
* Tested `nfs_headnode` locally with kitchen

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_ubuntu22+.rb
+++ b/cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_ubuntu22+.rb
@@ -29,11 +29,11 @@ action :setup do
   action_install_nfs
   action_install_nfs4
   action_disable_start_at_boot
-  node.default['nfs']['config']['server_template'] = '/etc/nfs.conf.d/parallelcluster-nfs.conf'
 end
 
 action_class do
   def override_server_template
+    node.default['nfs']['config']['server_template'] = '/etc/nfs.conf.d/parallelcluster-nfs.conf'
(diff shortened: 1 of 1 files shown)


cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_ubuntu22+.rb
```diff
@@ -29,11 +29,11 @@
   action_install_nfs
   action_install_nfs4
   action_disable_start_at_boot
-  node.default['nfs']['config']['server_template'] = '/etc/nfs.conf.d/parallelcluster-nfs.conf'
 end
 
 action_class do
   def override_server_template
+    node.default['nfs']['config']['server_template'] = '/etc/nfs.conf.d/parallelcluster-nfs.conf'
```
enrico-usai, line 36: nice catch!

enrico-usai (dismissed):

cookbooks/aws-parallelcluster-environment/resources/nfs/nfs_ubuntu22+.rb
enrico-usai: Probably the alternative was to move this override outside of the action class, to be "global".

enrico-usai (approved):

PR #2434, opened 2023-08-28
[develop] Update cinc-install.sh file to latest version

enrico-usai:
This new file should be referenced from CI/CD with version 1.2.0.

In the second commit I'm [adding custom changes on top of cinc installer to download artifacts from s3](https://github.com/aws/aws-parallelcluster-cookbook/pull/2434/commits/2150d97ca487b52abcfb36cd5e420a0bcb30c6d4) 

Changes added on top of the original file (downloaded from https://cinc.sh/download/) are:
 - changed format from `sh` to `bash`;
 - installer downloaded from ParallelCluster S3 bucket rather than from omnitruck website;
 - custom functions to retrieve AWS region and domain;
 - new `-b` option to permit to pass an optional bucket for downloading a custom Cinc installer;
 - unused options `-c`, `-p`, `-n`, setting the unused `$channel` variable has been removed;
 - improvement in the dpkg installation to avoid conflicts with other running installations;
 - added custom case to manage download of ubuntu packages from S3.

When updating this modified file, we need to remember to bump the version and reference it in the CI/CD.
 - cinc-install.sh v1.2.0

### References

File downloaded from https://cinc.sh/download/

### Tests
* Tested in first and second stage build by passing a custom installer url and using cinc 18.2.7

### DIFF
--- a/util/cinc-install.sh
+++ b/util/cinc-install.sh
@@ -1,5 +1,19 @@
-#!/bin/sh
-# WARNING: REQUIRES /bin/sh
+#!/bin/bash
+# WARNING: this file is a modified version of the installer from https://cinc.sh/download/
+#
+# Notable changes are:
+# - changed format from sh to bash;
+# - installer downloaded from ParallelCluster S3 bucket rather than from omnitruck website;
+# - custom functions to retrieve AWS region and domain;
+# - new -b option to permit to pass an optional bucket for downloading a custom Cinc installer;
+# - unused options -c, -p, -n, setting the unused $channel variable has been removed;
+# - added custom case to manage download of ubuntu packages from S3.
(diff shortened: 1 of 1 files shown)


util/cinc-install.sh
```diff
@@ -586,20 +603,6 @@ case $machine in
     ;;
 esac
 
-if test "$platform" = "ubuntu"; then
```
lukeseawalker, line 589: I think this if should be kept, depending on how the deb package is named, e.g. if it named after arm64.deb or amd64.deb

enrico-usai: re-added, thanks. 
This was not required when downloading from omnitruck because the url to use is in the metadata file.

lukeseawalker (approved):

PR #2443, opened 2023-09-01
[develop] Setup custom munge key

hehe7318:
### Description of changes
* Create a munge key manager
* Add logic to fetch the custom munge key, decode it and use it in the cluster
* Add related kitchen tests
* Create a custom munge key rotation script for customer to manually fetch munge key from secrets manager in case the secret ARN doesn't change but the value changes
* Create a munge key update manager
* Add logic to update the custom munge key -> Adding new, changing, removing the munge key secret Arn in YAML file

### Relate Pull Request
* See the PR of Setup custom munge key in cli: [[develop] Cli Setup custom munge key](https://github.com/aws/aws-parallelcluster/pull/5681)

### New sample YAML configuration

```
DevSettings:
  SlurmSettings:
    MungeKeySecretArn: Str
```


cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -0,0 +1,40 @@
...
+action :manage do
+  if new_resource.munge_key_secret_arn
+    # This block will fetch the munge key from Secrets Manager
+    bash 'fetch_and_decode_munge_key' do
+      user 'root'
+      group 'root'
+      cwd '/tmp'
+      code <<-FETCH_AND_DECODE
+        # Get encoded munge key from secrets manager and decode it
+        encoded_key=$(aws secretsmanager get-secret-value --secret-id #{new_resource.munge_key_secret_arn} --query 'SecretString' --output text)
```
EddyMM: What if there's an error fetching the secret? This will not return the value.
Can you check if `encoded_key` has a value and possibly echo and error & exit early before base64 decoding it?

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -0,0 +1,40 @@
+resource_name :munge_key_manager
+provides :munge_key_manager
+unified_mode true
+
+property :munge_key_secret_arn, String
+
+default_action :manage
+
+action :manage do
```
EddyMM: Can we use a more descriptive action e.g. `:setup_munge_key`?

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -0,0 +1,57 @@
...
+  if new_resource.munge_key_secret_arn
+    # This block will fetch the munge key from Secrets Manager
+    bash 'fetch_and_decode_munge_key' do
+      user 'root'
+      group 'root'
+      cwd '/tmp'
+      code <<-FETCH_AND_DECODE
+        # Get encoded munge key from secrets manager and decode it
+        encoded_key=$(aws secretsmanager get-secret-value --secret-id #{new_resource.munge_key_secret_arn} --query 'SecretString' --output text)
+        echo $encoded_key | base64 -d > /etc/munge/munge.key
```
EddyMM: Same here, what if base64 decoding fails? We need to add the base64 decoded key to `/etc/munge/munge.key` only if base64 decoding succeeds, otherwise, echo an error and exit with non-zero.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Yes, done!

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Sure!

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Done!

PR #2445, opened 2023-09-05
Add unit tests for cookbook EBS related scripts

judysng:
### Description of changes

- Wrote unit tests for functions ec2_dev_2_volid and manageVolume because these files lacked unit tests
- Separated the main function in ec2_dev_2_volid into multiple functions
- Fixed typo in manageVolume

### Tests

- Added tests for most functions in the two files except for some functions that just called the other tested functions

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
+++ b/cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
@@ -41,17 +41,7 @@ def validate_device_name(device_name):
     return True
 
 
-def main():
-    syslog.syslog("Starting ec2_dev_2_volid.py script")
-    # Get dev
-    try:
-        dev = str(sys.argv[1])
-        validate_device_name(dev)
-        syslog.syslog(f"Input block device is {dev}")
-    except IndexError:
-        syslog.syslog(syslog.LOG_ERR, "Provide block device i.e. xvdf")
-
-    # Convert dev to mapping format
+def adapt_device_name(dev):
     if "nvme" in dev:
         # For newer instances which expose EBS volumes as NVMe devices, translate the
         # device name so boto can discover it.
@@ -70,26 +60,10 @@ def main():
     else:
         dev = dev.replace("xvd", "sd")
         dev = "/dev/" + dev
+    return dev
 
-    # Get IMDSv2 token
-    token = get_imdsv2_token()
 
-    # Get instance ID
-    instance_id = requests.get(
-        "http://169.254.169.254/latest/meta-data/instance-id",
-        headers=token,
-        timeout=METADATA_REQUEST_TIMEOUT,
-    ).text
-
-    # Get region
-    region = requests.get(
-        "http://169.254.169.254/latest/meta-data/placement/availability-zone",
-        headers=token,
-        timeout=METADATA_REQUEST_TIMEOUT,
-    ).text
-    region = region[:-1]
-
-    # Parse configuration file to read proxy settings
+def parse_proxy_config():
     config = configparser.RawConfigParser()
     config.read("/etc/boto.cfg")
     proxy_config = Config()
@@ -97,18 +71,10 @@ def main():
         proxy = config.get("Boto", "proxy")
         proxy_port = config.get("Boto", "proxy_port")
         proxy_config = Config(proxies={"https": f"{proxy}:{proxy_port}"})
+    return proxy_config
 
-    # Configure the AWS CA bundle.
-    # In US isolated regions the dedicated CA bundle will be used.
-    # In any other region, the default bundle will be used (None stands for the default settings).
-    # Note: We want to apply a more general solution that applies to every region,
-    # but for the time being this is enough to support US isolated regions without
-    # impacting the other ones.
-    ca_bundle = f"/etc/pki/{region}/certs/ca-bundle.pem" if region.startswith("us-iso") else None
 
-    # Connect to AWS using boto
-    ec2 = boto3.client("ec2", region_name=region, config=proxy_config, verify=ca_bundle)
-
+def get_device_volume_id(ec2, dev, instance_id):
     # Poll for blockdevicemapping
     devices = ec2.describe_instance_attribute(InstanceId=instance_id, Attribute="blockDeviceMapping").get(
         "BlockDeviceMappings"
@@ -127,8 +93,48 @@ def main():
         dev_map = dict((d.get("DeviceName"), d) for d in devices)
         loop_count += 1
 
-    # Return volume ID
-    volume_id = dev_map.get(dev).get("Ebs").get("VolumeId")
+    return dev_map.get(dev).get("Ebs").get("VolumeId")
+
+
+def get_metadata_value(token, metadata_path):
+    return requests.get(
+        metadata_path,
+        headers=token,
+        timeout=METADATA_REQUEST_TIMEOUT,
+    ).text
+
+
+def main():
+    syslog.syslog("Starting ec2_dev_2_volid.py script")
+    try:
+        dev = str(sys.argv[1])
+        validate_device_name(dev)
+        syslog.syslog(f"Input block device is {dev}")
+    except IndexError:
+        syslog.syslog(syslog.LOG_ERR, "Provide block device i.e. xvdf")
+
+    dev = adapt_device_name(dev)
+
+    token = get_imdsv2_token()
+
+    instance_id = get_metadata_value(token, "http://169.254.169.254/latest/meta-data/instance-id")
+
+    region = get_metadata_value(token, "http://169.254.169.254/latest/meta-data/placement/availability-zone")
+    region = region[:-1]
+
+    proxy_config = parse_proxy_config()
+
+    # Configure the AWS CA bundle.
+    # In US isolated regions the dedicated CA bundle will be used.
+    # In any other region, the default bundle will be used (None stands for the default settings).
+    # Note: We want to apply a more general solution that applies to every region,
+    # but for the time being this is enough to support US isolated regions without
+    # impacting the other ones.
+    ca_bundle = f"/etc/pki/{region}/certs/ca-bundle.pem" if region.startswith("us-iso") else None
+
+    ec2 = boto3.client("ec2", region_name=region, config=proxy_config, verify=ca_bundle)
+
+    volume_id = get_device_volume_id(ec2, dev, instance_id)
     print(volume_id)
 
 
--- a/cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/manageVolume.py
+++ b/cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/manageVolume.py
@@ -32,7 +32,7 @@ def validate_device_name(device_name):
     return True
 
 
-def convert_dev(dev):
+def adapt_device_name(dev):
     # Translate the device name as provided by the OS to the one used by EC2
     # FIXME This approach could be broken in some OS variants, see  # pylint: disable=fixme
     # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html#identify-nvme-ebs-device
@@ -91,7 +91,7 @@ def get_imdsv2_token():
 
 def attach_volume(volume_id, instance_id, ec2):
     # Generate a list of system paths minus the root path
-    paths = [convert_dev(device) for device in get_all_devices()]
+    paths = [adapt_device_name(device) for device in get_all_devices()]
 
     # List of possible block devices
     block_devices = [
@@ -175,7 +175,6 @@ def detach_volume(volume_id, ec2):
 
 
 def parse_proxy_config():
-    """Parse configuration file to read proxy settings."""
     config = configparser.RawConfigParser()
     config.read("/etc/boto.cfg")
     proxy_config = Config()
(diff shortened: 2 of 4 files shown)


himani2411:
Failing tox and code-linter errors

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -70,6 +60,55 @@ def main():
...
+    config.read("/etc/boto.cfg")
+    proxy_config = Config()
+    if config.has_option("Boto", "proxy") and config.has_option("Boto", "proxy_port"):
+        proxy = config.get("Boto", "proxy")
+        proxy_port = config.get("Boto", "proxy_port")
+        proxy_config = Config(proxies={"https": f"{proxy}:{proxy_port}"})
+    return proxy_config
+
+
+def device_volume(ec2, dev, instance_id):
```
himani2411: Nitpick: Rename the function to something appropriate like `get_device_volume_id()` or `retrieve_device_volume_id `

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -110,26 +143,7 @@ def main():
     ec2 = boto3.client("ec2", region_name=region, config=proxy_config, verify=ca_bundle)
 
     # Poll for blockdevicemapping
```
himani2411: Nitpick: Change the comments as this function will not be just polling for Device Mappings. Add a comment for what we expect from this function and move this comment to appropriate place in `get_device_volume_id ()` or `retrieve_device_volume_id()`

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/manageVolume.py
```diff
@@ -22,7 +22,7 @@ def validate_device_name(device_name):
 
     The validation is done after forcing the encoding to be the standard Python Unicode / UTF-8
     :param device_name: an argument string to validate
-    :raise: Exception if the argument fails to match the patter
+    :raise: Exception if the argument fails to match the pattern
```
himani2411: Good Job on catching the spelling errors in this file!

test/unit/ec2_udev_rules/test_ec2_dev_2_volid.py
```diff
@@ -31,3 +31,112 @@ def test_validate_device_name(device_name, raises):
         assert_that(validate_device_name).raises(ValueError).when_called_with(device_name).contains("invalid pattern")
     else:
         assert_that(validate_device_name(device_name)).is_true()
+
+
+def test_get_imdsv2_token(mocker):
+    mock = mocker.Mock()
+    mocker.patch("requests.put", mock)
+    mock.return_value.status_code = 200
+    mock.return_value.content = {"key": "value"}
```
himani2411: Can you use the parameterization for the Value that you are testing it with. Even if its a Single value of `{"key": "value"}`. Try to use actual_value and expected _value parameters in this test like you have used in others.

test/unit/ec2_udev_rules/test_ec2_dev_2_volid.py
```diff
@@ -31,3 +31,112 @@ def test_validate_device_name(device_name, raises):
...
+        ("nvme0n1", "sdf"),
+        ("nvme0n1p1", "sdf"),
+        ("nvme0n1p128", "sdf"),
+        ("xvd0n1", "/dev/sd0n1"),
+        ("xvd0n1p1", "/dev/sd0n1p1"),
+    ],
+)
+def test_convert_dev(mocker, dev, expected_name, capsys):
+    mocker.patch("os.popen", mocker.mock_open(read_data=":sdf"))
+    if "nvme" in dev:
```
himani2411: This condition is already part of the `convert_dev()` function. You dont need to add this in your test itself. This test is supposed to test that condition by capturing the outputs from that test. You can use a different function for the ones which will raise an error.

himani2411: Or use another paramter which will mention that you expect an Exception like done [here](https://github.com/aws/aws-parallelcluster-cookbook/blob/492c3247e557765729f4eaf3d15bc26cedace7a2/test/unit/ec2_udev_rules/test_ec2_dev_2_volid.py#L18)

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -70,6 +60,55 @@ def main():
     else:
         dev = dev.replace("xvd", "sd")
         dev = "/dev/" + dev
+    return dev
+
+
+def parse_proxy_config():
```
himani2411, line 66: Are we not testing this function?

test/unit/ec2_udev_rules/test_ec2_dev_2_volid.py
```diff
@@ -31,3 +31,112 @@ def test_validate_device_name(device_name, raises):
...
+)
+def test_device_volume(ec2, dev, block, output_value, capsys):
+    ec2.describe_instance_attribute.return_value = block
+    if output_value == SystemExit:
+        with pytest.raises(SystemExit) as e:
+            device_volume(ec2, dev, 1)
+            assert_that(e.type).matches(SystemExit)
+            assert_that(e.value.code).is_equal_to(1)
+    else:
+        device_volume(ec2, dev, 1)
```
himani2411: You can add this line as the common as its repeated in Else condition too

test/unit/ec2_udev_rules/test_manage_volume.py
```diff
@@ -31,3 +43,165 @@ def test_validate_device_name(device_name, raises):
...
+        Exception,
+    ],
+)
+def test_is_volume_attached(ec2, volume_output, capsys):
+    ec2.describe_volumes.return_value = volume_output
+    is_attached = is_volume_attached(ec2, 0)
+
+    if volume_output == Exception:
+        captured = capsys.readouterr()
+        assert_that(captured.out).contains("exception")
```
himani2411, line 192: Can we check what is being printed when we have an exception

test/unit/ec2_udev_rules/test_manage_volume.py
```diff
@@ -31,3 +43,165 @@ def test_validate_device_name(device_name, raises):
...
+    ],
+)
+def test_is_volume_available(ec2, volume_output, capsys):
+    ec2.describe_volumes.return_value = volume_output
+    is_avail = is_volume_available(ec2, 0)
+
+    if volume_output == Exception:
+        captured = capsys.readouterr()
+        assert_that(captured.out).contains("exception")
+        assert_that(is_avail).is_false()
```
himani2411, line 172: Use the True and False value as an `expected_output` parameter in your test. Then you wont have to use condition where you check if volume_output has available or not

test/unit/ec2_udev_rules/test_manage_volume.py
```diff
@@ -31,3 +43,165 @@ def test_validate_device_name(device_name, raises):
...
+)
+def test_is_volume_attached(ec2, volume_output, capsys):
+    ec2.describe_volumes.return_value = volume_output
+    is_attached = is_volume_attached(ec2, 0)
+
+    if volume_output == Exception:
+        captured = capsys.readouterr()
+        assert_that(captured.out).contains("exception")
+        assert_that(is_attached).is_false()
+    elif volume_output.get("Volumes")[0].get("State") == "in-use":
```
himani2411: Use an `expected_output` paramter similar to `volume_output` to avoid checking these conditions.

himani2411 (changes requested):

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
judysng: I decided not to test this function because it seemed to just be calling lots of other functions without much logic in that function itself. Let me know if you think I should test this too though!

test/unit/ec2_udev_rules/test_manage_volume.py
judysng: Yes!

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -41,17 +41,7 @@ def validate_device_name(device_name):
...
-    # Get dev
-    try:
-        dev = str(sys.argv[1])
-        validate_device_name(dev)
-        syslog.syslog(f"Input block device is {dev}")
-    except IndexError:
-        syslog.syslog(syslog.LOG_ERR, "Provide block device i.e. xvdf")
-
-    # Convert dev to mapping format
+def convert_dev(dev):
```
enrico-usai: the name of this function could be: `adapt_device_name` in this way it's more aligned with existing `validate_device_name`

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -132,5 +98,54 @@ def main():
...
+        dev = str(sys.argv[1])
+        validate_device_name(dev)
+        syslog.syslog(f"Input block device is {dev}")
+    except IndexError:
+        syslog.syslog(syslog.LOG_ERR, "Provide block device i.e. xvdf")
+
+    # Convert dev to mapping format
+    dev = convert_dev(dev)
+
+    # Get IMDSv2 token
```
enrico-usai: Minor: You used the right self-explicative name for the function so you can avoid the comment.

This kind of comments are useless since in the comment you're writing the name of the function you're calling "get_imdsv2_token".

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -132,5 +98,54 @@ def main():
...
+
+    # Get instance ID
+    instance_id = requests.get(
+        "http://169.254.169.254/latest/meta-data/instance-id",
+        headers=token,
+        timeout=METADATA_REQUEST_TIMEOUT,
+    ).text
+
+    # Get region
+    region = requests.get(
```
enrico-usai: you could create a generic `get_metadata_value(token, metadata_path)` and reuse it for both instance-id and placement/availability-zone

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -132,5 +98,54 @@ def main():
...
+        "http://169.254.169.254/latest/meta-data/placement/availability-zone",
+        headers=token,
+        timeout=METADATA_REQUEST_TIMEOUT,
+    ).text
+    region = region[:-1]
+
+    # Parse configuration file to read proxy settings
```
enrico-usai: here as well the comment is useless.
You could write "why" you're doing it, but you don't need to explain "what" you're doing because the function name is self explicative.

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_dev_2_volid.py
```diff
@@ -132,5 +98,54 @@ def main():
...
+    # Note: We want to apply a more general solution that applies to every region,
+    # but for the time being this is enough to support US isolated regions without
+    # impacting the other ones.
+    ca_bundle = f"/etc/pki/{region}/certs/ca-bundle.pem" if region.startswith("us-iso") else None
+
+    # Connect to AWS using boto
+    ec2 = boto3.client("ec2", region_name=region, config=proxy_config, verify=ca_bundle)
+
+    # Print the volume id of the device volume
+    get_device_volume_id(ec2, dev, instance_id)
```
enrico-usai: I'd say we should have:
```python
volume_id = get_device_volume_id(ec2, dev, instance_id)
print(volume_id)
```

because the name of the function is `get_device...` so the scope is to retrieve it, the fact we're printing it it's instead part of the scope of the `main` function.

test/unit/ec2_udev_rules/test_ec2_dev_2_volid.py
```diff
@@ -31,3 +31,116 @@ def test_validate_device_name(device_name, raises):
...
+    ("dev", "expected_name", "raises"),
+    [
+        ("nvme0n1", "sdf", True),
+        ("nvme0n1p1", "sdf", True),
+        ("nvme0n1p128", "sdf", True),
+        ("xvd0n1", "/dev/sd0n1", False),
+        ("xvd0n1p1", "/dev/sd0n1p1", False),
+    ],
+)
+def test_convert_dev(mocker, dev, expected_name, raises, capsys):
```
enrico-usai: As you can see here we're duplicating the test code, because there is a code duplication. 
We have the same function defined in two different python modules.

We should instead:
1. rename manageVolume.py to manage_volume.py to have the same convention of ec2_dev_2_volid.py 
2. have a common module called for example ec2_volume.py with all the functions that are used by the two modules and import them to avoid duplication.

Anyway we can do it in a future patch.

enrico-usai (commented):
In general it's useful to split this kind of changes in multiple small commits, e.g.:
  * create parse_proxy_config function
  * create get_device_volume_id
  * fix typo in manageVolume function
  * add unit tests for xxx function
  
 This simplifies the review.
 
 Instead you should fixup the second commit you created, this doesn't have any meaning by itself, it's a fix of the previous commit so they should be merged together. See https://www.jetbrains.com/idea/guide/tips/interactive-rebase/
 
Anyway you did a great job, you can see [here](https://app.codecov.io/gh/aws/aws-parallelcluster-cookbook/pull/2445?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=aws) how your patch is increasing the coverage of about 6%.

test/unit/ec2_udev_rules/test_ec2_dev_2_volid.py
judysng: Makes sense, I went ahead and did those in separate commits.

In the most recent commit, I changed manageVolume.py to manage_volume.py.

In the commit before that, I added the ec2_volume.py file and took out the duplicate functions from the other two modules and put it in there (as well as addressed your other comments). One question: for adapt_device_name (previously convert_dev), the two modules actually have slight differences for that function. I left them as-is in their own files, but are they supposed to be different?

cookbooks/aws-parallelcluster-environment/files/default/ec2_udev_rules/ec2_volume.py
```diff
@@ -0,0 +1,56 @@
+import configparser
+import re
+
+import requests
+from botocore.config import Config
+
+METADATA_REQUEST_TIMEOUT = 60
+
+
+def get_imdsv2_token():
```
enrico-usai: Adding this file in the folder is not enough.

This will give you the possibility to experiment with the cookbook. 
If you search for `manageVolume.py` in the cookbook code you can see:

```
  cookbook_file 'manageVolume.py' do
    source 'ec2_udev_rules/manageVolume.py'
    cookbook 'aws-parallelcluster-environment'
    path '/usr/local/sbin/manageVolume.py'
    user 'root'
    group 'root'
    mode '0755'
  end
```

You should do something similar. 

You can start checking:
* `resources/ec2_udev_rules/partial/_common_udev_configuration.rb` --> where we are creating the file in the system
* `spec/unit/resources/ec2_udev_rules_spec.rb` --> unit test you can run with `cd cookbooks/aws-parallelcluster-environment; chef exec rspec ./spec/unit/resources/os_type_spec.rb`
* `test/controls/ec2_udev_rules_spec.rb` --> kitchen tests you can run with `bash kitchen.ec2.sh environment-install test ec2-udev-rules -c 6` --> see README first.

enrico-usai: You could evaluate to do these changes as part of this patch or avoid to create the new ec2_volume.py for now and just write unit tests even if the code is duplicate. Then in another patch you can work to have this code in common and experiment with cookbook changes to add the new file.

judysng: I see, I'll just have the duplicate code for now and work on the patch in the future then.

himani2411 (approved):

PR #2451, opened 2023-09-11
[develop] Add custom munge key config logic

hehe7318:
### Description of changes
* Create a munge key manager
* Add related kitchen tests

### Relate Pull Request
* See the PR of Setup custom munge key in cli: [[develop] Cli Setup custom munge key](https://github.com/aws/aws-parallelcluster/pull/5681)
* [[develop] Add custom munge key update logic](https://github.com/aws/aws-parallelcluster-cookbook/pull/2452)
* [[develop] Add custom munge key rotation script](https://github.com/aws/aws-parallelcluster-cookbook/pull/2453)

### New sample YAML configuration

```
DevSettings:
  SlurmSettings:
    MungeKeySecretArn: Str
```

PR #2452, opened 2023-09-11
[develop] Add custom munge key update logic

hehe7318:
### Description of changes
* Create a munge key update manager
* Add logic to update the custom munge key -> Adding new, changing, removing the munge key secret Arn in YAML file

### Relate Pull Request
* See the PR of Setup custom munge key in cli: [[develop] Cli Setup custom munge key](https://github.com/aws/aws-parallelcluster/pull/5681)
* [[develop] Add custom munge key config logic](https://github.com/aws/aws-parallelcluster-cookbook/pull/2451)
* [[develop] Add custom munge key rotation script](https://github.com/aws/aws-parallelcluster-cookbook/pull/2453)

### New sample YAML configuration

```
DevSettings:
  SlurmSettings:
    MungeKeySecretArn: Str
```


cookbooks/aws-parallelcluster-slurm/libraries/helpers.rb
```diff
@@ -85,6 +94,17 @@ def setup_munge_head_node
   share_munge_head_node
 end
 
+def update_munge_head_node
+  munge_key_update_manager 'update_munge_key' do
```
EddyMM: Instead of having a different manager for both set and update actions, we can have a single munge key resource with different actions (e.g. `setup_key`, `update_key` etc) then invoke the appropriate action depending on what we're doing in the recipe. In this case for example we'd invoke the update action of the same munge key resource.

hehe7318: Makes sense, I will do it when the config PR merged.

hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
```diff
@@ -71,3 +71,11 @@ def execute_command(command, user = "root", timeout = 300, raise_on_error = true
   raise_command_error(command, cmd) if raise_on_error && cmd.error?
   cmd.stdout.strip
 end
+
+# Verify if MungeKeySecretArn in SlurmSettings section of cluster configuration has been updated
+def is_custom_munge_key_updated?
+  require 'yaml'
```
EddyMM: These lines seem to be used multiple times in this update file.

```
  require 'yaml'
  config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
  previous_config = YAML.safe_load(File.read(node['cluster']['previous_cluster_config_path']))
```

Let's take this opportunity to add a reusable approach e.g.
```

def config_parameter_changed(param)?
  # Compares previous cluster config with the current one for changes in a parameter
  # Parameters:
  # - `param`: An array representing the sequence of nested keys to the parameter to be checked 
  require 'yaml'

  config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
  previous_config = YAML.safe_load(File.read(node['cluster']['previous_cluster_config_path']))

  return config.dig(*param) != previous_config.dig(*param)
end

def is_slurm_database_updated?
  ...
  return config_parameter_changed(["Scheduling", "SlurmSettings", "Database"])
end

def is_custom_munge_key_updated?
  ...
  return config_parameter_changed(["DevSettings", "SlurmSettings", "MungeKeySecretArn"])
end

...

```

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -70,3 +70,58 @@
...
+    bash 'generate_munge_key' do
+      user node['cluster']['munge']['user']
+      group node['cluster']['munge']['group']
+      cwd '/tmp'
+      code <<-GENERATE_KEY
+        set -e
+        /usr/sbin/mungekey --verbose
+        chmod 0600 /etc/munge/munge.key
+      GENERATE_KEY
+    end
```
EddyMM: This block of code is used during the `setup` action as well.
Can we move it to a shared function?

EddyMM: Same case for the block of code to fetch and decode the munge key.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -70,3 +70,58 @@
     end
   end
 end
+
+action :update_munge_key do
+  bash 'remove_current_munge_key' do
```
EddyMM: Is there a reason why we want to remove the current munge key before trying to generate or fetch a new one?
I'm thinking about a case where the generation or fetch fails, we may not want to have an empty/non-existent `/etc/munge/munge.key` file. It's preferable to override the existing munge key after a new key generation or fetch operation.

EddyMM:
Is it possible to maintain a temporary copy of the existing munge key and use it to restore the system back to a consistent state in case an error occurs after replacing the key (e.g. munge service fails to restart)? Or maybe fallback to a key generated by Munge? Let's sync on this.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Have been safely removed.

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
hehe7318: Makes sense. Done!

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Done. After removing the checking if the munge key exists section, the update section and the setup section looks the same. But I think we can keep them for now. Because later we may add new logic to update section -> Maintaining a backup copy of the existing Munge key to reuse in case of specific failures

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
```diff
@@ -53,11 +53,18 @@ def is_compute_node_bootstrap_timeout_updated?(previous_config, config)
   evaluate_compute_bootstrap_timeout(previous_config) != evaluate_compute_bootstrap_timeout(config)
 end
 
-def is_slurm_database_updated?
+def config_parameter_changed?(param)
```
EddyMM, line 56: Optional: It can be used for other functions in the file e.g. `are_bulk_custom_slurm_settings_updated`

EddyMM (approved):

PR #2464, opened 2023-09-19
[develop] Fix the creation failure when use custom munge key

hehe7318:
### Description of changes
* Fix the creation failure when use Custom Munge Key

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
+++ b/cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
@@ -57,12 +57,13 @@ end
 
 def generate_munge_key
   declare_resource(:bash, 'generate_munge_key') do
-    user node['cluster']['munge']['user']
-    group node['cluster']['munge']['group']
+    user 'root'
+    group 'root'
     cwd '/tmp'
     code <<-GENERATE_KEY
       set -e
       /usr/sbin/mungekey --verbose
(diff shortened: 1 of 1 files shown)


cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -57,21 +57,26 @@ def fetch_and_decode_munge_key
...
-    group node['cluster']['munge']['group']
+    user 'root'
+    group 'root'
     cwd '/tmp'
     code <<-GENERATE_KEY
-        set -e
-        /usr/sbin/mungekey --verbose
-        chmod 0600 /etc/munge/munge.key
+      set -e
+      # If the /etc/munge/munge.key already exists, the /usr/sbin/mungekey --verbose command will report an error
```
EddyMM: To avoid removing the munge key explicitly, we can go ahead and run `/usr/sbin/mungekey --force --verbose` instead.
[[See Docs](https://github.com/dun/munge/wiki/Man-8-mungekey)]
>-f, --force
>Force the keyfile to be overwritten if it already exists.

This way the current munge key is retained in case generation of a new one fails.

hehe7318: Thanks Eddy! Done!

EddyMM (approved):

PR #2465, opened 2023-09-19
Migrate internal storage to EFS from NFS exports

dreambeyondorange:
### Description of changes
[Develop] Migrate internal storage to EFS from NFS exports

- Add backup and restore recipes to move data in second stage images to shared filesystems
- Create a new mount_internal_use_fs.rb recipe to mount the internal shared filesystems
- Filter the efs filesystem arrays to mount internal shared fses in init and cx fses in config
- Refactor environment recipes to be clearer in functional description and remove unnecessary recipes

### Tests
* Launched a cluster with 2 static nodes and verified the fses were mounted on the head node and compute nodes
* Head node had all 4 fses
* Compute node had all except shared_login_dir

### Checklist


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Install [Spack](https://spack.io) by default in cluster user's home directory.
 
 **CHANGES**
+- Migrate NFS exports and associated data for intra-cluster shared storage to external AWS EFS filesystems attached to the cluster stack.
```
gmarciani: I'd specify that we are moving the NFS export from the head node root volume. Not only to be more specific on tech standpoint but because this has an impact on cost and on the required size of that volume (moving from EBS root volume to EFS).

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
```diff
@@ -10,15 +10,34 @@
...
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+return if on_docker?
+efs_shared_dir_array = node['cluster']['efs_shared_dirs'].split(',')
+efs_fs_id_array = node['cluster']['efs_fs_ids'].split(',')
+efs_encryption_in_transit_array = node['cluster']['efs_encryption_in_transits'].split(',')
+efs_iam_authorization_array = node['cluster']['efs_iam_authorizations'].split(',')
 
-# Mount EFS directory with efs resource
+cx_shared_dir_array = []
```
gmarciani: Minor: AFAIK CX stands for Customer Experience, so it may be not the right acronym.

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
```diff
@@ -10,15 +10,34 @@
...
+cx_efs_fs_id_array = []
+cx_efs_encryption_array = []
+cx_efs_iam_array = []
+
+# Identify the customer use filesystems and store their data in arrays for the EFS resource
+efs_shared_dir_array.each_with_index do |dir, index|
+  next if node['cluster']['internal_shared_dirs'].include?(dir)
```
gmarciani: Correct me if I'm wrong: `efs_shared_dir_array` contains the customer shared storage already (ie: the storage defined in SharedStorage section in config file). So this filter here is to prevent the customer to mount a storage over an internal storage. Right? If this is the goal, note that this filter should be applied at the CLI level, if it's not already there.

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+return if on_docker?
+
+case node['cluster']['node_type']
+when 'HeadNode'
+  Chef::Log.info("Internal Shared Storage #{node['cluster']['internal_shared_dirs']}")
```
gmarciani: Minor: I'd say something more specific like "Backing up internal...."

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
```diff
@@ -0,0 +1,36 @@
...
+  # For each, backup the data to a temp location
+  node['cluster']['internal_shared_dirs'].each do |dir|
+    bash "Backup #{dir}" do
+      user 'root'
+      group 'root'
+      code <<-EOH
+        mkdir -p /tmp#{dir}
+        rsync -a #{dir}/ /tmp#{dir}
```
gmarciani, line 28: I suggest as a best practice to preserve whatever attribute of the source files, not only regular (-a option) but also extended ones (-X option).

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
```diff
@@ -0,0 +1,36 @@
...
+      EOH
+    end
+  end
+
+when 'ComputeFleet', 'LoginNode'
+  Chef::Log.info("Backup only from the HeadNode")
+else
+  raise "node_type must be HeadNode, ComputeFleet, or LoginNode"
+end
```
gmarciani: I suggest one of the following two options in replacement of this section:
1. omit this section as a whole because we may consider redundant to log the skip of an action
2. have only `when HeadNode { do action } else { Log(Skipping action) }`; the advantage is that if we add other node types we do not need to refactor the statement.

Personally I'm more in favour of option 1 because it removes noise.

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_fs.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+# Copyright:: 2013-2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: Minor: copyright year must be `2023 Amazon.com`.

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
gmarciani, line 28: With rsync we are making a copy so we are doubling the space used by the internal directories within the head node root volume. Is there any risk for this operation to fail due to not enough space in the root volume?

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
dreambeyondorange: I've seen this a few times in tickets and other chats https://t.corp.amazon.com/D90018305/communication

CHANGELOG.md
dreambeyondorange: sounds good!

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
dreambeyondorange: In the CLI, I am adding the internal storage using the add_shared_storage mechanism and filtering the data here

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: This is just a debug message

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: Thanks!

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: I'm actually more in favor of removing the else statement and only having the if.  Do we really need to guard against missing node types?

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: Good point, should we remove the files after the copy?  There is of course a risk of not having enough space.

dreambeyondorange: Although I did not run into any issues while launching clusters with this cookbook

dreambeyondorange: There is no issue on alinux2 which root is only 53% full, but maybe on centos which needed a bump in size for 3.7

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: I should also note that this precedent was sent a couple months ago https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-environment/recipes/config/mount_shared.rb

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
dreambeyondorange: This filter is just to distinguish which fses need to be mounted when, not dealing with overlap (which I agree needs to happen in the cli)

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: I ran into issues with copying when using the -X option (rsync -X operation not supported when trying to setxattr), so retaining just -a for now

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Install [Spack](https://spack.io) by default in cluster user's home directory.
 
 **CHANGES**
+- Migrate NFS exports from the head node root volume and associated data for intra-cluster shared storage to external AWS EFS filesystems attached to the cluster stack.
```
enrico-usai: I'd write something like:

Use Amazon EFS in place of Amazon EBS as Head Node root volume. Use the new EFS for Intel and Slurm shared folders, removing the NFS exports from the Head Node.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
```diff
@@ -440,10 +440,10 @@ suites:
         fsx_dns_names: dns1,dns2
         fsx_mount_names: mount1,mount2
         fsx_volume_junction_paths: value1,value2
-  - name: fs_update_default_values
+  - name: update_fs_mapping_default_values
     run_list:
       - recipe[aws-parallelcluster-tests::setup]
-      - recipe[aws-parallelcluster-environment::fs_update]
+      - recipe[aws-parallelcluster-environment::update_fs_mapping]
```
enrico-usai, line 446: I like the name "update_fs_mapping" for the recipe. In general it's good to do this kind of changes in self-contained commit and then do your changes in a commit on top of the refactoring.

CHANGELOG.md
gmarciani: filesystems v/s filesystem: don't we use just one single file system? 
If it's just one let's specify it singular otherwise it could be scary in terms of costs.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
```diff
@@ -472,42 +472,43 @@ suites:
...
-      dependencies:
-        - recipe:aws-parallelcluster-platform::directories
-        - resource:nfs
-        - recipe:aws-parallelcluster-environment::mock_export_directories
-      cluster:
-        node_type: 'LoginNode'
-        head_node_private_ip: '127.0.0.1'
-        head_node_home_path: '/fake_headnode_home'
-        shared_dir_head: '/fake_headnode_shared'
+# TODO replacement for the mount_internal_use_fs recipe since it uses shared storage
```
gmarciani, line 475: Not clear to me what this TODO is expected for. May you please explain?

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
gmarciani: For sure not a blocking comment, but IMHO an info message is valuable if it reflects exactly what's ongoing. These are the messages that we look at when troubleshooting issues with recipes so I think it's helpful to make them descriptive.

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
gmarciani: Alright!

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
gmarciani: > I'm actually more in favor of removing the else statement and only having the if.

I completely agree. Can we do it?

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
gmarciani: In general, I think we should remove files from the root volume after the successful copy into the remote fs. The only exception for that should be the original /home directory because it could be useful to preserve the access (it contains SSH keys) in case of issues with the remote /home.

The drawback of this approach is the rollback on failure: not possible to restore the files in the original root volume in case of failure on the remote fs.

In conclusion, I suggest to not remove the original files, but only the temporary ones immediately after the successful copy.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
enrico-usai, line 475: Is this something to be removed?

cookbooks/aws-parallelcluster-environment/kitchen.environment-recipes.yml
```diff
@@ -0,0 +1,88 @@
+# Validates config recipes
```
enrico-usai: This file name is out of the standard convention. 

We were trying to split the tests in two files: <workstream>-config.yml and <workstream>-install.yml containing both tests for recipes and resources.

This is probably the reason why you have to redefine cluster attibutes, verifiers etc.

Can we move the controls you defined here, into the <workstream>-config.yml?

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
enrico-usai: I think we can split the `mount_cx_fs` in two recipes: `efs` and `fsx`.  

In this way we'll follow the convetion used for raid and ebs and we'll address Giacomo's suggestion about using `cx` word. We don't need to have a wrapper `mount_*_fs` recipe to call `fsx` and `efs` mount, we can move them in the config.rb

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: Yup, it’s part of the latest version of the pr

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: I’ll do this!

CHANGELOG.md
dreambeyondorange: The cost doesn’t change based on the number of filesystems. It’s all about throughput. It is a filesystem per mount point

cookbooks/aws-parallelcluster-environment/attributes/environment.rb
```diff
@@ -53,5 +53,6 @@
 default['cluster']['shared_dir_compute'] = node['cluster']['shared_dir']
 default['cluster']['shared_dir_head'] = node['cluster']['shared_dir']
 default['cluster']['shared_dir_login'] = node['cluster']['shared_dir_login_nodes']
+default['cluster']['internal_shared_dirs'] = [node['cluster']['shared_dir'], node['cluster']['shared_dir_login_nodes'], "/opt/slurm", "/opt/intel"]
```
enrico-usai: What will happen if we change the definition of `/opt/slurm` that is done in the slurm cookbook?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_login.rb
```diff
@@ -17,5 +17,3 @@
 
 # TODO: rename, find a better name that include login nodes
 setup_munge_compute_node
-
-include_recipe 'aws-parallelcluster-slurm::mount_slurm_dir'
```
enrico-usai, line 21: Where are we moving the "mount_slurm_dir"? step? Or should we delete the `mount_slurm_dir` if these steps are performed by any other recipe?

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
```diff
@@ -0,0 +1,29 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+return if on_docker?
+
+if node['cluster']['node_type'] == 'HeadNode'
+  # For each, backup the data to a temp location
```
enrico-usai, line 18: could you please add a comment in the code describing "why" we have to do these steps?

cookbooks/aws-parallelcluster-environment/recipes/config/mount_cx_fs.rb
enrico-usai: I think we can split the mount_cx_fs in two recipes: `efs` and `fsx`.

In this way we'll follow the convention used for raid and ebs and we'll address Giacomo's suggestion about using cx word. We don't need to have a wrapper `mount_*_fs` recipe to call fsx and efs mount, we can move them in the config.rb

dreambeyondorange: I use CX as many others do to refer to customers.  There are many instances where folks are using this.  I am using this to distinguish cx shared drives from the internal use shared drives.  Given that users can add both EFS and FSx filesystems, I'm not sure I see the value in separating them now, when they weren't already separated before.  The work done at the beginning of this file collects only the fses that need to be mounted during config, and not to remount the ones that were already mounted in the init.  This will also greatly complicate the config recipe with extra logic related to efs.  I think I can go with separating fsx and efs.  Calling FSX directly from the config and using a separate recipe for efs which encapsulates the extra logic outside of the resource.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
dreambeyondorange: This is to replace the tests below that were commented out.  Since these tests were performed by mocking storage, I don't quite know how to do this yet with actual external storage or if that is even possible.  So, I'm using this as reminder to make sure this is tested in the next task

dreambeyondorange: This is something to be addressed in updating kitchen and integ tests for this refactoring

cookbooks/aws-parallelcluster-environment/kitchen.environment-recipes.yml
dreambeyondorange: I was trying to replicate what is in the top level validate-config to see similar behavior to what runs in our system tests since there are some differences there from what is in the other -install and -config tests.  But I realized after creating this that the tests won't work anyway since we need file systems to attach.  I will remove this for now

cookbooks/aws-parallelcluster-environment/attributes/environment.rb
dreambeyondorange: It will break.  I can define the slurm install dir in the environment attributes instead!

cookbooks/aws-parallelcluster-slurm/recipes/config/config_login.rb
dreambeyondorange: this is done in the mount_internal_use_fs recipe, that is why I was removing it from here.  I'll also remove the recipe!

cookbooks/aws-parallelcluster-environment/recipes/init/backup_internal_use_shared_data.rb
dreambeyondorange: sure

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
dreambeyondorange: This is sometimes hard to do without knowing everything that needs to change in advance. Trying to separate changes within the file now in git is pretty cumbersome.  I definitely agree, especially with large refactoring efforts and will keep this in mind for the future!

CHANGELOG.md
dreambeyondorange: I'll say 
`Remove the NFS exports from the head node root volume and use Amazon EFS storage instead for intra-cluster shared ParallelCluster, Intel, Slurm, and Login Node data. `

gmarciani (approved):

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_fs.rb
```diff
@@ -0,0 +1,53 @@
...
+# Identify the internal use filesystems and store their data in arrays for the EFS resource
+efs_shared_dir_array.each_with_index do |dir, index|
+  next unless node['cluster']['internal_shared_dirs'].include?(dir)
+  # Don't mount the login nodes shared dir to compute nodes
+  next if node['cluster']['node_type'] == 'ComputeFleet' && dir == node['cluster']['shared_dir_login_nodes']
+  internal_shared_dir_array.push(dir)
+  internal_efs_fs_id_array.push(efs_fs_id_array[index])
+  internal_efs_encryption_array.push(efs_encryption_in_transit_array[index])
+  internal_efs_iam_array.push(efs_iam_authorization_array[index])
+end
```
enrico-usai, line 33: This block of code, from the beginning, it's "duplicated" in the efs.rb.

What about having a first recipe that parses the `efs_shared_dirs` creating the two `internal_*_array` vs `cx_*_array` ?

Then the `mount_internal_use_fs.rb` will just call the mount of the efs passing `internal_*_array` and the `efs` will mount the `cx_*` ones.

In the code it's enough to have it at the beginning of the `init.rb` before the `mount`, In the inspec tests we just need to add it as dependency to be executed as first.

What do you think?

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_fs.rb
```diff
@@ -0,0 +1,53 @@
...
+efs "mount internal use efs" do
+  shared_dir_array internal_shared_dir_array
+  efs_fs_id_array internal_efs_fs_id_array
+  efs_encryption_in_transit_array internal_efs_encryption_array
+  efs_iam_authorization_array internal_efs_iam_array
+  action :mount
+  not_if { internal_shared_dir_array.empty? }
+end
+
+# TODO: replace home as NFS with shared /home
```
enrico-usai, line 45: Is this the next PR?

cookbooks/aws-parallelcluster-environment/recipes/init/restore_internal_use_shared_data.rb
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+
+return if on_docker?
+
+if node['cluster']['node_type'] == 'HeadNode'
+  # For each, restore the shared storage if it doesn't already exist
+  # This is necessary to preserve any data in these directories that was
+  # generated during the build of ParallelCluster AMIs after converting to
+  # shared storage and backed up to a temporary location previously
+  # Remove the backup after the copy is done
```
enrico-usai, line 22: `# Remove the backup after the copy is done` can be moved in the code block before the `rm`.

cookbooks/aws-parallelcluster-slurm/test/controls/mount_slurm_dir_spec.rb
```diff
@@ -1,22 +0,0 @@
...
-# Licensed under the Apache License, Version 2.0 (the "License").
-# You may not use this file except in compliance with the License. A copy of the License is located at
-#
-# http://aws.amazon.com/apache2.0/
-#
-# or in the "LICENSE.txt" file accompanying this file.
-# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
-# See the License for the specific language governing permissions and limitations under the License.
-
-control 'mount_slurm_dir' do
```
enrico-usai, line 12: This test should be moved to `controls/mount_internal_use_fs_spec.rb`

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_fs.rb
dreambeyondorange: yes it is

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_fs.rb
dreambeyondorange: I'm not sure.  I prefer having code here that just filters the ones we need for the init and a separate one that filters what we need for the cx.  We do this filtering process twice, so I wonder if it might be better to just put this into the attributes file and then leverage the attributes we need in each recipe.  This did get auto merged with Giacomo's approval, but I will do this refactoring as a follow up

PR #2470, opened 2023-09-25
Add support for Database/Database name to render StorageLoc in slurmdbd configuration

jdeamicis:
### Description of changes
* Add support for `Scheduling/SlurmSettings/Database/DatabaseName` to be used to configure `StorageLoc` in the slurmdbd configuration include file generated by ParallelCluster.
* Add some unit tests (InSpec and ChefSpec) for the `config_slurm_accounting.rb` recipe.

### Tests
* Added case in unit test for the rendering of the Slurm configuration include files relevant for Slurm Accounting.
* Manually ran the InSpec and ChefSpec added for the `config_slurm_accounting.rb` recipe.

### References
* https://github.com/aws/aws-parallelcluster/pull/5718

### Checklist


cookbooks/aws-parallelcluster-slurm/kitchen.slurm-config.yml
```diff
@@ -191,3 +191,29 @@ suites:
...
+        - recipe:aws-parallelcluster-platform::users
+        - recipe:aws-parallelcluster-slurm::slurm_users
+        - recipe:aws-parallelcluster-slurm::mock_slurm
+        - recipe:aws-parallelcluster-slurm::mock_scripts_dir
+      cluster:
+        node_type: HeadNode
+        scheduler: 'slurm'
+        #slurm:
+        #  user: 'slurm'
+        #  group: 'slurm'
```
EddyMM: These were used for testing? I think they can be removed.

cookbooks/aws-parallelcluster-slurm/recipes/install/slurm_users.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-slurm
+# Recipe:: install_slurm
+#
+# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
EddyMM: NIT: `...-2023`

test/unit/slurm/test_slurm_config_generator.py
```diff
@@ -102,14 +102,17 @@ def test_generate_slurm_config_files_memory_scheduling(
...
 
 @pytest.mark.parametrize(
-    "slurm_accounting",
-    [(False), (True)],
+    "slurm_accounting, dbname_passed",
+    [(False, False), (True, False), (True, True)],
 )
-def test_generate_slurm_config_files_slurm_accounting(mocker, test_datadir, tmpdir, slurm_accounting):
+def test_generate_slurm_config_files_slurm_accounting(mocker, test_datadir, tmpdir, slurm_accounting, dbname_passed):
+    input_base_name = "sample_input"
```
EddyMM: Instead of building the names programmatically (which requires inspecting various parts of this function), can we pass the files names via Parametrize
`("input_file, output_file", [...])`.

In this case for example, we had to introduce a new if condition in order to test a new parameter.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_slurmdbd.conf
```diff
@@ -1,12 +1,17 @@
...
-{% if scaling_config.Database.Uri is defined %}
-DbdHost={{ head_node_config.head_node_hostname }}
-StorageHost={{ scaling_config.Database.Uri | uri_host }}
-StoragePort={{ scaling_config.Database.Uri | uri_port }}
-{# Dashes in StorageLoc cause issues with the database creation #}
-StorageLoc={{ cluster_name | replace("-", "_") }}
-StorageUser={{ scaling_config.Database.UserName }}
-StoragePass=dummy
-{% endif %}
+{# Adding comments at the beginning of each line is a trick to indent the template without affecting the output #}
```
EddyMM, line 1: I think we generate the configuration files using the `trim_blocks=True` and `lstrip_blocks=True` which should let us use indentation in the template without it appearing in the generated file.

Do the indentations appear in the generated file?

cookbooks/aws-parallelcluster-slurm/kitchen.slurm-config.yml
jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/recipes/install/slurm_users.rb
jdeamicis: Addressed.

test/unit/slurm/test_slurm_config_generator.py
jdeamicis: Addressed.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/templates/slurm_parallelcluster_slurmdbd.conf
jdeamicis: I had written a nice comment that got lost :/

Long story short, those options only remove the new lines and spaces caused by the presence of the `{# ... #}` Jinja blocks, not any indentation we may want to apply to the rest of the template to make the logic more understandable. Meaning that yes, if I indent lines that do not have Jinja blocks in them, they will remain indented in the rendered output.

I couldn't find a nicer alternative to this.

My impression is that this trick makes the logic embedded in the template more easy to follow, although it becomes slightly uglier to read. I would be in favor of keeping this trick.

EddyMM (approved):

PR #2471, opened 2023-09-25
[develop] Reuse the rotation script inside the munge resource

hehe7318:
### Description of changes
* Avoid code duplication by reusing the "rotation" script inside the munge resource when retrieving the custom munge key from Secrets Manager

### Checklist


cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -23,35 +23,16 @@
...
-
-      # Set ownership on the key
-      chown #{node['cluster']['munge']['user']}:#{node['cluster']['munge']['group']} /etc/munge/munge.key
-      # Enforce correct permission on the key
-      chmod 0600 /etc/munge/munge.key
-    FETCH_AND_DECODE
+    cwd File.dirname(script_path)
+    code <<-SHELL
+      ./#{File.basename(script_path)}
+    SHELL
```
jdeamicis: Please declare this as a resource of type `:execute`, and then use the same syntax as the `execute` resource.

jdeamicis (changes requested):
Good job. Just implement the change I requested and then we can merge it.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -15,12 +15,31 @@ MUNGE_USER="<%= @munge_user %>"
...
+        exit 1
+      fi
+      CONFIG_PROCESS=$OPTARG
+      ;;
+    *)
+      echo "Usage: $0 [-c True/False]" >&2
+      exit 1
+      ;;
+  esac
+done
```
jdeamicis: You can get the same behavior with a simple flag, without having to pass either `True` or `False` to the argument:

```bash
CONFIG_PROCESS=false
while getopts "c" opt; do
  case $opt in
    c) CONFIG_PROCESS=true;;
    ?) # echo usage output and exit
  esac
done
```

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -59,33 +78,35 @@ else
   exit 1
 fi
 
-# Enable and restart munge service
-systemctl enable munge
-echo "Restarting munge service"
-systemctl restart munge
+if [ "$CONFIG_PROCESS" == "False" ]; then
```
jdeamicis: Yesterday we discussed about this and we just said:
- if the script is called without any option, it will check the status of compute (and login) fleet (that's how users should use it directly on the head node).
- if the script is passed with the boolean flag, this flag disables the check of the resources (this is what we are going to do in our recipes, since in that case we are already covered by:
  - the fact that such resources do not exist yet at cluster creation time;
  - the update policies at cluster update time.
  
I'm sorry if there was any misunderstanding about this.  

The way this is implemented now, the script is just going to be completely disabled by the boolean flag (and we don't want that).

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -15,12 +15,31 @@ MUNGE_USER="<%= @munge_user %>"
 MUNGE_GROUP="<%= @munge_group %>"
 SHARED_DIRECTORY_COMPUTE="<%= @shared_directory_compute %>"
 SHARED_DIRECTORY_LOGIN="<%= @shared_directory_login %>"
+CONFIG_PROCESS="False"
```
jdeamicis: I would propose a variable called `DISABLE_RESOURCE_CHECK`, which is false by default, and in case the boolean flag is used, it is set to true.

If this variable is true, then the part of the check on the compute (and login) fleet status is skipped.

If we go for `DISABLE_RESOURCE_CHECK`, then I would use the option `d` in the `getopts` below.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -15,12 +15,31 @@ MUNGE_USER="<%= @munge_user %>"
...
-if ! echo "$compute_fleet_status" | grep -q '"status": "STOPPED"'; then
-    echo "Compute fleet is not stopped. Please stop it before updating the munge key."
-    exit 1
+if [ "$CONFIG_PROCESS" == "False" ]; then
+  # Check compute fleet status
+  compute_fleet_status=$(get-compute-fleet-status.sh)
+  if ! echo "$compute_fleet_status" | grep -q '"status": "STOPPED"'; then
+      echo "Compute fleet is not stopped. Please stop it before updating the munge key."
+      exit 1
+  fi
```
jdeamicis, line 37: Where is the check on the login nodes status? Is it in a different PR (that depends on this one)?

jdeamicis (changes requested):
This is not what we had discussed yesterday. Please address the comments.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
jdeamicis: It's in [this other PR](https://github.com/aws/aws-parallelcluster-cookbook/pull/2473/files#diff-449b77b3cae720dd4ab62600d04ca4262e20adb4e072ffb30de6de2580bc3bd2): please make sure we do not miss this when merging the various pull requests.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Thanks for reminding, Jacopo. Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Yes. I won't miss it.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Done as we discussed today.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -15,6 +15,29 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Copy pcluster config generator and templates
+remote_directory "#{node['cluster']['scripts_dir']}/slurm" do
+  source 'head_node_slurm/slurm'
+  mode '0755'
+  action :create
+  recursive true
+end
```
EddyMM: Why did we move this block to the top of the recipe?

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -23,35 +23,43 @@
...
+  end
+end
 
-      # Decode munge key and write to /etc/munge/munge.key
-      decoded_key=$(echo $encoded_key | base64 -d)
-      if [ $? -ne 0 ]; then
-        echo "Error decoding the munge key with base64"
-        exit 1
-      fi
+def share_munge_head_node
```
EddyMM: NIT: We can call this `share_munge` since we share with the login nodes too if they exist.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -23,35 +23,43 @@
...
+  share_munge_key_to_dir(node['cluster']['shared_dir'])
+  share_munge_key_to_dir(node['cluster']['shared_dir_login'])
```
EddyMM: Let's check the use of `node['cluster']['shared_dir_login_nodes']` instead of `node['cluster']['shared_dir_login']` here.

`node['cluster']['shared_dir_login']` is set in the environment cookbook but we should probably rely on variables set in the `aws-parallelcluster-shared` which all cookbooks depend on.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -23,35 +23,43 @@
...
-      echo "$decoded_key" > /etc/munge/munge.key
+def fetch_and_decode_munge_key
+  script_path = "#{node['cluster']['scripts_dir']}/slurm/update_munge_key.sh"
 
-      # Set ownership on the key
-      chown #{node['cluster']['munge']['user']}:#{node['cluster']['munge']['group']} /etc/munge/munge.key
-      # Enforce correct permission on the key
-      chmod 0600 /etc/munge/munge.key
-    FETCH_AND_DECODE
+  declare_resource(:execute, 'fetch_and_decode_munge_key') do
```
EddyMM, line 57: Why are we declaring a custom resource to execute a script instead of using in-built resources like `script` or `bash`?

hehe7318: Previously I use `execute` and `bash do`. But raised an error. After change to `declare_resource`, things go well. I remember it's because of def function in resource.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hehe7318: Because if we want to reuse the rotation script in the config process, the script needs to be created before the config process. So we move the create directory function and the create script function to the top.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Done! Also changed in other places.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
jdeamicis: It looks like you need to sort this out, because recently this was moved to a new recipe called `'aws-parallelcluster-slurm::config_head_node_directories'`

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -23,35 +23,43 @@
...
+    user 'root'
+    group 'root'
+    cwd ::File.dirname(script_path)
+    command "./#{::File.basename(script_path)} -d"
   end
 end
```
jdeamicis, line 62: I believe you can simplify this block:
```ruby
def fetch_and_decode_munge_key
  declare_resource(:execute, 'fetch_and_decode_munge_key') do
    user 'root'
    group 'root'
    command "/#{node['cluster']['scripts_dir']}/slurm/update_munge_key.sh -d"
  end
end
```

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
```diff
@@ -23,35 +23,43 @@
...
```
jdeamicis, line 63: Please put a comment here explaining why we do `restart_munge_service` and `share_munge` from within `generate_munge_key`.

Also, add a todo about renaming `generate_munge_key` and `fetch_and_decode_munge_key` into two more speaking names.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -15,12 +15,25 @@ MUNGE_USER="<%= @munge_user %>"
...
+      ;;
+  esac
+done
 
-# Check compute fleet status
-compute_fleet_status=$(get-compute-fleet-status.sh)
-if ! echo "$compute_fleet_status" | grep -q '"status": "STOPPED"'; then
-    echo "Compute fleet is not stopped. Please stop it before updating the munge key."
-    exit 1
+if ! $DISABLE_RESOURCE_CHECK; then
```
jdeamicis, line 31: This is implicitly using the [`true`](https://www.gnu.org/software/coreutils/manual/html_node/true-invocation.html#true-invocation) and [`false`](https://www.gnu.org/software/coreutils/manual/html_node/false-invocation.html#false-invocation). Are we sure this is solid enough?

jdeamicis (commented):
You have to rebase your PR on top of other developments, so I will just provide a comment here.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Great comment! Thank you Jacopo.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Since in today's discussion we don't have a better solution for now. And this PR needs to be merged ASAP. Let's leave it here and if we decide to modify it, we can open a new PR for it.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
jdeamicis: OK, it was done.

cookbooks/aws-parallelcluster-slurm/resources/munge_key_manager.rb
jdeamicis: I wonder if we actually need the first `/` character in the `command` variable. Though, in principle `scripts_dir` will always be an absolute path, so it *shouldn't* hurt.

jdeamicis (approved):

PR #2472, opened 2023-09-25
[develop] check dcv sessions using uid to avoid username truncation

timfurlong:
### Description of changes
* The `_is_session_valid` function of the DCV authenticator would not work for long usernames. This is because the `ps  aux` command truncates lengthy usernames. For example, the processes started by the `dcvextauth` user display as `dcvexta+` in the `ps aux` output
* This change fixes the above by querying by UID rather than username

### Tests
* `test/unit/dcv/test_dcv_authenticator.py` mocked out the changed functions with `mock_verify_session_existence`, and I couldn't think of an easy way to replicate the behavior without actually starting a process using a long username. Suggestions for a better approach to testing this are more than welcome!
* Since automated testing seemed impractical in this case, I went with a manual approach. I started a dcv session using `dcv create-session` using a user with a long username (14 characters long). I then restarted the dcv authenticator process, and queried it using:
```
curl --retry 3 --max-time 5 -s -k -X GET -G https://localhost:8444 -d action=requestToken \
        -d authUser=$longusername -d sessionID=$sessionID
```
Prior to this change, this curl command would result in a timeout.

### Checklist


cookbooks/aws-parallelcluster-platform/files/dcv/pcluster_dcv_authenticator.py
```diff
@@ -361,13 +361,17 @@ def _is_session_valid(user, session_id):
         # TODO change this method if DCV updates his behaviour.
         """
         logger.info("Verifying NICE DCV session validity..")
+
+        # Query by uid rather than username to avoid truncation by ps command
+        uid = subprocess.check_output(["id", "-u", user]).decode("utf-8").strip()
```
enrico-usai: here you should use: `/usr/bin/id` as first argument, see https://bandit.readthedocs.io/en/1.7.5/plugins/b607_start_process_with_partial_path.html

enrico-usai: Could you also test by passing `shell=True` to `check_output`?
see https://bandit.readthedocs.io/en/1.7.5/plugins/b603_subprocess_without_shell_equals_true.html

enrico-usai:
Hi @timfurlong thanks for creating this patch!
I think we never faced this issue because DCV is enabled only for default users that have short names (ec2-user, centos, ubuntu).

I left a comment.
Thanks!

timfurlong:
> Hi @timfurlong thanks for creating this patch! I think we never faced this issue because DCV is enabled only for default users that have short names (ec2-user, centos, ubuntu).
> 
> I left a comment. Thanks!

Hi @enrico-usai thanks for the feedback! I made a couple of small changes in response to your comment. Please let me know if you'd like to see any other changes.

timfurlong:
@enrico-usai Just bumping this PR. Can you please approve/merge?

test/unit/dcv/test_dcv_authenticator.py
```diff
@@ -163,6 +163,26 @@ def test_get_request_token_parameter(parameters, keys, result):
             DCVAuthenticator._extract_parameters_values(parameters, keys)
 
 
+def test_is_session_valid(mocker):
```
enrico-usai, line 166: An alternative to define this test and avoid code duplication (e.g. mocking part) is using `@pytest.mark.parametrize(`, passing `sessionid, expected_error` as parameters.
You can see `test_get_request_token_parameter` defined above as an example.

PR #2473, opened 2023-09-26
[develop] Add a script of checking if there's running LoginNodes inside HeadNode used in munge key rotation script

hehe7318:
### Description of changes
* Add a script of checking if there's runing LoginNodes inside HeadNode using the same way as CLI does

### References
* [What CLI does](https://github.com/aws/aws-parallelcluster/blob/08a76e09ed3c6495aec6e1e7d42020c1232f853b/cli/src/pcluster/models/cluster.py#L749)
* This PR is based on [[develop] Add IAM policies for checking LoginNodes status script inside HeadNode](https://github.com/aws/aws-parallelcluster/pull/5725)

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/libraries/update.rb
+++ b/cookbooks/aws-parallelcluster-slurm/libraries/update.rb
@@ -83,3 +83,7 @@ end
 def is_custom_munge_key_updated?
   config_parameter_changed?(%w(DevSettings MungeKeySettings MungeKeySecretArn))
 end
+
+def is_login_nodes_pool_name_updated?
+  config_parameter_changed?(%w(LoginNodes Pools 0 Name))
+end
--- a/cookbooks/aws-parallelcluster-slurm/libraries/update.rb
+++ b/cookbooks/aws-parallelcluster-slurm/libraries/update.rb
@@ -11,6 +11,7 @@
 # or in the "LICENSE.txt" file accompanying this file.
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+# rubocop:disable Style/SingleArgumentDig
 
 # Helpers functions used by update recipe steps.
 require 'chef/mixin/shell_out'
@@ -86,4 +87,11 @@ end
 
 def is_login_nodes_pool_name_updated?
   config_parameter_changed?(['LoginNodes', 'Pools', 0, 'Name'])
+end
+
+def is_login_nodes_removed?
+  require 'yaml'
+  config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
+  previous_config = YAML.safe_load(File.read(node['cluster']['previous_cluster_config_path']))
+  previous_config.dig("LoginNodes") and !config.dig("LoginNodes")
 end
--- a/cookbooks/aws-parallelcluster-slurm/libraries/update.rb
+++ b/cookbooks/aws-parallelcluster-slurm/libraries/update.rb
@@ -85,5 +85,5 @@ def is_custom_munge_key_updated?
 end
 
 def is_login_nodes_pool_name_updated?
-  config_parameter_changed?(%w(LoginNodes Pools 0 Name))
+  config_parameter_changed?(['LoginNodes', 'Pools', 0, 'Name'])
 end
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
@@ -227,11 +227,10 @@ template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_status.sh" d
   variables(
     stack_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
     login_nodes_pool_name: lazy do
-      if lazy { node['cluster']['config']['LoginNodes'].nil? }
-        nil
-      else
-        lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) }
-      end
+      # rubocop:disable Style/SingleArgumentDig
+      login_nodes_config = node['cluster']['config'].dig(:LoginNodes)
+      # rubocop:enable Style/SingleArgumentDig
+      login_nodes_config ? login_nodes_config.dig(:Pools, 0, :Name) : nil
     end,
     region: node['cluster']['region']
   )
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
@@ -219,6 +219,24 @@ execute "check slurmctld status" do
   retry_delay 2
 end unless redhat_on_docker?
 
+template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_status.sh" do
+  source 'slurm/head_node/check_login_nodes_status.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0755'
+  variables(
+    stack_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
+    login_nodes_pool_name: lazy do
+      if lazy { node['cluster']['config']['LoginNodes'].nil? }
+        nil
+      else
+        lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) }
+      end
+    end,
+    region: node['cluster']['region']
+  )
+end
+
 template "#{node['cluster']['scripts_dir']}/slurm/update_munge_key.sh" do
   source 'slurm/head_node/update_munge_key.sh.erb'
   owner 'root'
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
@@ -14,6 +14,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+# rubocop:disable Style/SingleArgumentDig
 
 include_recipe 'aws-parallelcluster-slurm::config_munge_key'
 
@@ -219,21 +220,19 @@ execute "check slurmctld status" do
   retry_delay 2
 end unless redhat_on_docker?
 
-template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_status.sh" do
-  source 'slurm/head_node/check_login_nodes_status.sh.erb'
+template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_stopped.sh" do
+  source 'slurm/head_node/check_login_nodes_stopped.sh.erb'
   owner 'root'
   group 'root'
-  mode '0755'
+  mode '0700'
   variables(
-    stack_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
-    login_nodes_pool_name: lazy do
-      # rubocop:disable Style/SingleArgumentDig
-      login_nodes_config = node['cluster']['config'].dig(:LoginNodes)
-      # rubocop:enable Style/SingleArgumentDig
-      login_nodes_config ? login_nodes_config.dig(:Pools, 0, :Name) : nil
-    end,
+    cluster_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
+    login_nodes_pool_name: lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) },
     region: node['cluster']['region']
   )
+  only_if do
+    node['cluster']['config'].dig(:LoginNodes)
+  end
 end
 
 template "#{node['cluster']['scripts_dir']}/slurm/update_munge_key.sh" do
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
@@ -14,7 +14,6 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
-# rubocop:disable Style/SingleArgumentDig
 
 include_recipe 'aws-parallelcluster-slurm::config_munge_key'
 
@@ -220,20 +219,7 @@ execute "check slurmctld status" do
   retry_delay 2
 end unless redhat_on_docker?
 
-template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_stopped.sh" do
-  source 'slurm/head_node/check_login_nodes_stopped.sh.erb'
-  owner 'root'
-  group 'root'
-  mode '0700'
-  variables(
-    cluster_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
-    login_nodes_pool_name: lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) },
-    region: node['cluster']['region']
-  )
-  only_if do
-    node['cluster']['config'].dig(:LoginNodes)
-  end
-end
+include_recipe 'aws-parallelcluster-slurm::config_check_login_stopped_script'
 
 template "#{node['cluster']['scripts_dir']}/slurm/update_munge_key.sh" do
   source 'slurm/head_node/update_munge_key.sh.erb'
--- a/cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
@@ -14,7 +14,6 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
-# rubocop:disable Style/SingleArgumentDig
 
 execute 'stop clustermgtd' do
   command "#{cookbook_virtualenv_path}/bin/supervisorctl stop clustermgtd"
@@ -203,21 +202,8 @@ ruby_block "Update Slurm Accounting" do
 end unless on_docker?
 
 # Update check login nodes status script to update pool name
-template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_stopped.sh" do
-  source 'slurm/head_node/check_login_nodes_stopped.sh.erb'
-  owner 'root'
-  group 'root'
-  mode '0700'
-  variables(
-    cluster_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
-    login_nodes_pool_name: lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) },
-    region: node['cluster']['region']
-  )
-  only_if do
-    node['cluster']['config'].dig(:LoginNodes) &&
-      ::File.exist?(node['cluster']['previous_cluster_config_path']) &&
-      is_login_nodes_pool_name_updated?
-  end
+if ::File.exist?(node['cluster']['previous_cluster_config_path']) && is_login_nodes_pool_name_updated?
+  include_recipe 'aws-parallelcluster-slurm::config_check_login_stopped_script'
 end
 
 file "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_stopped.sh" do
--- a/cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
@@ -14,6 +14,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
+# rubocop:disable Style/SingleArgumentDig
 
 execute 'stop clustermgtd' do
   command "#{cookbook_virtualenv_path}/bin/supervisorctl stop clustermgtd"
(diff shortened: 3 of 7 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -282,6 +282,23 @@
   retry_delay 2
 end unless redhat_on_docker?
 
+template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_status.sh" do
+  source 'slurm/head_node/check_login_nodes_status.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0755'
```
jdeamicis: I'd say this script must be 700.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -282,6 +282,23 @@
...
+  variables(
+    stack_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
+    login_nodes_pool_name: lazy do
+      # rubocop:disable Style/SingleArgumentDig
+      login_nodes_config = node['cluster']['config'].dig(:LoginNodes)
+      # rubocop:enable Style/SingleArgumentDig
+      login_nodes_config ? login_nodes_config.dig(:Pools, 0, :Name) : nil
```
jdeamicis: Why don't you put the if condition around the whole `template` statement instead of having this logic inside the `variables` argument?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,75 @@
...
+
+STACK_NAME="<%= @stack_name %>"
+LOGIN_NODES_POOL_NAME="<%= @login_nodes_pool_name %>"
+REGION="<%= @region %>"
+
+# Check if LOGIN_NODES_POOL_NAME is empty
+if [[ -z "${LOGIN_NODES_POOL_NAME}" ]]; then
+    echo "No LoginNodes are specified. Please refer to ParallelCluster official documentation for more information."
+    exit 0
+fi
```
jdeamicis: If you move the logic to check if login nodes are being used outside the variables definition in the `template` above, then you don't need this part of the script.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
jdeamicis: Also, this file should not exist if login nodes are not being used, so we should definitely move the if condition around the creation of the script.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,75 @@
...
+load_balancer_arn=''
+for arn in $(echo "${load_balancers}" | jq -r '.LoadBalancers[].LoadBalancerArn'); do
+    # Get tags for the current Load Balancer
+    tags=$(aws elbv2 describe-tags --resource-arns "${arn}" --region ${REGION})
+
+    # Check if the tags match the desired stack name and login nodes pool name
+    stack_name_match=$(echo "${tags}" | jq -r --arg key "parallelcluster:cluster-name" --arg value "${STACK_NAME}" '.TagDescriptions[] | select(.Tags[]? | (.Key == $key and .Value == $value))')
+    login_nodes_pool_name_match=$(echo "${tags}" | jq -r --arg key "parallelcluster:login-nodes-pool" --arg value "${LOGIN_NODES_POOL_NAME}" '.TagDescriptions[] | select(.Tags[]? | (.Key == $key and .Value == $value))')
+
+    # If both tags are found, store the ARN and break the loop
```
jdeamicis: Please put an additional comment saying that this implies that there's only one pool of login nodes per cluster.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,75 @@
...
+# Check if LOGIN_NODES_POOL_NAME is empty
+if [[ -z "${LOGIN_NODES_POOL_NAME}" ]]; then
+    echo "No LoginNodes are specified. Please refer to ParallelCluster official documentation for more information."
+    exit 0
+fi
+
+# List all Load Balancers
+load_balancers=$(aws elbv2 describe-load-balancers --region ${REGION})
+
+# Iterate over Load Balancers to find the one with matching tags
```
jdeamicis: I believe you should be able to use `aws elbv2 describe-load-balancers --load-balancer-names <name>` and you should already get the load balancer related to the cluster.
 
 The name of the load balancer should follow a precise schema defined in https://github.com/aws/aws-parallelcluster/blob/d8da6bb0bc30ad3c3957da081d6ed39b399bd099/cli/src/pcluster/templates/login_nodes_stack.py#L290

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+
+set -e
+
+STACK_NAME="<%= @stack_name %>"
+LOGIN_NODES_POOL_NAME="<%= @login_nodes_pool_name %>"
+REGION="<%= @region %>"
+
+# Check if LOGIN_NODES_POOL_NAME is empty
+if [[ -z "${LOGIN_NODES_POOL_NAME}" ]]; then
+    echo "No LoginNodes are specified. Please refer to ParallelCluster official documentation for more information."
```
EddyMM: I think logging `No LoginNodes pool name specified.` is enough since the documentation won't resolve an issue where the pool name doesn't exist when executing this script.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+        load_balancer_arn="${arn}"
+        break
+    fi
+done
+
+# Output result
+if [[ -n "${load_balancer_arn}" ]]; then
+    echo "Load Balancer ARN found: ${load_balancer_arn}"
+else
+    echo "No Load Balancer found for the specified stack and login nodes pool."
```
EddyMM: We can include the stack and pool name in the message for debugging purposes.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+set -e
+
+STACK_NAME="<%= @stack_name %>"
+LOGIN_NODES_POOL_NAME="<%= @login_nodes_pool_name %>"
+REGION="<%= @region %>"
+
+# Check if LOGIN_NODES_POOL_NAME is empty
+if [[ -z "${LOGIN_NODES_POOL_NAME}" ]]; then
+    echo "No LoginNodes are specified. Please refer to ParallelCluster official documentation for more information."
+    exit 0
```
EddyMM: Why are we exiting with `0` here but `1` [here](https://github.com/aws/aws-parallelcluster-cookbook/pull/2473/files#diff-144edfeb6e9a7251e4567fb838e57f73bbd3efdfe4f5e4dab837d4b3e427dd9cR47)?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -15,6 +15,21 @@ MUNGE_USER="<%= @munge_user %>"
 MUNGE_GROUP="<%= @munge_group %>"
 SHARED_DIRECTORY_COMPUTE="<%= @shared_directory_compute %>"
 SHARED_DIRECTORY_LOGIN="<%= @shared_directory_login %>"
+CHECK_LOGIN_NODES_SCRIPT_PATH="<%= node['cluster']['scripts_dir'] %>/slurm/check_login_nodes_status.sh"
+
+# Create a temporary file to store the output
+TEMP_FILE=$(mktemp)
```
EddyMM: You can use the `trap` command to ensure that the file is removed once the script exits.
```
trap "rm -f ${TEMP_FILE}" EXIT
```

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+# It first retrieves the ARN of the Load Balancer associated with the specified stack and login nodes pool.
+# If a Load Balancer is found, it then retrieves the ARN of the Target Group associated with the Load Balancer.
+# Lastly, it checks the health of the targets in the Target Group to determine the number of healthy and unhealthy login nodes.
+# If there are any healthy or unhealthy nodes found, it concludes that there are running login nodes.
+#
+# Usage: ./check_if_has_running_login_nodes.sh
+
+set -e
+
+STACK_NAME="<%= @stack_name %>"
```
jdeamicis: Let's call this `CLUSTER_NAME`.

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -201,7 +201,26 @@ def update_nodes_in_queue(strategy, queues)
...
+  group 'root'
+  mode '0755'
+  variables(
+    stack_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
+    login_nodes_pool_name: lazy do
+      # rubocop:disable Style/SingleArgumentDig
+      login_nodes_config = node['cluster']['config'].dig(:LoginNodes)
+      # rubocop:enable Style/SingleArgumentDig
+      login_nodes_config ? login_nodes_config.dig(:Pools, 0, :Name) : nil
+    end,
```
jdeamicis: There should be already a comment about this. Recommenting for this commit (in case the other comment is not immediately visible to you).

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+
+healthy_count=$(echo $target_healths | jq -r '.TargetHealthDescriptions[] | select(.TargetHealth.State == "healthy") | .Target.Id' | wc -l)
+unhealthy_count=$(echo $target_healths | jq -r '.TargetHealthDescriptions[] | select(.TargetHealth.State != "healthy") | .Target.Id' | wc -l)
+
+# Check if there are running login nodes
+total_nodes=$((healthy_count + unhealthy_count))
+if [[ $total_nodes -gt 0 ]]; then
+    echo "Login nodes are running."
+else
+    echo "Login nodes are stopped."
```
jdeamicis: Shouldn't we exit 0 here? Actually, you could end the if condition earlier, remove the else and just exit 0 after echoing that login nodes are stopped.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+    --target-group-arn $target_group_arn \
+    --region ${REGION})
+
+healthy_count=$(echo $target_healths | jq -r '.TargetHealthDescriptions[] | select(.TargetHealth.State == "healthy") | .Target.Id' | wc -l)
+unhealthy_count=$(echo $target_healths | jq -r '.TargetHealthDescriptions[] | select(.TargetHealth.State != "healthy") | .Target.Id' | wc -l)
+
+# Check if there are running login nodes
+total_nodes=$((healthy_count + unhealthy_count))
+if [[ $total_nodes -gt 0 ]]; then
+    echo "Login nodes are running."
```
jdeamicis: Shouldn't we exit 1 here?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -15,6 +15,21 @@ MUNGE_USER="<%= @munge_user %>"
...
+
+# Check if login nodes are running
+$CHECK_LOGIN_NODES_SCRIPT_PATH > $TEMP_FILE
```
jdeamicis: Why aren't you just checking the return code of the bash script? E.g.: https://stackoverflow.com/questions/49849957/bash-conditional-based-on-exit-code-of-command

jdeamicis (changes requested):
Good job.

The main thing we need to address is either handling pagination (even with `--no-paginate`), or do something else to avoid the risk of not finding the right load balancer if there are many ELBs in the account.

I left some other comments. Please address.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Thanks Jacopo, I refactor the logic here and the checking LoginNodes script. If exit 0, then means LoginNodes are stopped, so that we can continue the rotation script.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Yes, after the discussion, we have agreement that there should be exit 1. Have modified.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: I think we have discussed about this? Is this comment out of date?

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Have deleted this logic. Instead, now we are using the exit code to check if LoginNodes are stopped.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Have deleted this logic. Now if the LoginNodes section is not specified, the script will not exist.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done! Thank you for reminding!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
```diff
@@ -83,3 +84,14 @@ def execute_command(command, user = "root", timeout = 300, raise_on_error = true
...
+
+def is_login_nodes_pool_name_updated?
+  config_parameter_changed?(['LoginNodes', 'Pools', 0, 'Name'])
+end
+
+def is_login_nodes_removed?
+  require 'yaml'
+  config = YAML.safe_load(File.read(node['cluster']['cluster_config_path']))
+  previous_config = YAML.safe_load(File.read(node['cluster']['previous_cluster_config_path']))
+  previous_config.dig("LoginNodes") and !config.dig("LoginNodes")
```
EddyMM, line 96: Does this logic work in case of an `UPDATE_ROLLBACK`? In case of an error in another recipe, and a rollback is triggered, the configs are reversed i.e. `previous_config` becomes `config` and vice versa.

This function will assume login nodes are being added. It's worth testing this manually by mocking an error in another recipe section and observe if the script is removed after the UPDATE_ROLLBACK.

hehe7318: I don't see the point.

Case 1:
Create a cluster with LoginNodes, the script will exist. And we update the cluster with removing the LoginNodes section. The script will be deleted. And if the `UPDATE_ROLLBACK`, the LoginNodes will be added back. The other logic will create script back. That's what we want.

Case 2:
Create a cluster without LoginNodes, the script will not exist. We update cluster with adding new LoginNodes section. The script will be created by the other logic. And if `UPDATE_ROLLBACK`, the logic will delete the script. That's what we want too.

EddyMM: Case 2.
Thanks for clarifying this. The `previous_config` and `config` will be reversed during rollback but i've just seen the `:delete` action in the update recipe which ensures that the script is removed during a rollback.

EddyMM (dismissed):
Let's follow this up with a kitchen test to confirm the presence or absence of the script based on the variations of the cluster config.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -282,6 +283,21 @@
   retry_delay 2
 end unless redhat_on_docker?
 
+template "#{node['cluster']['scripts_dir']}/slurm/check_login_nodes_stopped.sh" do
+  source 'slurm/head_node/check_login_nodes_stopped.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0700'
+  variables(
+    cluster_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
```
jdeamicis: I'm not a big fan of this `||` logic. Are there any circumstances where one of these two variables does not exist? Do we have one of them that exists for certain?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -282,6 +283,21 @@
...
+    login_nodes_pool_name: lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) },
+    region: node['cluster']['region']
+  )
+  only_if do
+    node['cluster']['config'].dig(:LoginNodes)
+  end
+end
+
```
jdeamicis: Since we are re-using this block also in `update_head_node.rb`, can we move it to a sub-recipe that we call from both `config_head_node.rb` and `update_head_node.rb`?

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -201,7 +202,30 @@ def update_nodes_in_queue(strategy, queues)
...
+  mode '0700'
+  variables(
+    cluster_name: node['cluster']['cluster_name'] || node['cluster']['stack_name'],
+    login_nodes_pool_name: lazy { node['cluster']['config'].dig(:LoginNodes, :Pools, 0, :Name) },
+    region: node['cluster']['region']
+  )
+  only_if do
+    node['cluster']['config'].dig(:LoginNodes) &&
+      ::File.exist?(node['cluster']['previous_cluster_config_path']) &&
+      is_login_nodes_pool_name_updated?
```
jdeamicis: If you move this block to a sub-recipe, you should be able to add an if clause with these two conditions outside the `include_recipe` only here in `update_head_node.rb`.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/check_login_nodes_status.sh.erb
jdeamicis: Yes, I guess so.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
jdeamicis: Also, ideally we could add unit tests (ChefSpec tests) for the sub-recipe. But we can leave it for later.

jdeamicis (changes requested):

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hehe7318: I am not sure. I use this because other place that uses cluster_name use `||` logic like this.

Like https://github.com/hehe7318/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-environment/recipes/config/custom_actions_setup.rb line 24

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
jdeamicis: I guess we can resolve this conversation.

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -201,7 +201,17 @@ def update_nodes_in_queue(strategy, queues)
   only_if { ::File.exist?(node['cluster']['previous_cluster_config_path']) && is_slurm_database_updated? }
 end unless on_docker?
 
-# Update rotation script to update secret arn
+# Update check login nodes status script to update pool name
```
jdeamicis: Let's re-write this comment as:
```python3
# Cover the following two scenarios:
# - a cluster without login nodes is updated to have login nodes;
# - a cluster with login nodes is updated to use another pool name.
```

jdeamicis (dismissed):

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
jdeamicis: Xuanqi addressed this.

jdeamicis (approved):

PR #2477, opened 2023-10-02
Modify permissions of /etc/sysconfig directory

judouk:
Directory permissions set to 0644 incorrectly leading to permissions problems when a non-root user tries to read files from /etc/sysconfig

### Description of changes
* Changed default permissions of a directory. Suspect that this was originally done inadvertently as a directory should not have perms of 0644

### Tests
* n/a

### References
* Found during a debug of https://groups.google.com/g/slurm-users/c/TGYoK82KRvk


enrico-usai (approved):
Thanks @judouk for fixing this!

PR #2480, opened 2023-10-04
[Develop] Add a new internal storage option, EFS, keeping EBS as the default

dreambeyondorange:
[Develop] Add a new internal storage option, EFS, keeping EBS as the default

Add backup and restore recipes to move data in second stage images to the shared filesystem 
Create a new mount_internal_use_efs.rb resource to mount the internal shared filesystem 
Filter the efs filesystem arrays to mount internal shared fs in init and cx fses in config 
Refactor environment recipes as needed to adapt to having two separate modes for internal storage

### Tests
* Launched clusters with static compute nodes using both the EFS and EBS modes.
* Tested the default mode, EBS, with nothing specified in the config file.

### References
* https://github.com/aws/aws-parallelcluster/pull/5715
* https://github.com/aws/aws-parallelcluster-cookbook/pull/2465 - this is the PR that was reverted that has the inital changes to move to EFS for internal storage

### Checklist


cookbooks/aws-parallelcluster-environment/recipes/config.rb
```diff
@@ -18,9 +18,25 @@
...
+include_recipe 'aws-parallelcluster-environment::export_internal_use_ebs' if node['cluster']['internal_shared_storage_type'] == 'ebs'
+# Mount intel on compute and login nodes
+include_recipe 'aws-parallelcluster-environment::mount_intel_dir' if node['cluster']['internal_shared_storage_type'] == 'ebs'
 include_recipe 'aws-parallelcluster-environment::ebs'
 include_recipe 'aws-parallelcluster-environment::raid'
-include_recipe "aws-parallelcluster-environment::fs_mount"
+include_recipe "aws-parallelcluster-environment::efs"
+
+# Mount FSx directory with manage_fsx resource
+lustre "mount fsx" do
```
himani2411, line 38: Why remove `fs_mount.rb`? you could still keep this part of mounting Fsx Directories in `fs_mount` and keep `include_recipe "aws-parallelcluster-environment::fs_mount"`

dreambeyondorange: `fs_mount.rb` was an ambiguous recipe that didn't really seem to add much value.  The mounting of shared efs volumes gets much more complicated in this PR, since we need to distinguish internal use EFS vs customer use EFS, so it made sense to separate that out.  The FSX portion is just calling a resource, so wrapping that in a recipe doesn't help much.  IMO, this simplifies the code and provides better visibility into what is happening in this recipe.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Add support for Rocky Linux 8.
 - Install [Spack](https://spack.io) by default in cluster user's home directory.
 - Add support for `Scheduling/SlurmSettings/Database/DatabaseName` parameter to render `StorageLoc` in the slurmdbd configuration generated by ParallelCluster.
+- Add the option to use EFS storage instead of NFS exports from the head node root volume for intra-cluster shared ParallelCluster, Intel, Slurm, and Login Node data.
```
enrico-usai: same comments of the cli https://github.com/aws/aws-parallelcluster/pull/5715#discussion_r1347410562

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
```diff
@@ -475,7 +475,7 @@ suites:
   - name: mount_shared_compute
     run_list:
       - recipe[aws-parallelcluster-tests::setup]
-      - recipe[aws-parallelcluster-environment::mount_shared]
+      - recipe[aws-parallelcluster-environment::mount_internal_use_ebs]
```
enrico-usai, line 478: should we have a test for mount_internal_use_efs as well?

cookbooks/aws-parallelcluster-environment/recipes/config.rb
```diff
@@ -18,9 +18,25 @@
...
 end
 include_recipe 'aws-parallelcluster-environment::ephemeral_drives'
-# fs_update generates the shared storages mapping file so must be executed before shared storages recipes
-include_recipe 'aws-parallelcluster-environment::fs_update'
-include_recipe 'aws-parallelcluster-environment::shared_storages'
+# update_fs_mapping generates the shared storages mapping file so must be executed before shared storages recipes
+include_recipe 'aws-parallelcluster-environment::update_fs_mapping'
+# Export home and internal use dirs from the head node
+include_recipe 'aws-parallelcluster-environment::export_home'
+include_recipe 'aws-parallelcluster-environment::export_internal_use_ebs' if node['cluster']['internal_shared_storage_type'] == 'ebs'
```
enrico-usai: since we have 2 recipes under the same condition we could use:
```ruby
if node['cluster']['internal_shared_storage_type'] == 'ebs'
  # Export home and internal use dirs from the head node
  include_recipe 'aws-parallelcluster-environment::export_internal_use_ebs'
  # Mount intel on compute and login nodes
  include_recipe 'aws-parallelcluster-environment::mount_intel_dir'
end
```

cookbooks/aws-parallelcluster-environment/recipes/init.rb
```diff
@@ -13,12 +13,18 @@
 #
 
 include_recipe "aws-parallelcluster-environment::cfnconfig_mixed"
-include_recipe "aws-parallelcluster-environment::mount_shared"
 cloudwatch "Configure CloudWatch" do
   action :configure
 end
+
+include_recipe "aws-parallelcluster-environment::mount_internal_use_efs" if node['cluster']['internal_shared_storage_type'] == 'efs'
```
enrico-usai: we could have an if/elsif/else-raise block to catch the case where we're not passing this value.

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_efs.rb
```diff
@@ -0,0 +1,96 @@
...
+      user 'root'
+      group 'root'
+      mode '0755'
+      action :create
+      recursive true
+    end
+  end unless initial_shared_dir_array.empty?
+
+  # Unmount the root of the EFS after creating the shared directories
+  # TODO this doesn't seem to unmount the EFS
```
enrico-usai, line 61: do you mean that we have an additional mounted folder?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_compute.rb
```diff
@@ -24,7 +24,7 @@
   mode '0700'
 end
 
-include_recipe 'aws-parallelcluster-slurm::mount_slurm_dir'
+include_recipe 'aws-parallelcluster-slurm::mount_slurm_dir' if node['cluster']['internal_shared_storage_type'] == 'ebs'
```
enrico-usai: There is something strange here. I mean the code is correct, but seems counterintuitive that we're mounting the slurm dir only when using ebs.

I think it's better to move this if into the `mount_slurm_dir` recipe itself. It's responsibility of the recipe to know if must be executed or not. In this way we'll avoid the same check in the config_login and config_compute.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
dreambeyondorange: I'm not sure it's useful since we can't currently mount efs volumes in kitchen tests

cookbooks/aws-parallelcluster-environment/recipes/config.rb
dreambeyondorange: I considered this but it seems to me to be cleaner this way and only uses 2 lines of code instead of 4

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
dreambeyondorange: But I am planning to address kitchen tests in a separate follow up task

cookbooks/aws-parallelcluster-environment/recipes/config.rb
dreambeyondorange: I'll update this thanks!

cookbooks/aws-parallelcluster-environment/recipes/init.rb
dreambeyondorange: ill update this thanks!

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_efs.rb
dreambeyondorange: Yes, init_shared is still mounted after this step

cookbooks/aws-parallelcluster-slurm/recipes/config/config_compute.rb
dreambeyondorange: Yup, I can do this.

enrico-usai (approved):

PR #2486, opened 2023-10-06
[Develop] Add support for an external shared /home

dreambeyondorange:
Users can now specify `/home` as a mount point in SharedStorage.  This was previously a reserved directory When a user specifies this, the data in `/home` is transferred without replacement to the external filesystem. This means that in order to share the `/home` directory across clusters, users must specify the same security credentials when creating the clusters.

### Tests
* Tested launching a cluster with /home mounted to EFS, FSx, and the default EBS head node root volume
* Tested launching a cluster with an external fs and not /home mounted on EFS, FSx and default EBS.

### References
* https://github.com/aws/aws-parallelcluster/pull/5745

### Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Install [Spack](https://spack.io) by default in cluster user's home directory.
 - Add support for `Scheduling/SlurmSettings/Database/DatabaseName` parameter to render `StorageLoc` in the slurmdbd configuration generated by ParallelCluster.
 - Add the option to use EFS storage instead of NFS exports from the head node root volume for intra-cluster shared ParallelCluster, Intel, Slurm, and login node data.
+- Add the option to use EFS or FSx as external shared storage via the `SharedStorage` section of the config file for `/home`
```
gmarciani: Same comment as in https://github.com/aws/aws-parallelcluster/pull/5745

cookbooks/aws-parallelcluster-environment/recipes/config/export_home.rb
```diff
@@ -14,6 +14,18 @@
...
+
+# Check if home is a shared filesystem and return if it is because
+# there is nothing to export
+is_home_shared = false
+efs_shared_dir_array = node['cluster']['efs_shared_dirs'].split(',')
+fsx_shared_dir_array = node['cluster']['fsx_shared_dirs'].split(',')
+(efs_shared_dir_array + fsx_shared_dir_array).each do |dir|
+  is_home_shared = true if dir == "/home"
+end
+return if is_home_shared
```
gmarciani, line 27: What if the /home is the mountdir of an EBS shared storage and the internal shared storage is EBS?
We would end up skipping to export the /home in that case, that is not desirable.

cookbooks/aws-parallelcluster-environment/recipes/config/export_home.rb
```diff
@@ -14,6 +14,18 @@
 
 return if on_docker?
 
+return unless node['cluster']['internal_shared_storage_type'] == 'ebs'
```
gmarciani: What if the /home is the mountdir of an EBS shared storage and the internal shared storage is EBS?
We would end up skipping to export the /home in that case, that is not desirable.

cookbooks/aws-parallelcluster-environment/recipes/init/backup_home_shared_data.rb
```diff
@@ -0,0 +1,30 @@
...
+if node['cluster']['node_type'] == 'HeadNode'
+  # For each, backup the data to a temp location
+  # This is necessary to preserve any data in these directories that was
+  # generated during the build of ParallelCluster AMIs after converting to
+  # shared storage
+  bash "Backup /home" do
+    user 'root'
+    group 'root'
+    code <<-EOH
+      mkdir -p /tmp/home
```
gmarciani, line 26: Security best practice: use random names for temporary folders to prevent attacks based on name clashing. Eg: better to use `mktemp -d` rather than `mkdir`.

I imagine that you did this way because you need a deterministic name to use that folder somewhere else.
A solution may be to write the temporary folder into a file that is readable only by root and retrieve the folder name from that file where you need it.

cookbooks/aws-parallelcluster-environment/recipes/config/export_home.rb
dreambeyondorange: Yes, I remembered this a few minutes ago. We need a validator to disable EBS as an option in shared storage for /home. This effort is being tracked internally as a followup before the release

cookbooks/aws-parallelcluster-environment/recipes/init/backup_home_shared_data.rb
dreambeyondorange: How would an attacker be able to leverage this directory?  I'm not familiar with a name clashing attack

cookbooks/aws-parallelcluster-environment/recipes/config/export_home.rb
dreambeyondorange: Actually nevermind, I'll add EBS to this recipe

cookbooks/aws-parallelcluster-environment/recipes/config/efs.rb
```diff
@@ -11,31 +11,26 @@
...
-# Identify the customer use filesystems and store their data in arrays for the EFS resource
-efs_shared_dir_array.each_with_index do |dir, index|
-  next if node['cluster']['internal_shared_dirs'].include?(dir)
-  cx_shared_dir_array.push(dir)
-  cx_efs_fs_id_array.push(efs_fs_id_array[index])
-  cx_efs_encryption_array.push(efs_encryption_in_transit_array[index])
-  cx_efs_iam_array.push(efs_iam_authorization_array[index])
+# Identify the previously mounted filesystems and remove them from the set of filesystems to mount
+shared_dir_array.each_with_index do |dir, index|
+  next unless node['cluster']['internal_shared_dirs'].include?(dir) || dir == "/home"
```
gmarciani: dir could also be `home`, not only `/home`, right?

cookbooks/aws-parallelcluster-environment/recipes/init/backup_home_shared_data.rb
gmarciani: It's a scenario where an attacker compromises your data leveraging the fact that you temporarily put it in a less restrictive folder with a well known name and then copy back from there. If you use random name, it's harder for the attacker to know where to look at. The mitigations are:
1. restrict permissions to the temporary folder (the most important mitigation)
2. random names for temporary folders
3. minimize the exposure time of temporary data, i.e. explicitly remove it after being used)

gmarciani: This is not a blocking comment because the temp folder is created as root, so by default the folder has write permissions granted only to root (mitigation 1), and tmp data is explicitly remove after being used (mitigation 3)

cookbooks/aws-parallelcluster-environment/recipes/config/fsx.rb
```diff
@@ -0,0 +1,42 @@
...
+id_array = node['cluster']['fsx_fs_ids'].split(',')
+type_array = node['cluster']['fsx_fs_types'].split(',')
+shared_dir_array = node['cluster']['fsx_shared_dirs'].split(',')
+dns_name_array = node['cluster']['fsx_dns_names'].split(',')
+mount_name_array = node['cluster']['fsx_mount_names'].split(',')
+volume_junction_path_array = node['cluster']['fsx_volume_junction_paths'].split(',')
+
+# Identify the previously mounted filesystems and remove them from the set of filesystems to mount
+shared_dir_array.each_with_index do |dir, index|
+  next unless dir == "/home"
```
gmarciani: Same as above: in other conditions we assume that we may receive also `home` not only `/home`, so we should also exclude `home`.

That said I think we should enforce absolute path at cluster config level to avoid ambiguity at cookbook level, but this would be out of scope for this PR.

cookbooks/aws-parallelcluster-environment/recipes/config/mount_home.rb
```diff
@@ -11,12 +11,115 @@
...
+        fstype 'nfs'
+        options node['cluster']['nfs']['hard_mount_options']
+        retries 10
+        retry_delay 6
+      end
+    end
+  when 'raid'
+    case node['cluster']['node_type']
+    when 'HeadNode'
+      raid "add raid /home" do
```
gmarciani, line 106: *minor*: in line 20 we define `shared_storage['raid']` as an array to be consistent with the object type used by the other storage types. If we want to keep consistency, why not adopting the same approach adopted in the other when clause: `shared_storage['raid'].each_with_index do ...`?

cookbooks/aws-parallelcluster-environment/recipes/config/efs.rb
dreambeyondorange: indeed!

cookbooks/aws-parallelcluster-environment/recipes/init/backup_home_shared_data.rb
dreambeyondorange: got it thanks!

gmarciani (dismissed):

cookbooks/aws-parallelcluster-environment/recipes/config/fsx.rb
dreambeyondorange: Yes, I was surprised that we don't clean that up in the cli

dreambeyondorange: I'll clean this up as well

cookbooks/aws-parallelcluster-environment/recipes/config/mount_home.rb
dreambeyondorange: In this case, I wanted to avoid having to check for both /home and home.  And this is guaranteed to be correct since we identified the directory prior.  The array conversion doesn't help here.  I did that so I could iterate over all of them in a single `each`

gmarciani (approved):

PR #2487, opened 2023-10-09
[develop][fix] Add mock munge key logic in compute setup recipe during kitchen tests

hehe7318:
### Description of changes
* Fix Kitchen tests on compute nodes caused by not existed Munge Key

### References
* Failed kitchen tests on Jenkins
  * https://jenkins-commercial-dev.parallelcluster.ds9.nice.aws.a2z.com/blue/organizations/jenkins/kitchen_tests_3/detail/kitchen_tests_3/1407/pipeline

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/recipes/test/mock_munge_key.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/test/mock_munge_key.rb
@@ -5,10 +5,16 @@ munge_dirs.each do |munge_dir|
     mode '0700'
   end
 
-  file "#{munge_dir}/.munge.key" do
-    mode '0600'
-    content 'munge-key'
-    owner node['cluster']['munge']['user']
-    group node['cluster']['munge']['group']
+  bash "mock_munge_key" do
+    user 'root'
+    group 'root'
+    code <<-MOCK_KEY
+      set -e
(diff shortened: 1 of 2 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/test/mock_munge_key.rb
```diff
@@ -6,6 +6,7 @@
   end
 
   file "#{munge_dir}/.munge.key" do
+    mode '0600'
     content 'munge-key'
```
jdeamicis: `munge-key` is not a good content for a munge key :D

The content must be binary, and at least 32 bytes long. Please find a way to pass a binary blob to this file.

jdeamicis (changes requested):
We have to pass a binary blob to the munge key, otherwise the munge service will fail.

cookbooks/aws-parallelcluster-slurm/recipes/test/mock_munge_key.rb
hehe7318: Done! I use the same logic as the rotation script does(which means the config and update process do).

jdeamicis (approved):

PR #2494, opened 2023-10-11
[Spack] Changing the default Spack Root directory from /home to /opt

himani2411:
### Description of changes
* Changing the default location of Spack root folder from `/home` to `/opt/parallelcluster/shared`

### Tests
* Locally Creating an AMI and testing to install SW using Spack 

release-3.8 https://github.com/aws/aws-parallelcluster-cookbook/pull/2519

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/attributes/environment.rb
+++ b/cookbooks/aws-parallelcluster-environment/attributes/environment.rb
@@ -60,3 +60,5 @@ default['cluster']['internal_shared_dirs'] = [node['cluster']['shared_dir'], nod
 default['cluster']['internal_initial_shared_dir'] = "#{node['cluster']['base_dir']}/init_shared"
 
 default['cluster']['head_node_private_ip'] = nil
+
+default['cluster']['spack_shared_dir'] = "#{node['cluster']['shared_dir']}/spack"
--- a/cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
+++ b/cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
@@ -14,11 +14,8 @@
 unified_mode true
 default_action :setup
 
-property :spack_user, String, required: false,
-         default: node['cluster']['cluster_user']
-
 property :spack_root, String, required: false,
-         default: "/home/#{node['cluster']['cluster_user']}/spack"
+         default: node['cluster']['spack_shared_dir']
 
 action :install_spack do
   return if on_docker?
@@ -32,8 +29,8 @@ action :install_spack do
 
   git new_resource.spack_root do
     repository 'https://github.com/spack/spack'
-    user new_resource.spack_user
-    group new_resource.spack_user
+    user 'root'
+    group 'root'
   end
 
   template '/etc/profile.d/spack.sh' do
@@ -47,41 +44,18 @@ action :install_spack do
 
   spack = "#{new_resource.spack_root}/bin/spack"
   spack_configs_dir = "#{new_resource.spack_root}/etc/spack"
-  arch_target = `#{spack} arch -t`.strip
-
-  # Find libfabric version to be used in package configs
-  libfabric_version = nil
-  ::File.open(libfabric_path).each do |line|
-    if line.include?('Version:')
-      libfabric_version = line.split[1].strip
-      break
-    end
-  end
-
-  # Pull architecture dependent package config
-  begin
-    template "#{spack_configs_dir}/packages.yaml" do
-      cookbook 'aws-parallelcluster-environment'
-      source "spack/packages-#{arch_target}.yaml.erb"
-      owner new_resource.spack_user
-      group new_resource.spack_user
-      variables(libfabric_version: libfabric_version)
-    end
-  rescue Chef::Exceptions::FileNotFound
-    Chef::Log.warn "Could not find template for #{arch_target}"
-  end
 
   cookbook_file "#{spack_configs_dir}/modules.yaml" do
     cookbook 'aws-parallelcluster-environment'
     source 'spack/modules.yaml'
-    owner new_resource.spack_user
-    group new_resource.spack_user
+    owner 'root'
+    group 'root'
     mode '0755'
   end
 
   bash 'setup Spack' do
-    user new_resource.spack_user
-    group new_resource.spack_user
+    user 'root'
+    group 'root'
     code <<-SPACK
       source #{new_resource.spack_root}/share/spack/setup-env.sh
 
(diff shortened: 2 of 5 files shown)


cookbooks/aws-parallelcluster-environment/attributes/environment.rb
```diff
@@ -60,3 +60,5 @@
 default['cluster']['internal_initial_shared_dir'] = "#{node['cluster']['base_dir']}/init_shared"
 
 default['cluster']['head_node_private_ip'] = nil
+
+default['cluster']['spack_shared_dir'] = "#{node['cluster']['shared_dir']}/spack"
```
gmarciani, line 64: Have you tested if using this directory works also when the user logs into the login nodes rather than head node?
I think we should use a directory that is shared across the whole cluster (so also login nodes), but this directory is shared by default only between head and compute.

himani2411: Good Catch! will move the directory to node['cluster']['base_dir']

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
```diff
@@ -15,10 +15,10 @@
 default_action :setup
 
 property :spack_user, String, required: false,
-         default: node['cluster']['cluster_user']
+         default: 'root'
```
dreambeyondorange: I don't think the spack user should be root.  Was there a specific reason you needed this?

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
```diff
@@ -15,10 +15,10 @@
...
 
 property :spack_root, String, required: false,
-         default: "/home/#{node['cluster']['cluster_user']}/spack"
+         default: "#{node['cluster']['spack_shared_dir']}"
```
dreambeyondorange: I think you should be able to just reference the variable 
```
 default: node['cluster']['spack_shared_dir']
```

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
```diff
@@ -59,16 +59,14 @@
...
-  rescue Chef::Exceptions::FileNotFound
-    Chef::Log.warn "Could not find template for #{arch_target}"
+  
+  template "#{spack_configs_dir}/packages.yaml" do
+    cookbook 'aws-parallelcluster-environment'
+    source "spack/packages-#{arch_target}.yaml.erb"
+    owner new_resource.spack_user
+    group new_resource.spack_user
+    variables(libfabric_version: libfabric_version)
+    only_if { ::File.exist?( "spack/packages-#{arch_target}.yaml.erb") }
```
dreambeyondorange: Cookstyle is complaining about the space before "spack
```
only_if { ::File.exist?( "spack/packages-#{arch_target}.yaml.erb") }
                        ^
```

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
```diff
@@ -14,11 +14,14 @@
...
 default_action :setup
 
-property :spack_user, String, required: false,
-         default: node['cluster']['cluster_user']
-
 property :spack_root, String, required: false,
-         default: "/home/#{node['cluster']['cluster_user']}/spack"
+         default: node['cluster']['spack_shared_dir']
+
+property :spack, String, required: false,
```
dreambeyondorange: I don't think this should be a property because it is not something we want the caller of this resource to set, since it is determined by the root

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
```diff
@@ -14,11 +14,14 @@
...
+         default: "#{node['cluster']['spack_shared_dir']}/bin/spack"
+
+property :spack_configs_dir, String, required: false,
```
dreambeyondorange: same here as above

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
```diff
@@ -45,61 +48,33 @@
...
+      #{new_resource.spack} compiler add --scope site
 
       # Add external packages to Spack user's packages.yaml config
-      #{spack} external find --scope site
+      #{new_resource.spack} external find --scope site
 
       # Remove all autotools/buildtools packages. These versions need to be managed by spack or it will
       # eventually end up in a version mismatch (e.g. when compiling gmp).
-      #{spack} tags build-tools | xargs -I {} #{spack} config --scope site rm packages:{}
+      #{new_resource.spack} tags build-tools | xargs -I {} #{new_resource.spack} config --scope site rm packages:{}
```
dreambeyondorange: if you need to reference this multiple times, it probably makes sense to assign it to a local variable and ref that

himani2411: Thanks. I added the properties as a workaround for adding the same reference multiple time in install and config phases. Though I undertsand your point that this "workaround" I used can be avoided for better code readability.

cookbooks/aws-parallelcluster-environment/resources/spack/partial/_spack_common.rb
himani2411: These two proerties are something which will be determined by the Spack installer. But I get the point that these two property shouldnt eb allowed to be overriden anytime.

cookbooks/aws-parallelcluster-environment/test/controls/spack_spec.rb
```diff
@@ -34,13 +34,13 @@
...
-    it { should be_grouped_into (node['cluster']['spack']['user']).to_s }
+    it { should be_owned_by 'root' }
+    it { should be_grouped_into 'root' }
   end
 
   spack = "#{node['cluster']['spack']['root']}/bin/spack"
-  describe "spack commands can run as cluster default user #{node['cluster']['spack']['user']}" do
-    subject { bash("su - #{node['cluster']['spack']['user']} -c '#{spack} find'") }
+  describe "spack commands can run as cluster default user #{node['cluster']['cluster_user']}" do
+    subject { bash("su - #{node['cluster']['cluster_user']} -c 'sudo #{spack} find'") }
```
dreambeyondorange: I don't think this needs to be run as sudo does it?

himani2411: Nope it doesnt. Made those changes

dreambeyondorange (approved):

PR #2511, opened 2023-10-23
[develop][fix] Avoid store decoded binary data in bash varibles in rotation script

hehe7318:
### Description of changes
* Avoid store decoded binary data in bash varibles in rotation script

### Tests
* Manually tests passed

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
+++ b/cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
@@ -59,6 +59,7 @@ if [ -n "${SECRET_ARN}" ]; then
   fi
 
   # Decode munge key and write to munge.key file
+  decoded_key=$(printf "%s" "${encoded_key}" | base64 -d)
   if [ $? -ne 0 ]; then
     echo "Error decoding the munge key with base64"
     exit 1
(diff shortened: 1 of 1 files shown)


cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
```diff
@@ -59,20 +59,19 @@ if [ -n "${SECRET_ARN}" ]; then
   fi
 
   # Decode munge key and write to munge.key file
-  decoded_key=$(printf "%s" "${encoded_key}" | base64 -d)
```
jdeamicis: Please, leave this line (without using the stored value) so that we are able to capture any failure in the decoding of the fetched encoded key (also because otherwise the following if condition makes no sense).

jdeamicis (changes requested):
Added small comment, please address.

cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_munge_key.sh.erb
hehe7318: Great point! Thanks for your reminding Jacopo!

jdeamicis (approved):

PR #2534, opened 2023-11-13
Use jinja templating to refactor cloudwatch config to make is easier to add and remove an OS

hgreebe:
### Description of changes
* Uses jinja templating to refactor cloudwatch config to make is easier to add and remove an OS.
* Creates a list of default platforms that can be used in the platforms section of a log_config.

### Tests
* Tested locally by running write_cloudwatch_agent.json

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2543, opened 2023-11-15
Patch for use of jinja templating to refactor cloudwatch config

hgreebe:
### Description of changes
* Patch for https://github.com/aws/aws-parallelcluster-cookbook/pull/2534
* Make descriptions more in line with changes.


cookbooks/aws-parallelcluster-environment/files/cloudwatch/write_cloudwatch_agent_json.py
```diff
@@ -64,7 +64,7 @@ def add_instance_log_stream_prefixes(configs):
 
 
 def read_data(config_path):
-    """Read in log configuration data from config_path."""
+    """Read in and render jinja template of the log configuration data from config_path."""
```
himani2411: Why not add this in the `cookbooks/aws-parallelcluster-environment/files/cloudwatch/cloudwatch_agent_common_utils.py` if its being used in bot the other Python modules while keeping the comments to explain w.r.t these python modules

himani2411:
Also, I suggest adding more descriptive commit message, as someone would have to go though the PR's and the reviews in the PR to understand why these changes were made.

cookbooks/aws-parallelcluster-environment/resources/cloudwatch/partial/_cloudwatch_common.rb
```diff
@@ -137,7 +137,7 @@ def package_path
     path common_module_path
     user 'root'
     group 'root'
-    mode '0644'
+    mode '0755'
```
gmarciani, line 140: Just out of my curiosity: in the first version of this change the script was not executable by root, so how did you test the change if the script is executed by root on node bootstrap?

cookbooks/aws-parallelcluster-environment/files/cloudwatch/cloudwatch_agent_common_utils.py
```diff
@@ -15,3 +17,20 @@ def render_jinja_template(template_file_path):
     with open(template_file_path, "w", encoding="utf-8") as f:
         f.write(rendered_template)
     return template_file_path
+
+
+def read_jinja_template_at(path):
+    """Read the JSON file at path as a Jinja template."""
```
gmarciani: According to this function description, the function simply reads a json from a file, but it actually does more than that: it overwrites the input jinja template with the resulting rendered JSON.

I suggest to:
1. keep the input template and the resulting rendered template separated
2. do not modify/overwrite the input template, especially not as a side effect of a function that seems to be supposed to only read it.

gmarciani (changes requested):

cookbooks/aws-parallelcluster-environment/resources/cloudwatch/partial/_cloudwatch_common.rb
hgreebe: I ran the kitchen tests, logged in to the image, and ran whoami to check the permissions.

gmarciani: > whoami to check the permissions.

You mean to check the current user?

> I ran the kitchen tests

Do kitchen tests verify that the generate cwagent config is as expected?

cookbooks/aws-parallelcluster-environment/files/cloudwatch/cloudwatch_agent_config_util.py
```diff
@@ -69,26 +69,14 @@ def _read_json_at(path):
...
-
-
 def _read_schema():
     """Read the schema for the CloudWatch log configs file."""
     return _read_json_at(SCHEMA_PATH)
 
 
 def _read_log_configs():
     """Read the current version of the CloudWatch log configs file, cloudwatch_agent_config.json."""
-    return _read_jinja_template_at(LOG_CONFIGS_PATH)
```
gmarciani, line 91: why removing the render from the Jinja template?

cookbooks/aws-parallelcluster-environment/files/cloudwatch/cloudwatch_agent_config_util.py
```diff
@@ -132,6 +120,7 @@ def _validate_log_config_fields_uniqueness(input_json):
 def validate_json(input_json=None):
     """Ensure the structure of input_json matches that of the file it will be added to."""
```
gmarciani: This function is expected only to read something and validate it, but it actually write a file. Let's relfect this or separate it in two functions.

cookbooks/aws-parallelcluster-environment/files/cloudwatch/cloudwatch_agent_config_util.py
hgreebe: because now I am rendering it before call _read_log_configs

cookbooks/aws-parallelcluster-environment/resources/cloudwatch/partial/_cloudwatch_common.rb
hgreebe: The kitchen tests check that 
```
it 'validates cloudwatch' do
  is_expected.to run_execute('cloudwatch-config-validation').with(
    user: 'root',
    timeout: 300,
    environment: {
      'CW_LOGS_CONFIGS_SCHEMA_PATH' => config_schema_path,
      'CW_LOGS_CONFIGS_PATH' => config_data_path,
    },
    command: "#{cookbook_venv_path}/bin/python #{validator_script_path}"
  )
end
```
which calls and executes cloudwatch_agent_config_util.py as the root user. This python script uses the common module and the kitchen tests would only pass of the jinja template was rendered correctly because they check that it is a valid json file.

gmarciani (approved):

cookbooks/aws-parallelcluster-environment/resources/cloudwatch/partial/_cloudwatch_common.rb
himani2411: I am wondering if we should change the Permission. The permissions given to the new common utils module is similar to the permissions given to the other python modules, how is there no issue when existing python modules have same non-executable permissions.

PR #2549, opened 2023-11-17
[Develop] Add /home to internal EFS when not specified in SharedStorage

dreambeyondorange:
We are going to include /home with the new `SharedStorageType` parameter to support a larger set of use cases. With this change, when a user specifies `Efs` for `SharedStorageType` pcluster will replace all EBS exports with shared EFS storage.

### Tests
* Tested creating clusters locally that use a shared /home in the SharedStorage config and Efs for SharedStorageType will use SharedStorage over SharedStorageType for /home

### References
* https://github.com/aws/aws-parallelcluster/pull/5867

### Checklist


cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_efs.rb
```diff
@@ -57,6 +57,15 @@
...
   end unless initial_shared_dir_array.empty?
 
+  # Add the mount point for home
+  directory "#{node['cluster']['internal_initial_shared_dir']}/home" do
+    user 'root'
+    group 'root'
+    mode '0755'
+    action :create
+    recursive true
+  end
```
hanwen-pcluste: If `/home` mount is from the `SharedStorage` section, the `/home` directory in this EFS is not used. Do you think it is confusing for users?

dreambeyondorange: It could potentially be, but the users would never see it if they don't mount it.  I was trying to avoid remounting the initial shared dir for the shared home section

hanwen-pcluste (approved):

PR #2557, opened 2023-11-22
[develop] Fail build image when using a kernel version not supported by FSx for Lustre

enrico-usai:
Initially we decided to not fail, because when we released the first version of RH8 support only 8.4 was available, so we didn't want to release a broken functionality.

Anyway at that point we can use the same approach we're using for CentOS7 and fail the build if the kernel version is not compatible with FSx for Lustre.

This will permit us and users to discover this incompatibility at build image time, rather than at cluster creation time if using FSx storage.

### Tests
Verified with an AMI with kernel `4.18.0-425.13.1.el8_7.x86_64`

```
bash kitchen.ec2.sh environment-install converge lustre-rhel8
...
[2023-11-22T15:13:09+00:00] FATAL: RuntimeError: test_resource[resource] (aws-parallelcluster-tests::test_resource line 30) had an error:
RuntimeError: lustre[test] (aws-parallelcluster-tests::test_resource line 8) had an error:
RuntimeError: FSx for Lustre is not supported in kernel version 4.18.0-425.13.1.el8_7.x86_64 of RHEL 8.7, please update the kernel version
```

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/spec/unit/resources/lustre_setup_spec.rb
+++ b/cookbooks/aws-parallelcluster-environment/spec/unit/resources/lustre_setup_spec.rb
@@ -153,35 +153,15 @@ describe 'lustre:setup' do
     end
   end
 
-  context "on redhat lower than 8.2" do
-    cached(:chef_run) do
-      runner = runner(
-        platform: 'redhat', version: '8',
-        step_into: ['lustre']
-      ) do |node|
-        node.automatic['platform_version'] = "8.1"
-        node.override['cluster']['kernel_release'] = "4.18.0-147.9.1.el8"
-      end
-      Lustre.setup(runner)
-    end
-
-    it 'can not install lustre' do
-      expect { chef_run }.to(raise_error do |error|
-        expect(error).to be_a(Exception)
-        expect(error.message).to include("FSx for Lustre is not supported in this RHEL version 8.1, supported versions are >= 8.2")
-      end)
-    end
-  end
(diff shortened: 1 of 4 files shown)


jdeamicis (dismissed):

cookbooks/aws-parallelcluster-environment/spec/unit/resources/lustre_setup_spec.rb
```diff
@@ -185,8 +187,10 @@ def self.setup(chef_run)
       end
 
       it 'can not install lustre' do
-        is_expected.to write_log("FSx for Lustre is not supported in kernel version #{kernel_version} of RHEL #{platform_version}, please update the kernel version")
-          .with(level: :warn)
+        expect { chef_run }.to(raise_error do |error|
+          expect(error).to be_a(Exception)
+          expect(error.message).to include("FSx for Lustre is not supported in kernel version #{kernel_version} of RHEL #{platform_version}, please update the kernel version")
```
jdeamicis: Isn't there a similar test for Rocky? Maybe it's not really necessary though...

enrico-usai: Good point, we missed this test for rocky

enrico-usai: added

jdeamicis (approved):

PR #2563, opened 2023-11-23
[develop] Remove useless mechanism to retrieve OS minor version from RHEL and Rocky

enrico-usai:
This mechanism was used on CentOS 7, to retrieve the OS minor version from the kernel version and use it to download the right version of the Lustre client.

This mechanism is not required for RHEL and Rocky because the minor version automatically changes when updating the kernel. This is reflected in the `/etc/os-release` file too.

This change will permit us to automatically support new minor versions when they will be available, without the need to update the internal mapping.

### Tests
* `bash kitchen.ec2.sh environment-install destroy lustre-rhel8`

I had to temporary skip installation and test of lustre client on rhel/rocky on docker.
Latest dokken/rockylinux-8 image is 8.9 and there is no Lustre client available for it.

Previously installation was passing because we were setting a fake `kernel_release` version attribute through `kitchen.docker.yml`.

### References

https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/resources/lustre/lustre_centos7.rb
+++ b/cookbooks/aws-parallelcluster-environment/resources/lustre/lustre_centos7.rb
@@ -104,6 +104,13 @@ def find_centos_minor_version
   os_minor_version
 end
 
+def find_kernel_patch_version
+  # kernel release is in the form 3.10.0-1127.8.2.el7.x86_64
+  kernel_patch_version = node['cluster']['kernel_release'].match(/^\d+\.\d+\.\d+-(\d+)\..*$/)
+  raise "Unable to retrieve the kernel patch version from #{node['cluster']['kernel_release']}." unless kernel_patch_version
+  kernel_patch_version[1]
+end
+
 action_class do
   def base_url
     "https://fsx-lustre-client-repo.s3.amazonaws.com/#{base_url_prefix(arm_instance?)}/7.#{find_centos_minor_version}/#{node['kernel']['machine']}/"
--- a/cookbooks/aws-parallelcluster-environment/resources/lustre/partial/_install_lustre_centos_redhat.rb
+++ b/cookbooks/aws-parallelcluster-environment/resources/lustre/partial/_install_lustre_centos_redhat.rb
@@ -35,10 +35,3 @@ action :install_lustre do
 
   kernel_module 'lnet' unless redhat_on_docker? || rocky_on_docker?
 end
-
-def find_kernel_patch_version
-  # kernel release is in the form 3.10.0-1127.8.2.el7.x86_64
-  kernel_patch_version = node['cluster']['kernel_release'].match(/^\d+\.\d+\.\d+-(\d+)\..*$/)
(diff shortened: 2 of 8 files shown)


cookbooks/aws-parallelcluster-environment/spec/unit/resources/lustre_setup_spec.rb
```diff
@@ -203,6 +203,7 @@ def self.setup(chef_run)
             platform: platform, version: '8',
             step_into: ['lustre']
           ) do |node|
+            node.automatic['platform_version'] = "8.#{minor_version}"
```
lukeseawalker, line 215: is the mapping for minor_version done a couple of lines above, still needed?

jdeamicis (dismissed):

cookbooks/aws-parallelcluster-environment/spec/unit/resources/lustre_setup_spec.rb
enrico-usai: it's not strictly required for all the versions but there is a specific combination: 8.7 with a specific version of the kernel (425.13 and 425.3) that are not supported, here we're testing multiple combinations just to have more coverage.

cookbooks/aws-parallelcluster-environment/resources/lustre/lustre_centos7.rb
```diff
@@ -104,6 +104,13 @@ def find_centos_minor_version
   os_minor_version
 end
 
+def find_kernel_patch_version
+  # kernel release is in the form 3.10.0-1127.8.2.el7.x86_64
+  kernel_patch_version = node['cluster']['kernel_release'].match(/^\d+\.\d+\.\d+-(\d+)\..*$/)
+  raise "Unable to retrieve the kernel patch version from #{node['cluster']['kernel_release']}." unless kernel_patch_version
+  kernel_patch_version[1]
+end
+
```
jdeamicis, line 113: Can we have a unit test for this function?

cookbooks/aws-parallelcluster-environment/resources/lustre/partial/_install_lustre_centos_redhat.rb
```diff
@@ -28,17 +28,12 @@
     command "yum-config-manager --setopt=\*.skip_if_unavailable=1 --save"
   end
 
+  # TODO: restore installation on docker when Lustre is available for RH8.9
```
jdeamicis, line 31: Can't we force an 8.8 image for docker? Mmm. We are using the images from kitchen-dokken, so it looks like we can't. I wonder if we should start defining our own dokken images...

cookbooks/aws-parallelcluster-environment/resources/lustre/lustre_centos7.rb
enrico-usai: Added

jdeamicis (approved):

PR #2573, opened 2023-11-27
Add unit tests for cloudwatch agent scripts

hgreebe:
### Description of changes
* Add unit tests for cloudwatch agent scripts (Used to have 0% coverage)
* Allows for early discover of errors in python code in the cookbooks

### Tests
* Ran unit tests with coverage
   *  cloudwatch_agent_config_util.py - 70%
   *  write_cloudwatch_agent_json.py - 71%


test/unit/cloudwatch_agent/test_write_cloudwatch_agent_json.py
```diff
@@ -0,0 +1,209 @@
...
+def test_select_configs_for_scheduler():
+    configs = select_configs_for_scheduler(CONFIGS, "slurm")
+    assert_that(len(configs)).is_equal_to(2)
+    configs = select_configs_for_scheduler(CONFIGS, "awsbatch")
+    assert_that(len(configs)).is_equal_to(3)
+
+
+@pytest.mark.asyncio
+def test_select_configs_for_node_role():
+    configs = select_configs_for_node_role(CONFIGS, "ComputeFleet")
```
enrico-usai: A way to avoid code duplication for this kind of tests is to have a parametrize, first param is platform and second param is expected_length

You can also think to unify `test_select_configs_for_scheduler`, `test_select_configs_for_platform` etc, having a single parameter called dimension, given that you're doing the same kind of assertion for all of them.

cookbooks/aws-parallelcluster-environment/files/cloudwatch/cloudwatch_agent_config_util.py
```diff
@@ -136,12 +136,12 @@ def _write_log_configs(log_configs):
...
     log_configs = _read_log_configs()
-    log_configs["log_configs"].extend(input_json.get("log_configs"))
+    input_json["log_configs"].extend(log_configs.get("log_configs"))
 
     # NOTICE: the input JSON's timestamp_formats dict is the one that is
     # updated, so that those defined in the original config aren't clobbered.
-    log_configs["timestamp_formats"] = input_json["timestamp_formats"].update(log_configs.get("timestamp_formats"))
-    _write_log_configs(log_configs)
+    input_json["timestamp_formats"].update(log_configs.get("timestamp_formats"))
+    _write_log_configs(input_json)
```
himani2411: You are changing the file in which we are writing data. I am not sure if that is the correct approach

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+from assertpy import assert_that
+from cloudwatch_agent_config_util import validate_json, write_validated_json
+
+
+@pytest.mark.parametrize(
+    "error_type",
+    [None, "Duplicates", "Timestamp"],
+)
+def test_validate_json_content(mocker, error_type):
+    input_json = {
```
himani2411: Why not use the same Json file you have created  `test/unit/cloudwatch_agent/test_cloudwatch_agent/test_validate_json_input/config.json`

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+        elif error_type == "Timestamp":
+            assert_that(e.args[0]).contains("contains an invalid timestamp_format_key")
+
+
+@pytest.mark.parametrize(
+    "error_type",
+    [None, "FileNotFound", "Schema"],
+)
+def test_validate_json_invalid(mocker, error_type):
+    input_json = {
```
himani2411: Same as above

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+        "timestamp_formats": {
+            "month_first": "%b %-d %H:%M:%S",
+        },
+        "log_configs": [{"timestamp_format_key": "month_first", "log_stream_name": "test"}],
+    }
+
+    if error_type == "Schema":
```
himani2411: You can parameterize the Schema too, instead of using if-else conditions, use
```
@pytest.mark.parametrize(
    "error_type, expected_error_response, schema",
    [(None,None,{"type": "object", "properties": {"timestamp_formats": {"type": "object"}}}),
("FileNotFound","No file exists" ,"{"type": "object", "properties": {"timestamp_formats": {"type": "object"}}}"),
("Schema", "Failed validating 'type' in schema", "{"type": "string"}")],
)
```

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+            "cloudwatch_agent_config_util._read_log_configs",
+            return_value=input_json,
+        )
+
+    try:
+        validate_json(input_json)
+        assert_that(error_type).is_none()
+    except SystemExit as e:
+        assert_that(error_type).is_not_none()
+        if error_type == "FileNotFound":
```
himani2411: Parameterize the Error Reponse too

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+        "cloudwatch_agent_config_util._read_log_configs",
+        return_value=input_json,
+    )
+
+    try:
+        validate_json(input_json)
+        assert_that(error_type).is_none()
+    except SystemExit as e:
+        assert_that(error_type).is_not_none()
+        if error_type == "Duplicates":
```
himani2411: Parameterize the Error Reponse using @pytest.mark.parametrize(

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+            assert_that(e.args[0]).contains("The following log_stream_name values are used multiple times: test")
+        elif error_type == "Timestamp":
+            assert_that(e.args[0]).contains("contains an invalid timestamp_format_key")
+
+
+@pytest.mark.parametrize(
+    "error_type",
+    [None, "FileNotFound", "Schema"],
+)
+def test_validate_json_invalid(mocker, error_type):
```
himani2411: You can combine these and `test_validate_json_content` functions if you use `@pytest.mark.parametrize` in the example I gave below

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,160 @@
...
+    except SystemExit as e:
+        assert_that(error_type).is_not_none()
+        if error_type == "FileNotFound":
+            assert_that(e.args[0]).contains("No file exists")
+        elif error_type == "Schema":
+            assert_that(e.args[0]).contains("Failed validating 'type' in schema")
+
+
+@pytest.mark.parametrize("no_error", [True, False])
+def test_validate_json_input(mocker, test_datadir, no_error):
```
himani2411: So far all the functions have been to validate_json, with or without the input. Could you rename the functions to something else or add comments of what this is supposed to test.

test/unit/cloudwatch_agent/test_write_cloudwatch_agent_json.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+@pytest.mark.parametrize(
+    "node",
+    [
+        {"role": "ComputeFleet", "length": 2},
+        {"role": "HeadNode", "length": 3},
+    ],
+)
+def test_select_configs_for_node_role(node):
```
himani2411: Either lets combine this with the `test_select_configs_for_dimesion` or keep the function separate for all 3.

test/unit/cloudwatch_agent/test_write_cloudwatch_agent_json.py
```diff
@@ -0,0 +1,194 @@
...
+        {"scheduler": "awsbatch", "length": 3},
+    ],
+)
+def test_select_configs_for_dimesion(dimensions):
+    if "platform" in dimensions.keys():
+        configs = select_configs_for_platform(CONFIGS, dimensions["platform"])
+        assert_that(len(configs)).is_equal_to(dimensions["length"])
+    else:
+        configs = select_configs_for_scheduler(CONFIGS, dimensions["scheduler"])
+        assert_that(len(configs)).is_equal_to(dimensions["length"])
```
himani2411: This line of code is same for both of them. You can keep that as common

test/unit/cloudwatch_agent/test_write_cloudwatch_agent_json.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+@pytest.mark.parametrize(
+    "info",
+    [
+        {"node_info": {"dcv_enabled": "head_node", "directory_service": {"enabled": "true"}}, "length": 3},
+        {"node_info": {"directory_service": {"enabled": "true"}}, "length": 2},
+        {"node_info": {"enabled": "true"}, "length": 1},
+    ],
```
himani2411, line 130: You can keep this as to reduce using the usage of dict keys
 @pytest.mark.parametrize(
    "node_info, length",
    [
        ({"dcv_enabled": "head_node", "directory_service": {"enabled": "true"}}, "length": 3),
        ({"directory_service": {"enabled": "true"}}, "length": 2),
       ({"enabled": "true"}, "length": 1),
    ],

test/unit/cloudwatch_agent/test_write_cloudwatch_agent_json.py
```diff
@@ -0,0 +1,194 @@
...
+    configs = add_timestamps(CONFIGS, timestamp_formats)
+    for config in configs:
+        timestamp_format = timestamp_formats[config["timestamp_format_key"]]
+        assert_that(config["timestamp_format"]).is_equal_to(timestamp_format)
+
+
+def test_filter_output_fields():
+    desired_keys = ["log_stream_name", "file_path", "timestamp_format", "log_group_name"]
+    configs = filter_output_fields(CONFIGS)
+    for config in configs:
```
himani2411, line 151: I am not sure if using part of the function code that you are testing in the test function itself is a good strategy.

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
hgreebe: because I wanted to test the mechanism where validate_json takes in the input_json parameter

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani, line 1: Nit: change the copyright year to 2023

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,167 @@
...
+                mocker.patch(
+                    "builtins.open",
+                    side_effect=ValueError,
+                )
+            else:
+                mocker.patch(
+                    "builtins.open",
+                    return_value=input_file,
+                )
+            validate_json()
```
gmarciani, line 131: What is this method validating since it does not get any input?

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,167 @@
...
+
+    mocker.patch(
+        "cloudwatch_agent_config_util._validate_timestamp_keys",
+        return_value=None,
+    )
+    try:
+        with open(input_file_path, encoding="utf-8") as input_file:
+            if not no_error:
+                mocker.patch(
+                    "builtins.open",
```
gmarciani, line 123: In general, this type of mocking is risky because you're patching all the open() functions, whereas mocks should target specifically the part of the code you're testing.

Probably in this case it's not worth it to change because you expect just one open() call in the stack, but just to let you know. 

As an alternative, you can wrap the open call into a cloudwatch_agent_config_util and mock only that call.

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
```diff
@@ -0,0 +1,167 @@
...
+        return_value=input_json,
+    )
+
+    mocker.patch("os.environ.get", return_value=output_file)
+
+    write_validated_json(input_json2)
+
+    expected = os.path.join(test_datadir, "output.json")
+
+    with open(output_file, encoding="utf-8") as f, open(expected, encoding="utf-8") as exp_f:
```
gmarciani, line 166: In this test you're verifying that the expected file is written.
That's ok as long as you create files from tests in tmp folder, but when testing functions that are expected to write files you can also avoid the actual write of the file and mock it instead.
In fact, you're not interested in testing that the `write` function actually writes a file, but that the function you're testing is calling the write function with the expected argument. If you test the actual file creation you're also testing the built in write function.

gmarciani (approved):

test/unit/cloudwatch_agent/test_cloudwatch_agent.py
hgreebe: it validates the json file that is stored at input_file_path

PR #2583, opened 2023-11-30
[Test] Add Rocky8 to tests.

gmarciani:
### Description of changes
Add Rocky8 to kitchen and spec tests.

### Tests
* Verified that ` ./kitchen.ec2.sh platform-install test nvidia-gdrcopy` launches test for Rocky8.

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/spec/unit/resources/cloudwatch_spec.rb
+++ b/cookbooks/aws-parallelcluster-environment/spec/unit/resources/cloudwatch_spec.rb
@@ -33,7 +33,16 @@ describe 'cloudwatch:setup' do
       cached(:signature_path) { "#{package_path}.sig" }
 
       context "when not on arm" do
-        cached(:platform_url_component) { platform == 'amazon' ? 'amazon_linux' : platform }
+        cached(:platform_url_component) do
+          case platform
+          when 'amazon'
+            'amazon_linux'
+          when 'rocky'
+            'redhat'
+          else
+            platform
+          end
+        end
         cached(:package_url) { "#{package_url_prefix}/#{platform_url_component}/amd64/latest/amazon-cloudwatch-agent.#{package_extension}" }
         cached(:chef_run) do
           runner = runner(platform: platform, version: version, step_into: ['cloudwatch']) do |node|
@@ -105,7 +114,7 @@ describe 'cloudwatch:setup' do
           case platform
           when 'amazon'
             'amazon_linux'
-          when 'centos'
+          when 'centos', 'rocky'
             'redhat'
           else
             platform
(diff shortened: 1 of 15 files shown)


cookbooks/aws-parallelcluster-environment/spec/unit/resources/cloudwatch_spec.rb
```diff
@@ -33,7 +33,16 @@ def self.configure(chef_run)
       cached(:signature_path) { "#{package_path}.sig" }
 
       context "when not on arm" do
-        cached(:platform_url_component) { platform == 'amazon' ? 'amazon_linux' : platform }
+        cached(:platform_url_component) do
+          case platform
+          when 'amazon'
+            'amazon_linux'
+          when 'rocky'
```
jdeamicis, line 40: Any reason why CentOS x86_64 differs from CentOS arm (see your next change at line 117)?

cookbooks/aws-parallelcluster-platform/spec/unit/resources/install_packages_spec.rb
```diff
@@ -53,6 +55,22 @@ def self.setup(chef_run)
...
+        end
+
+        it 'installs kernel source' do
+          is_expected.to run_bash('Install kernel source')
+            .with(user: 'root')
+            .with_code(/set -e/)
+            .with_code(/dnf install -y #{kernel_source_package}-#{kernel_source_package_version} --releasever #{platform}/)
+            .with_code(/dnf clean all/)
+        end
+
```
jdeamicis: I would move this up and do the following:
```ruby
if %w(amazon centos redhat rocky).include?(platform)
  it 'installs default packages' do
    # This is common even to Rocky 8
    ...
  end
  
  if platform == 'rocky'
    it 'installs kernel source' do
       # Rocky methodology via bash block
       ...
    end
  else
    it 'install kernel source' do
      # Other RHEL derivatives methodology via `package`
      ...
    end
  end
end
```

jdeamicis (dismissed):
Approved with two comments.

cookbooks/aws-parallelcluster-environment/spec/unit/resources/cloudwatch_spec.rb
gmarciani: There is no CWagent for CentOS on ARM, so we fall back to RHEL package.
This is not the case for CentOS x86.

cookbooks/aws-parallelcluster-platform/spec/unit/resources/install_packages_spec.rb
gmarciani: Done

jdeamicis (approved):

PR #2611, opened 2024-01-08
Change Shellcheck severity level to warning and make fixes

judysng:
### Description of changes
* Changed the shellcheck severity level to warning to discover potential problems earlier
* Fixed the warnings, the warnings are in this check in my [fork](https://github.com/judysng/aws-parallelcluster-cookbook/actions/runs/7414056129/job/20174303506)

### Tests
* Tested the bash line changes locally
* Ran bump-version.sh

### References
* Also changed severity level to warning in the [CLI](https://github.com/aws/aws-parallelcluster/pull/5996) and [node](https://github.com/aws/aws-parallelcluster-node/pull/610)

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/files/setup-ephemeral-drives.sh
+++ b/cookbooks/aws-parallelcluster-environment/files/setup-ephemeral-drives.sh
@@ -17,11 +17,10 @@ LVM_PATH="/dev/${LVM_VG_NAME}/${LVM_NAME}"
 LVM_ACTIVE_STATE="a"
 FS_TYPE="ext4"
 MOUNT_OPTIONS="noatime,nodiratime"
-# cfn_ephemeral_dir is set in the environment by cfnconfig sourcing
+# shellcheck disable=SC2154 # cfn_ephemeral_dir is set in the environment by cfnconfig sourcing
 INPUT_MOUNTPOINT="${cfn_ephemeral_dir}"
 
 function log {
-  SCRIPT=$(basename "$0")
   MESSAGE="$1"
   echo "ParallelCluster - ${MESSAGE}"
 }
@@ -73,7 +72,7 @@ function print_block_device_mapping {
 function check_instance_store {
   if ls /dev/nvme* >& /dev/null; then
     IS_NVME=1
-    MAPPINGS=$(realpath --relative-to=/dev/ -P /dev/disk/by-id/nvme*Instance_Storage* | grep -v "*Instance_Storage*" | uniq)
+    MAPPINGS=$(realpath --relative-to=/dev/ -P /dev/disk/by-id/nvme*Instance_Storage* | grep -v "Instance_Storage" | uniq)
   else
     IS_NVME=0
     set_imds_token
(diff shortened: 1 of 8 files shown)


cookbooks/aws-parallelcluster-environment/files/setup-ephemeral-drives.sh
```diff
@@ -73,7 +72,7 @@ function print_block_device_mapping {
 function check_instance_store {
   if ls /dev/nvme* >& /dev/null; then
     IS_NVME=1
-    MAPPINGS=$(realpath --relative-to=/dev/ -P /dev/disk/by-id/nvme*Instance_Storage* | grep -v "*Instance_Storage*" | uniq)
+    MAPPINGS=$(realpath --relative-to=/dev/ -P /dev/disk/by-id/nvme*Instance_Storage* | grep -v "Instance_Storage" | uniq)
```
judysng: For this [shellcheck warning SC2063](https://www.shellcheck.net/wiki/SC2063), I deleted the `*`'s as advised, but I am unsure what behavior we want:

Before, the `grep -v "*Instance_Storage*"` wouldn't filter out any results, because the `*` at the beginning wouldn't match anything, so the statement would've been the same as `MAPPINGS=$(realpath --relative-to=/dev/ -P /dev/disk/by-id/nvme*Instance_Storage* | uniq)`. Is that what we want instead?

With the deletion of the `*`'s, `MAPPINGS` would be empty, because `grep` will filter out all results since all the results contain `Instance_Storage`.

util/cinc-install.sh
```diff
@@ -75,6 +75,7 @@ checksum_mismatch() {
 unable_to_retrieve_package() {
   echo "Unable to retrieve a valid package!"
   report_bug
+  # shellcheck disable=SC2154
   echo "Metadata URL: $metadata_url"
```
judysng: I am unsure where `metadata_url` comes from

cookbooks/aws-parallelcluster-environment/files/setup-ephemeral-drives.sh
hanwen-pcluste: Let's do this:
`MAPPINGS=$(realpath --relative-to=/dev/ -P /dev/disk/by-id/nvme*Instance_Storage* | grep -v "\*Instance_Storage\*" | uniq)`


Explanation: The command does not filter out any results on instance types without instance store. If the command is run on instance types with instance store, there will be results.

The `-v` in `grep` is revert filter. This is to filter out `/dev/disk/by-id/nvme*Instance_Storage*` from `realpath` when running on instance types without instance store.

My proposal uses `\` to escape `*` to tell shellcheck the `*`s are not Regex keywords

I usually sub-command one by one to understand a command so big :)

util/cinc-install.sh
enrico-usai: As you can see in the first comment of this bash script, this file is a copy of the https://omnitruck.cinc.sh/install.sh file, with some modification (you can see the modification in the comments on top of the file).

During the modification we removed the part to download the installer from cinc website and we're downloading it from S3 instead. You can remove all the lines were we're printing the metadata_url, it's useless for us.

cookbooks/aws-parallelcluster-environment/files/setup-ephemeral-drives.sh
judysng: Ohh got it, thanks for the explanation. I have updated the line

util/cinc-install.sh
judysng: Makes sense, I deleted the lines with metadata_url. Thanks!

hanwen-pcluste (approved):

PR #2618, opened 2024-01-17
[DFSM] Make compute nodes save cluster config version to DynamoDB.

gmarciani:
### Description of changes
Make compute nodes save cluster config version to DynamoDB during create (`finalize` recipe) and update (`update_compute` recipe).

In particular every compute node stores the following record on the cluster main DDB table (`parallelcluster-$CLUSTERNAME`):

```
{
  "Id": {
    "S": "CLUSTER_CONFIG.i-0ae39403fd72e34e4"
  },
  "Data": {
    "M": {
      "cluster_config_version": {
        "S": "HoqyEZYBkMig3gSxaMARUv0NGcG0rrso"
      },
      "lastUpdateTime": {
        "S": "2024-01-16 18:59:18 UTC"
      }
    }
  }
}
```

Also:
1. Added new spec tests to cover changed/new recipes.
1. Removed the dummy log line that was emitted in the initial PR for DFSM to prove updates are triggered.

Note: we derive the DDB table name on the fly from clustername (parallelcluster-$CLUSTERNAME) rather than reading it from dna.json. This is because compute node are currently missing such info in their dna.json. We will add it as part of another iteration.

### Tests
Executed the below manual tests and verified that every compute node stores the correct cluster config version and the last update time in the main DDB table.
1. Created a cluster
1. Updated the same cluster
1. Updated the same cluster with injected error causing rollback (forced update adding a non existing external EBS volume): compute nodes update the cluster config version with the faulty one, then the previous one is restored on the DB as consequence of the rollback. Notice that the version reported in S3 is the faulty one as it is the last one being submitted by the user.

Unit Tests:
1. Executed all existing spec tests on the aws-parallelcluster-slurm cookbook
3. Executed all new spec tests.

Integ Tests:
1. Test `test_update_slurm` succeeded

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
+++ b/cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# Copyright:: 2024 Amazon.com, Inc. and its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Saves on the DynamoDB table 'parallelcluster-$CLUSTERNAME' the correspondence between
+# the instance id and the cluster config version deployed on that instance.
+# This is done to understand what is the deployed cluster config in each node, for example, during a CFN Stack update.
+def save_instance_config_version_to_dynamodb
+  unless on_docker? || kitchen_test? && !node['interact_with_ddb']
+    # TODO: the table name should be read from a dedicated node attribute,
+    #   but it is currently missing in the dna.json for compute nodes.
+    table_name = "parallelcluster-#{node['cluster']['cluster_name']}"
+    item_id = "CLUSTER_CONFIG.#{node['ec2']['instance_id']}"
+    item_data = "{\"cluster_config_version\": {\"S\": \"#{node['cluster']['cluster_config_version']}\"}, \"lastUpdateTime\": {\"S\": \"#{Time.now.utc}\"}}"
+    item = "{\"Id\": {\"S\": \"#{item_id}\"}, \"Data\": {\"M\": #{item_data}}}"
+
+    execute "Save cluster config version to DynamoDB" do
+      command "#{cookbook_virtualenv_path}/bin/aws dynamodb put-item" \
+                " --table-name #{table_name}"\
+                " --item '#{item}'" \
+                " --region #{node['cluster']['region']}"
+      retries 3
+      retry_delay 5
+    end
+  end
+end
--- a/cookbooks/aws-parallelcluster-slurm/recipes/finalize/finalize_compute.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/finalize/finalize_compute.rb
@@ -4,7 +4,7 @@
 # Cookbook:: aws-parallelcluster-slurm
 # Recipe:: finalize_compute
 #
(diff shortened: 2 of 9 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/finalize/finalize_compute.rb
```diff
@@ -4,7 +4,7 @@
 # Cookbook:: aws-parallelcluster-slurm
 # Recipe:: finalize_compute
 #
-# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2024 Amazon.com, Inc. and its affiliates. All Rights Reserved.
```
enrico-usai: minor: 2013-2024

cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
```diff
@@ -0,0 +1,28 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+def save_cluster_config_version_to_dynamodb
```
enrico-usai: let's add a comment for this function, e.g.:
```
The correspondence between instance id and cluster config version is saved in the DB to understand the status of each node for example during a CFN Stack update.
```

cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
```diff
@@ -0,0 +1,28 @@
...
+  table_name = "parallelcluster-#{node['cluster']['cluster_name']}"
```
enrico-usai: let's add a TODO here to remember to change it.

cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
enrico-usai: what about: `store_instance_config_version` or `save_instance_config_version_to_dynamodb`?

I'd like to have instance in the name to clarify that we're saving this "mapping" and not just the config version.

"cluster" might be omitted

cookbooks/aws-parallelcluster-slurm/recipes/update/update_compute.rb
```diff
@@ -0,0 +1,20 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+unless on_docker? || kitchen_test? && !node['interact_with_ddb']
```
enrico-usai: what about having all this logic `unless on_docker? || kitchen_test? && !node['interact_with_ddb']` into the `save_cluster...` function?

Is responsibility of the function to know if it can be called during kitchen tests or in docker, it's not a responsibility of the caller to know how it's implemented. This will help to avoid having the same guards in every place we're calling it.

enrico-usai (commented):
Thanks for increasing chefspec coverage!

cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
gmarciani: Sure, I like save_instance_config_version_to_dynamodb

cookbooks/aws-parallelcluster-slurm/recipes/update/update_compute.rb
gmarciani: Very good point! I completely agree

enrico-usai (approved):

PR #2620, opened 2024-01-22
Process the config option DeploymentSettings/DefaultUserHome to change the default user's home directory

dreambeyondorange:
…is set to local

Refactored the cookbook to use an attribute for the default user's home directory so that it can be
changed when using the new config option.  Also simplified the users attribute file since the exta OS
specific files were not adding any value and duplicating code.

### Tests
* Tested the changes manually with different clusters using both Shared and Local options.

### References
* https://github.com/aws/aws-parallelcluster/pull/6028/files

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 ------
 
 **ENHANCEMENTS**
+- Add the configuration parameter `DeploymentSettings/DefaultUserHome` to allow users to move the default user's home directory to `/local/home` instead of `/home` (default).
 - Add support for installing Intel OneAPI Base Toolkit and HPC Toolkit, and Intel Python.
   - Intel OneAPI Base Toolkits: 2023.2.0
   - Intel OneAPI HPC Toolkits: 2023.2.0
--- a/cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
+++ b/cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
@@ -752,3 +752,39 @@ suites:
       dependencies:
         - recipe:aws-parallelcluster-platform::directories
         - resource:spack:setup
+  - name: default_user_local_home_head_node
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-environment::config_default_user_home]
+    verifier:
+      controls:
+        - local_default_user_home
+    attributes:
+      cluster:
+        node_type: 'HeadNode'
+        scheduler: 'slurm'
+        default_user_home: 'local'
+  - name: default_user_local_home_compute
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-environment::config_default_user_home]
+    verifier:
+      controls:
+        - local_default_user_home
+    attributes:
+      cluster:
+        node_type: 'ComputeFleet'
+        scheduler: 'slurm'
+        default_user_home: 'local'
+  - name: default_user_local_home_login
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-environment::config_default_user_home]
+    verifier:
+      controls:
+        - local_default_user_home
+    attributes:
+      cluster:
+        node_type: 'LoginNode'
+        scheduler: 'slurm'
+        default_user_home: 'local'
--- a/cookbooks/aws-parallelcluster-environment/recipes/init.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init.rb
@@ -17,6 +17,9 @@ cloudwatch "Configure CloudWatch" do
   action :configure
 end
 
+# move the default user dir out of /home if the config param is set to 'local'
+include_recipe "aws-parallelcluster-environment::config_default_user_home"
+
 case node['cluster']['shared_storage_type']
 when 'efs'
   include_recipe "aws-parallelcluster-environment::mount_internal_use_efs"
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+#
+# Copyright:: 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+return if node['cluster']['default_user_home'] == 'shared'
+
+# Backup the cluster user's default home directory
+bash "Backup #{node['cluster']['cluster_user_home']}" do
+  user 'root'
+  group 'root'
+  code <<-EOH
+    set -e
+    if [ -d /tmp#{node['cluster']['cluster_user_home']} ]; then
+      echo "/tmp#{node['cluster']['cluster_user_home']} exists!"
+      exit 1
+    else
+      mkdir -p /tmp#{node['cluster']['cluster_user_home']}
+    fi
+    rsync -a #{node['cluster']['cluster_user_home']}/ /tmp#{node['cluster']['cluster_user_home']}
+  EOH
+end
+
+# move the cluster user's default home directory
+bash "Move #{node['cluster']['cluster_user_home']}" do
+  user 'root'
+  group 'root'
+  code <<-EOH
+    set -e
+    mkdir -p #{node['cluster']['cluster_user_local_home']}
+    rsync -a /tmp#{node['cluster']['cluster_user_home']}/ #{node['cluster']['cluster_user_local_home']}
+    usermod -d #{node['cluster']['cluster_user_local_home']} #{node['cluster']['cluster_user']}
+    chown -R #{node['cluster']['cluster_user']}: #{node['cluster']['cluster_user_local_home']}
+    rm -rf /tmp#{node['cluster']['cluster_user_home']}
+    rm -rf #{node['cluster']['cluster_user_home']}
+  EOH
+end
+
+node.override['cluster']['cluster_user_home'] = node['cluster']['cluster_user_local_home']
--- a/cookbooks/aws-parallelcluster-platform/recipes/config/cluster_user.rb
+++ b/cookbooks/aws-parallelcluster-platform/recipes/config/cluster_user.rb
@@ -12,52 +12,87 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+node.override['cluster']['cluster_user_home'] = node['cluster']['cluster_user_local_home'] if node['cluster']['default_user_home'] == 'local'
+
 case node['cluster']['node_type']
 when 'HeadNode'
   # Setup cluster user
   user node['cluster']['cluster_user'] do
     manage_home true
     comment 'AWS ParallelCluster user'
-    home "/home/#{node['cluster']['cluster_user']}"
+    home "#{node['cluster']['cluster_user_home']}"
     shell '/bin/bash'
   end
 
   # Setup SSH auth for cluster user
   bash "ssh-keygen" do
-    cwd "/home/#{node['cluster']['cluster_user']}"
+    cwd "#{node['cluster']['cluster_user_home']}"
     code <<-KEYGEN
       set -e
       su - #{node['cluster']['cluster_user']} -c \"ssh-keygen -q -t ed25519 -f ~/.ssh/id_ed25519 -N ''\"
     KEYGEN
-    not_if { ::File.exist?("/home/#{node['cluster']['cluster_user']}/.ssh/id_ed25519") }
+    not_if { ::File.exist?("#{node['cluster']['cluster_user_home']}/.ssh/id_ed25519") }
   end
 
   bash "copy_and_perms" do
-    cwd "/home/#{node['cluster']['cluster_user']}"
+    cwd "#{node['cluster']['cluster_user_home']}"
     code <<-PERMS
       set -e
       su - #{node['cluster']['cluster_user']} -c \"cat ~/.ssh/id_ed25519.pub >> ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/authorized_keys && touch ~/.ssh/authorized_keys_cluster\"
     PERMS
-    not_if { ::File.exist?("/home/#{node['cluster']['cluster_user']}/.ssh/authorized_keys_cluster") }
+    not_if { ::File.exist?("#{node['cluster']['cluster_user_home']}/.ssh/authorized_keys_cluster") }
+  end
+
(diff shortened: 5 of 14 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 ------
 
 **ENHANCEMENTS**
+- Add the config option `DefaultUserHome` under `HeadNode` to allow users to move the default user's home directory to `/local/home` instead of `/home` (default)
```
gmarciani: Same comment as https://github.com/aws/aws-parallelcluster/pull/6028#discussion_r1462940407

cookbooks/aws-parallelcluster-environment/recipes/init/move_default_user.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+#
+# Copyright:: 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: Copyright year: 2024.

cookbooks/aws-parallelcluster-environment/recipes/init.rb
```diff
@@ -17,6 +17,9 @@
   action :configure
 end
 
+# move the default user out of home if the config param is set
+include_recipe "aws-parallelcluster-environment::move_default_user" if node['cluster']['default_user_home'] == 'local'
```
gmarciani: For better definition of responsibilities, I suggest to:
* rename the recipe as `default_user_home.rb`: its responsibility is to configure the default user home, i.e. do nothing if default_user_home is `shared` and do something if it is `local`
* always include the recipe `default_user_home.rb` in the `init` recipe: ideally, the responsibility of the `init` recipe is to include the respective recipes for every component that could be configured, even if the configuration is "do nothing" in some cases.
* let the recipe `default_user_home.rb` decide what to do according to the value of `default_user_home`

cookbooks/aws-parallelcluster-environment/recipes/init/move_default_user.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Backup the cluster user's default home directory
+bash "Backup #{node['cluster']['cluster_user_home']}" do
+  user 'root'
+  group 'root'
+  code <<-EOH
```
gmarciani: Juts out of my curiosity: why EOH, what does H stand for?

cookbooks/aws-parallelcluster-environment/recipes/init/move_default_user.rb
```diff
@@ -0,0 +1,37 @@
...
+    mkdir -p /tmp#{node['cluster']['cluster_user_home']}
```
gmarciani: How is cluster_user_home defined?

cookbooks/aws-parallelcluster-platform/recipes/config/cluster_user.rb
```diff
@@ -12,42 +12,46 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
+Chef::Log.info("config - user home before #{node['cluster']['cluster_user_home']}")
+node.override['cluster']['cluster_user_home'] = node['cluster']['cluster_user_local_home'] if node['cluster']['default_user_home'] == 'local'
```
gmarciani, line 15: Why not having this conditional logic in the attribute definition in `aws-parallelcluster-shared`, rather than making an override?

cookbooks/aws-parallelcluster-shared/attributes/users.rb
```diff
@@ -18,3 +18,19 @@
...
+elsif platform?('rocky') && node['platform_version'].to_i == 8
+  default['cluster']['cluster_user'] = 'rocky'
+elsif platform?('ubuntu')
+  default['cluster']['cluster_user'] = 'ubuntu'
+else
+  raise "The OS must be one of the following: Amazon Linux 2, Ubuntu, CentOS 7, RHEL 8, or Rocky 8"
+end
+
+default['cluster']['cluster_user_home'] = "/home/#{node['cluster']['cluster_user']}"
+default['cluster']['cluster_user_local_home'] = "/local/home/#{node['cluster']['cluster_user']}"
```
gmarciani: *Minor*
What about making `cluster_user_local_home ` depend on `cluster_user_home ` to avoid duplication of `home`?

```
default['cluster']['cluster_user_local_home'] = "/local/#{node['cluster']['cluster_user_home']}"
```

cookbooks/aws-parallelcluster-shared/attributes/users.rb
```diff
@@ -18,3 +18,19 @@
 default['cluster']['munge']['user_id'] = node['cluster']['reserved_base_uid'] + 2
 default['cluster']['munge']['group'] = node['cluster']['munge']['user']
 default['cluster']['munge']['group_id'] = node['cluster']['munge']['user_id']
+
+if (platform?('amazon') && node['platform_version'].to_i == 2) ||
```
gmarciani: Why did you choose to move the conditional logic here and remove the OS-based attributes files?

cookbooks/aws-parallelcluster-shared/attributes/users.rb
```diff
@@ -18,3 +18,19 @@
...
+   (platform?('redhat') && node['platform_version'].to_i == 8)
+  default['cluster']['cluster_user'] = 'ec2-user'
+elsif platform?('centos') && node['platform_version'].to_i == 7
+  default['cluster']['cluster_user'] = 'centos'
+elsif platform?('rocky') && node['platform_version'].to_i == 8
+  default['cluster']['cluster_user'] = 'rocky'
+elsif platform?('ubuntu')
+  default['cluster']['cluster_user'] = 'ubuntu'
+else
+  raise "The OS must be one of the following: Amazon Linux 2, Ubuntu, CentOS 7, RHEL 8, or Rocky 8"
```
gmarciani: *Minor*
What about having an constant with the supported OSes and use such enum values here?
I think it could help self-documentation and when we add new OSes we would need to only change a constant rather than going through different message strings.

cookbooks/aws-parallelcluster-environment/recipes/init.rb
gmarciani: Let's add a unit test verifying that the recipe is included in init

cookbooks/aws-parallelcluster-environment/recipes/init/move_default_user.rb
```diff
@@ -0,0 +1,37 @@
...
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Backup the cluster user's default home directory
+bash "Backup #{node['cluster']['cluster_user_home']}" do
```
gmarciani: What about covering this recipe with a unit test?

cookbooks/aws-parallelcluster-environment/recipes/init.rb
dreambeyondorange: I see your point.  The rest of the top level recipes follow this pattern, so even though it adds a layer of obfuscation, I will make it consistent.  Thanks!

cookbooks/aws-parallelcluster-shared/attributes/users.rb
dreambeyondorange: good idea!

cookbooks/aws-parallelcluster-shared/attributes/users.rb
dreambeyondorange: because the files weren't adding any value.  Several of them use the same logic and they are all just one line of code.  This way you can see the definition for every os in one file and the logic is pretty straightforward

cookbooks/aws-parallelcluster-shared/attributes/users.rb
dreambeyondorange: I agree, I will check if that already exists somewhere.  I'm sure this has come up before

cookbooks/aws-parallelcluster-environment/recipes/init/move_default_user.rb
dreambeyondorange: had to google it myself https://discourse.chef.io/t/solved-eoh-what-does-it-mean/12411

cookbooks/aws-parallelcluster-environment/recipes/init/move_default_user.rb
dreambeyondorange: It's defined in the shared attributes file and overridden in the move_default_user recipe

cookbooks/aws-parallelcluster-shared/attributes/users.rb
dreambeyondorange: Actually this doesn't work because I don't want to have a circular dependency in the recipe file.  Unless we know that it won't matter when the cluster_user_home is overridden?

cookbooks/aws-parallelcluster-platform/recipes/config/cluster_user.rb
dreambeyondorange: oh yea that's true, I have all the info I need up front

cookbooks/aws-parallelcluster-shared/attributes/users.rb
dreambeyondorange: i think just setting the cluster_user_home to what it needs to be in the attribute file makes sense and we don't need this extra variable

dreambeyondorange: i remember now that the both are used in the move process, so we need to have both

cookbooks/aws-parallelcluster-shared/attributes/users.rb
dreambeyondorange: I think the issue with constants is the order of attribute file loading.  The files are loaded in lexical order in the same cookbook, and I don't want to have dependencies between attribute files in the same cookbook and have them break when filenames change.  I would need to define the constants in this file, which doesn't make sense as the `users` attribute file

dreambeyondorange: I think we would need to refactor the attribute file naming convention to purposefully order them, which I think is out of scope for this PR

cookbooks/aws-parallelcluster-platform/recipes/config/cluster_user.rb
dreambeyondorange: actually this doesn't work because i need both values in the recipe

dreambeyondorange: I think it is also useful to have both defined in the attribute file to keep engineers aware of the two

cookbooks/aws-parallelcluster-environment/recipes/init.rb
dreambeyondorange: We don't currently have any unit tests to make sure any recipes are included in any top level cookbook

dreambeyondorange: but now that I am moving all the login into the child recipe, it shouldn't matter

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 ------
 
 **ENHANCEMENTS**
+- Add the configuration parameter `DefaultUserHome` under `HeadNode` to allow users to move the default user's home directory to `/local/home` instead of `/home` (default).
```
gmarciani: *Minor*
If you want to specify the section where the parameter is placed (good practice), we should write "configuration parameter `DeploymentSettings/DefaultUserHome`"

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
```diff
@@ -734,3 +734,15 @@ suites:
...
+  - name: default_user_local_home
+    run_list:
+      - recipe[aws-parallelcluster-tests::setup]
+      - recipe[aws-parallelcluster-environment::config_default_user_home]
+    verifier:
+      controls:
+        - local_default_user_home
+    attributes:
+      cluster:
+        node_type: 'HeadNode'
```
gmarciani, line 764: Since the setting should be applied to every cluster node, shouldn't we test also on compute and login nodes?

cookbooks/aws-parallelcluster-shared/attributes/users.rb
enrico-usai: The idea of having different files for different users is to simplify addition/removal of an os. Having files with os specific logic will isolate the code for the different OSes.

In past we had a lot of if/else in all the recipes and resources to differentiate the behaviour for the different oses, but was very hard to search for them and fix all of them when adding a new OS.
Having os specific files permits to search for <os-suffix>.rb and remove all of them when deprecating an OS, or create OS specific files if the behaviour is specific for them when adding a new OS.

If we want to reduce the number of files we could create one for "platform" (e.g. unify amazon2 and redhat2) or we can rename users_<os>.rb to a more generic name like environment_<os>.rb to be a container for all the "os" specific attributes we can have in future.

dreambeyondorange: I think that makes sense.  The other issue in this case is that we need to know the OS in order to be able to determine the cluster user and then the home directory.  Since the behavior is the same for all the OSes, we end up with the same two lines of code duplicated across 5 different files.  Ruby and Chef don't really allow the kind of hierarchy we need to order our attribute definitions cleanly.  Is there another way?  Would you prefer we have the logic 

```
default['cluster']['cluster_user_home'] = "/home/#{node['cluster']['cluster_user']}"
default['cluster']['cluster_user_local_home'] = "/local#{node['cluster']['cluster_user_home']}"
```
in each OS specific user file?

enrico-usai: One alternative (not sure if feasible) would be to use lazy evaluation for cluster_user_home and cluster_user_local_home attributes.

Another alternative, if there is "OS specific" logic, is to create a small custom-resource for it.
In the rest of the code all the OS specific logic is confined in the custom-resources, this is the only case where we have this kind of definition in the attributes.

We can create a "users" resource, in this way we can have an hierarchical definition and avoid duplication.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+return if node['cluster']['default_user_home'] == 'shared'
+
+# Backup the cluster user's default home directory
+bash "Backup #{node['cluster']['cluster_user_home']}" do
+  user 'root'
+  group 'root'
+  code <<-EOH
+    mkdir -p /tmp#{node['cluster']['cluster_user_home']}
```
gmarciani: Security best practices:
1. use randomly generated names for tmp folders.
2. if not possible add a check before the copy that fails if the destination directory exists

The above solutions mitigate an attack scenario where the attacker writes the folder with malicious files that would then be copied to the destination local home.

cookbooks/aws-parallelcluster-environment/kitchen.environment-config.yml
dreambeyondorange: We actually can't test on compute and login nodes because it depends on a shared directory, but we can add integ tests for these cases

himani2411 (approved):

PR #2642, opened 2024-02-06
[develop] Fix network route tables configuration

hanwen-pcluste:
### Description of changes
* This commit fixes a bug from https://github.com/aws/aws-parallelcluster-cookbook/pull/2632/files. This bug caused consistent creation failure with hpc6id cluster.

The logic ami at specifying better priority to primary network interface, which is the network interface with index 0.

NetworkCardIndex is the index of the physical network card
DeviceNumber is the index of the virtual network interfaces
An instance can have one or multiple network cards and every network card can have one or multiple virtual network interfaces

### Tests
* Cluster creations with c6g, hpc6id, and c5 are successful

### Checklist


cookbooks/aws-parallelcluster-environment/recipes/config/network_interfaces.rb
```diff
@@ -14,8 +14,8 @@
 
 return if on_docker?
 
-def device_number(mac, token)
-  uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/device-number")
+def network_card_index(mac, token)
+  uri = URI("http://169.254.169.254/latest/meta-data/network/interfaces/macs/#{mac}/network-card")
```
enrico-usai, line 18: Could you please add more details about the fix in the commit message?

Why do we need to use network card index rather than device number? 
Which was the issue? Which is the new behaviour?

hanwen-pcluste: Done, added more description

jdeamicis: I randomly checked the metadata from a t2.medium, and the `network-card` field is not available there.

How can we be sure that that field is available at least on all multi-NIC instance types we support?

hanwen-pcluste: The logic is only run on multi-nics instances. We have integration tests running on multiple mult-nics instances and didn't discover any issues.

jdeamicis: How many different multi-NIC instance types are we checking?

jdeamicis (approved):

PR #2645, opened 2024-02-08
[DFSM] Add head node checks and storage updates for login nodes + prevent drain/terminate when live updates are supported 

gmarciani:
### Description of changes
1. Include login nodes in cluster readiness checks done by head node before signaling update completion
2. Include storage update as part of the update of login nodes.
3. Prevent drain/terminate when live updates are supported
4. Mount internal EBS volume `/opt/parallelcluster/shared` also in login nodes as it contain files that are useful, such as cluster config, change-set and so on. 
5. Added spec tests to cover the above changes.
6. Added log lines to facilitate troubleshooting of the update workflow.

### Tests
* Manual testing: created a cluster without shared storage, updated by adding and removing external EFS in draining mode (no node replacement occur, as expected), updated by adding and removing a managed EFS in draining mode (node replacement occur as expected).
* Unit testing (both existing and new ones)
* Integ test `test_update_slurm`

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_ebs.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_ebs.rb
@@ -30,6 +30,17 @@ when 'ComputeFleet'
   end
 
 when 'LoginNode'
+  # Mount /opt/parallelcluster/shared over NFS
+  volume "mount #{node['cluster']['shared_dir']}" do
+    action :mount
+    shared_dir node['cluster']['shared_dir']
+    device(lazy { "#{node['cluster']['head_node_private_ip']}:#{node['cluster']['shared_dir_head']}" })
+    fstype 'nfs'
+    options node['cluster']['nfs']['hard_mount_options']
+    retries 10
+    retry_delay 6
+  end
+
   # Mount /opt/parallelcluster/shared_login_nodes over NFS
   volume "mount #{node['cluster']['shared_dir_login']}" do
     action :mount
--- a/cookbooks/aws-parallelcluster-environment/spec/unit/recipes/mount_internal_use_ebs_spec.rb
+++ b/cookbooks/aws-parallelcluster-environment/spec/unit/recipes/mount_internal_use_ebs_spec.rb
@@ -1,0 +1,93 @@
+# frozen_string_literal: true
+
+# Copyright:: 2024 Amazon.com, Inc. and its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'spec_helper'
+
+describe 'aws-parallelcluster-environment::mount_internal_use_ebs' do
+  SHARED_DIR = "/opt/parallelcluster/shared"
+  SHARED_DIR_HEAD = "/opt/parallelcluster/shared"
+  SHARED_DIR_COMPUTE = "/opt/parallelcluster/shared"
+  SHARED_DIR_LOGIN = "/opt/parallelcluster/shared_login"
+  HEAD_NODE_PRIVATE_IP = "0.0.0.0"
+  HARD_MOUNT_OPTIONS = "HARD_MOUNT_OPTIONS"
+
+  for_all_oses do |platform, version|
+    context "on #{platform}#{version}" do
+      %w(HeadNode ComputeFleet LoginNode).each do |node_type|
+        context "on #{node_type}" do
+          cached(:chef_run) do
+            runner = runner(platform: platform, version: version) do |node|
+              node.override['cluster']['node_type'] = node_type
+              node.override['cluster']['shared_dir'] = SHARED_DIR
+              node.override['cluster']['shared_dir_head'] = SHARED_DIR_HEAD
+              node.override['cluster']['shared_dir_compute'] = SHARED_DIR_COMPUTE
+              node.override['cluster']['shared_dir_login'] = SHARED_DIR_LOGIN
+              node.override['cluster']['head_node_private_ip'] = HEAD_NODE_PRIVATE_IP
+              node.override['cluster']['nfs']['hard_mount_options'] = HARD_MOUNT_OPTIONS
+            end
+            runner.converge(described_recipe)
+          end
+          cached(:node) { chef_run.node }
+
+          case node_type
+          when "HeadNode"
+            it 'does not mount the internal EBS volume shared with all cluster nodes' do
+              is_expected.not_to mount_volume("mount #{SHARED_DIR}")
+            end
+            it 'does not mount the internal EBS volume shared with login nodes' do
+              is_expected.not_to mount_volume("mount #{SHARED_DIR_LOGIN}")
+            end
+
+          when "ComputeFleet"
+            it 'mounts the internal EBS volume shared with all cluster nodes' do
+              is_expected.to mount_volume("mount #{SHARED_DIR}").with(
+                shared_dir: SHARED_DIR,
+                device: "#{HEAD_NODE_PRIVATE_IP}:#{SHARED_DIR_HEAD}",
+                fstype: "nfs",
+                options: HARD_MOUNT_OPTIONS,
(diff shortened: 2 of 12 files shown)


cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_ebs.rb
```diff
@@ -30,6 +30,17 @@
   end
 
 when 'LoginNode'
+  # Mount /opt/parallelcluster/shared over NFS
+  volume "mount #{node['cluster']['shared_dir']}" do
```
enrico-usai, line 34: This does not seem correct.

We should not mount the shared folder that we're using for the compute nodes. The folder to share in the login nodes is `shared_dir_login`.

cookbooks/aws-parallelcluster-platform/resources/fetch_config.rb
```diff
@@ -68,7 +73,26 @@
     # load cluster config into a node object
     load_cluster_config(node['cluster']['cluster_config_path'])
   when 'LoginNode'
-    raise "Cluster config not found in #{node['cluster']['login_cluster_config_path']}" unless ::File.exist?(node['cluster']['login_cluster_config_path'])
+    if kitchen_test?
```
enrico-usai, line 76: suggstion: it seems that the two branches for Login and Compute nodes are the same, except for some "variables", we could unify them to avoid code duplication.

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
```diff
@@ -0,0 +1,120 @@
...
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+# rubocop:disable Style/SingleArgumentDig
+
+STORAGE_TYPES_SUPPORTING_LIVE_UPDATES = %w(Efs FsxLustre FsxOntap FsxOpenZfs FileCache).freeze
```
enrico-usai, line 16: why storage.rb is in the slurm package? this should be in the environment one.
All these "file system" info should stay confined in the environment cookbook.

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_ebs.rb
enrico-usai: This is a design choice, to isolate what the login nodes can see, we should instead download the files required for the update in the shared login folder.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/mount_internal_use_ebs_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+              )
+            end
+
+            it 'does not mount the internal EBS volume shared with login nodes' do
+              is_expected.not_to mount_volume("mount #{SHARED_DIR_LOGIN}")
+            end
+
+          when "LoginNode"
+            it 'mounts the internal EBS volume shared with all cluster nodes' do
+              is_expected.to mount_volume("mount #{SHARED_DIR}").with(
```
enrico-usai, line 69: to be changed.

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_ebs.rb
gmarciani: I agree that is a design choice, but let's discuss about the correctness.
Observations:
1. the separation between shared storage for login and compute nodes have been done to prevent compute nodes from accessing secrets stored in the login nodes shared folder; not the other way around
2. the content of the shared folder is a common content without any secret (mainly config files, changeset, ...)

If we want to keep the separation, we should rather have a folder with common files shared with the whole cluster, one folder for login nodes and another for compute. Anyway, I think that it is over killing considering the non sensitive content of the comoute node shared folder

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
```diff
@@ -0,0 +1,120 @@
...
+#       "StorageType": "Efs",
+#       "EfsSettings": {
+#         "FileSystemId": "fs-123456789"
+#       },
+#     "currentValue": "-"
+#    }
+#
+# @param [List[change]] changes the list fo changes provided by a change-set.
+# @return [Boolean] true if the change supports live updates; false, otherwise.
+def storage_change_supports_live_update?(changes)
```
enrico-usai, line 97: note: it's possible to unit tests these library functions, 
see for example: `describe 'aws-parallelcluster-environment:libraries:format_directory' do`

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -93,15 +93,25 @@ def get_all_queues(config)
   end
 
   def get_queues_with_changes(config)
```
enrico-usai, line 95: same here, we can unit test this function mocking external logic.

cookbooks/aws-parallelcluster-platform/resources/fetch_config.rb
gmarciani: Good point, I agree with you. Nevertheless, I would like to isolate this type of refactoring in a separate PR and keep here only the strictly necessary changes

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
enrico-usai, line 16: Probably it's worth to add in the CLI a comment to point to this constant in the cookbook, because we're defining something similar in the CLI.

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
gmarciani: Good point, thanks. I assumed it was not. I will include it in the next iteration.

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
gmarciani: Good point, thanks. I assumed it was not. I will include it in the next iteration.

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
gmarciani: Sure, thanks

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/mount_internal_use_ebs_spec.rb
gmarciani: To be changed according to the decision here https://github.com/aws/aws-parallelcluster-cookbook/pull/2645#discussion_r1484255301

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
gmarciani: it's in the SLURM cookbook because it's there that it is used, that is where the decision about replacing or not nodes is taken.

enrico-usai: what about calling a recipe that is in environment cookbook and move the library there if possible?

enrico-usai (approved):

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_ebs.rb
gmarciani: Confirmed with login nodes feature owners. The reason why the shared storage was separated in one folder for login nodes and another one for compute nodes was to prevent compute nodes from reading/writing secrets in lgon nodes shared storage. Login nodes can access compute shared storage in any case by submitting a job that reads/writes it.

Conclusions: it's correct to mount the common shared storage (/opt/parallelcluster/shared) to both compute and login nodes.

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
gmarciani: Improvement for next iteration: move storage.rb to environment cookbook and make slurm coobook dependant on environment's.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/mount_internal_use_ebs_spec.rb
gmarciani: no need to change

PR #2655, opened 2024-02-19
[BugFix] Fix issue making job fail when submitted as active directory user from login nodes.

gmarciani:
### Description of changes
Fix issue making job fail when submitted as active directory user from login nodes.

#### Explanation
Before this change, the user used to see the below consistent error when submitting jobs as active directory use rform login nodes:

```
[user1@ip-3-5-4-151 ~]$ srun -n 1 -N 1 hostname
srun: error: Unable to create step for job 1: Error generating job credential
```

To unblock the submission from login nodes, the user needed to trigger the fetch of active directory data from within the head node, by switching to an AD user. We were aware of a similar limitation and that's why we introduced [this code](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-environment/recipes/config/directory_service.rb#L237-L245) in the cookbook. We figured out two limitations on that approach:
1. it was always failing because at the time of its execution the AD data was not ready to be fetched, but the failure was ignored;
2. it is not enough to execute a getent passwd to trigger the fetching of Ad data, but it's also necessary to do it within a switch-user.

So, we moved that step from config time to finalize time with retries (Ad data is ready to be fetched at that point), and we wrapped the getent command within a switch user.

### Tests
* Create cluster integrated with external MsAD with LDAPS. Submitted a job as AD user as very first action on the cluster and it succeeded. Verified also that the newly introduced step in the finalize was successfully executed.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   - web_viewer: `2023.1.16388-1`
 
 **BUG FIXES**
+- Fix issue making job fail when submitted as active directory user from login nodes. 
+  The issue was caused by an incomplete configuration of the integration with the external Active Directory on the head node.
+  This fix comes with a breaking change: now cluster creation/update would fail if the integration with the Active Directory does not work.
 
 3.8.0
 ------
--- a/cookbooks/aws-parallelcluster-environment/recipes/finalize/finalize_directory_service.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/finalize/finalize_directory_service.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster
+# Recipe:: finalize_directory_service
+#
+# Copyright:: 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+return if node['cluster']["directory_service"]["enabled"] == 'false'
+
+if %w(HeadNode LoginNode).include? node['cluster']['node_type']
+  default_user = node['cluster']['cluster_user']
+  read_only_user = domain_service_read_only_user_name(node['cluster']['directory_service']['domain_read_only_user'])
+
+  execute 'Fetch user data from remote directory service' do
+    # The switch-user (sudo -u) is necessary to trigger the fetching of AD data
+    command "sudo -u #{default_user} getent passwd #{read_only_user}"
+    user 'root'
+    retries 10 # Retries are just a safe guard in case the node is still fetching data from the AD
+    retry_delay 3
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   - web_viewer: `2023.1.16388-1`
 
 **BUG FIXES**
+- Fix issue making job fail when submitted as active directory user from login nodes.
```
jdeamicis: Nitpick: should we expand on this with a sub-bullet, saying that this was due to an incomplete configuration of AD on the head node?

cookbooks/aws-parallelcluster-environment/recipes/finalize/finalize_directory_service.rb
```diff
@@ -0,0 +1,31 @@
...
+  default_user = node['cluster']['cluster_user']
+  read_only_user = domain_service_read_only_user_name(node['cluster']['directory_service']['domain_read_only_user'])
+
+  execute 'Fetch user data from remote directory service' do
+    # The switch-user (sudo -u) is necessary to trigger the fetching of AD data
+    command "sudo -u #{default_user} getent passwd #{read_only_user}"
+    user 'root'
+    retries 10 # Retries are just a safe guard in case the node is still fetching data from the AD
+    retry_delay 3
+  end
```
jdeamicis, line 30: Are we not ignoring failures any more?

cookbooks/aws-parallelcluster-environment/recipes/finalize/finalize_directory_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+return if node['cluster']["directory_service"]["enabled"] == 'false'
+
+if %w(HeadNode LoginNode).include? node['cluster']['node_type']
+  default_user = node['cluster']['cluster_user']
+  read_only_user = domain_service_read_only_user_name(node['cluster']['directory_service']['domain_read_only_user'])
+
+  execute 'Fetch user data from remote directory service' do
+    # The switch-user (sudo -u) is necessary to trigger the fetching of AD data
+    command "sudo -u #{default_user} getent passwd #{read_only_user}"
```
jdeamicis, line 26: Why do you need to run it as the default user?

Then, if this is really needed, just use the `user "#{default_user}#` parameter below (instead of using `root`).

jdeamicis (commented):
Left some questions

cookbooks/aws-parallelcluster-environment/recipes/finalize/finalize_directory_service.rb
gmarciani: No, if it fails, it fails the step. 

Now I'm thinking that this is a breaking change that should be highlighted in the changelog. This will be the first release where we will fail the integration does not work.

cookbooks/aws-parallelcluster-environment/recipes/finalize/finalize_directory_service.rb
gmarciani: The getent alone is not enough to trigger the fetch of AD data. The switch user is required. I'm using the default user because it is the one that for sure is always deployed on the cluster.

gmarciani: Using the `ùser` attribute of the chef resource would be equivalent and does not come with any more advantage. On the other hand a more explicit sudo seems more intentional.

CHANGELOG.md
gmarciani: Makes sense. Fixed in next rev.

jdeamicis (approved):

PR #2656, opened 2024-02-19
[Security] Disable unused background services: wpa_supplicant and cups.

gmarciani:
### Description of changes
Disable unused background services: `wpa_supplicant` and `cups`.

### Test
* Spec tests for aws-parallelcluster-platform
* Kitchen tests `./kitchen.ec2.sh platform-install test disable-services`

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/recipes/install/disable_services.rb
+++ b/cookbooks/aws-parallelcluster-platform/recipes/install/disable_services.rb
@@ -27,3 +27,13 @@ end unless on_docker?
 service 'log4j-cve-2021-44228-hotpatch' do
   action %i(disable stop mask)
 end unless on_docker?
+
+# Necessary on Ubuntu and Amazon Linux 2
+service 'cups' do
+  action %i(disable stop mask)
+end unless on_docker?
+
+# Necessary on Ubuntu 22
+service 'wpa_supplicant' do
+  action %i(disable stop mask)
+end unless on_docker?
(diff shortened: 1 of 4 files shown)


cookbooks/aws-parallelcluster-platform/recipes/install/disable_services.rb
```diff
@@ -27,3 +27,13 @@
 service 'log4j-cve-2021-44228-hotpatch' do
   action %i(disable stop mask)
 end unless on_docker?
+
+# Necessary on Ubuntu and Amazon Linux 2
+service 'cups' do
```
enrico-usai, line 32: This recipe is not following the general code structure.

It's not strictly related to your change, but when there is code "os specific" we should create a Chef Resource for it.

The idea is to have in the recipes only common code, the recipe should be the orchestrator that will call the resource to perform OS specific actions.

In this way if we want to fix/test/search code for a specific OS we don't need to search in all the recipes but we can rely on the resources.
Then if we want to add/remove support for a specific OS we just need to update the resources avoiding to touch the recipes.


In this case we can create a resource called "service_manager" or something like this and call it directly in in the install recipe, with a different flavour for alinux2, ubuntu, u22, etc.

gmarciani: I agree. Refactoring captured in our backlog. We will decouple the refactoring from this change though.

gmarciani:
Closed in favour of https://github.com/aws/aws-parallelcluster-cookbook/pull/2775

gmarciani:
Closed in favour of https://github.com/aws/aws-parallelcluster-cookbook/pull/2775

PR #2661, opened 2024-02-21
[DFSM] Fix major gaps in the update workflow

gmarciani:
### Description of changes
This change addresses the major gaps in the update workflows. In particular:
1. If the change-set does not require a live update on compute and login nodes, they save a DDB record with status SKIPPED_LIVE_UPDATE and exists from the update without performing and storage change.
2. If the change-set requires a live update on compute and login nodes, they perform the storage change and save a DDB record with status DEPLOYED.
3. The cluster readiness check performed by the head node considers successful all nodes having either status DEPLOYED or SKIPPED_LIVE_UPDATE.
4. Cluster readiness check now emits a more readable log about retrieved DDB items.

### Tests
* Manual tests: 
  * created a cluster without shared storage
  * updated it with external storage supporting live updates
  * updated it with external storage supporting live updates and adding more static nodes (to increase noise)
  * updated it with shared storage not supporting live updates
  * updated with a mix of shared storage changes that support and des not support live updates
  * updated shared storage with a change also in static fleet capacity. 
* Integ tests: `test_update_slurm` (with the fixes to the test code made in https://github.com/aws/aws-parallelcluster/pull/6110)
* Integ test `test_update_instance_list`

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
+++ b/cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
@@ -14,13 +14,24 @@
 # Saves on the DynamoDB table 'parallelcluster-$CLUSTERNAME' the correspondence between
 # the instance id and the cluster config version deployed on that instance.
 # This is done to understand what is the deployed cluster config in each node, for example, during a CFN Stack update.
-def save_instance_config_version_to_dynamodb
+
+
+DDB_CONFIG_STATUS = {
+  DEPLOYED: "DEPLOYED",
+  SKIPPED_LIVE_UPDATE: "SKIPPED_LIVE_UPDATE"
+}
+
+def save_instance_config_version_to_dynamodb(status)
   unless on_docker? || kitchen_test? && !node['interact_with_ddb']
     # TODO: the table name should be read from a dedicated node attribute,
     #   but it is currently missing in the dna.json for compute nodes.
     table_name = "parallelcluster-#{node['cluster']['cluster_name']}"
     item_id = "CLUSTER_CONFIG.#{node['ec2']['instance_id']}"
-    item_data = "{\"cluster_config_version\": {\"S\": \"#{node['cluster']['cluster_config_version']}\"}, \"lastUpdateTime\": {\"S\": \"#{Time.now.utc}\"}}"
+    node_type = node['cluster']['node_type']
+    config_version = node['cluster']['cluster_config_version']
+    last_update_time = Time.now.utc
+
+    item_data = "{\"cluster_config_version\": {\"S\": \"#{config_version}\"}, \"status\": {\"S\": \"#{status}\"}, \"node_type\": {\"S\": \"#{node_type}\"}, \"lastUpdateTime\": {\"S\": \"#{last_update_time}\"}}"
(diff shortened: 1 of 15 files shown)


cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
```diff
@@ -14,13 +14,23 @@
 # Saves on the DynamoDB table 'parallelcluster-$CLUSTERNAME' the correspondence between
 # the instance id and the cluster config version deployed on that instance.
 # This is done to understand what is the deployed cluster config in each node, for example, during a CFN Stack update.
-def save_instance_config_version_to_dynamodb
+
+DDB_CONFIG_STATUS = {
```
enrico-usai, line 18: suggestion: it's worth to add a description of these two status

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
```diff
@@ -104,7 +104,7 @@ def storage_change_supports_live_update?(changes)
 
   if storage_changes.empty?
     Chef::Log.info("No shared storage change found: assuming live update is supported")
-    return true
+    return false
```
enrico-usai: Why? I'd think if there are no changes live update is supported, as mentioned in the log message.

enrico-usai (dismissed):

cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
gmarciani: I agree, will do

cookbooks/aws-parallelcluster-slurm/libraries/storage.rb
gmarciani: You right. I needed to change it because this function is used in two places: in one i needed this check to return false, in another one it does not matter what it returns because it is called only if the changeset contains shared storage changes.

So your comment highlighted the need to revise this logic 
Addressing.

gmarciani: But in any case, I need to adapt the log line, addressing...

gmarciani: Addressed

cookbooks/aws-parallelcluster-slurm/libraries/dynamo.rb
gmarciani: Addresses

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
```diff
@@ -95,3 +95,43 @@ def is_login_nodes_removed?
   previous_config = YAML.safe_load(File.read(node['cluster']['previous_cluster_config_path']))
   previous_config.dig("LoginNodes") and !config.dig("LoginNodes")
 end
+
+def is_live_update_required?
```
enrico-usai, line 99: in future we can add unit tests for this specific function.

enrico-usai (approved):

cookbooks/aws-parallelcluster-slurm/libraries/update.rb
gmarciani: Yes, tracked in non blocking gaps to address

PR #2662, opened 2024-02-22
[CfnHup] Define cfn-hup configuration files within the environment config recipe.

gmarciani:
**This PR depends on https://github.com/aws/aws-parallelcluster/pull/6112**

### Description of changes
Define cfn-hup configuration files within the environment config recipe.
These files used to be defined within the nodes user data, but we decided to move them to the cookbook in order to resize the CloudFormation template.

### Tests
* Spec tests
* Manually created a cluster with compute nodes and login nodes, then updated it by adding new storage => the update is triggered as expected.
* integ test `test_update.py::test_update_slurm` to verify that we did not break the update workflow.
* integ test `test_cluster_custom_resource.py::test_cluster_create` to verify that we reduce the template size enough to make this test succeed (it used to fail due to template size limit exceeded)

### Checklist


cookbooks/aws-parallelcluster-environment/templates/cfn_bootstrap/cfn-hup.conf.erb
```diff
@@ -0,0 +1,5 @@
+[main]
+stack=<%= @stack_id %>
+region=<%= @region %>
+url=<%= @cloudformation_url %>
+role=<%= @cfn_init_role %>
```
hanwen-pcluste, line 5: Why adding this new parameter "role"?

cookbooks/aws-parallelcluster-environment/templates/cfn_bootstrap/cfn-hook-update.conf.erb
```diff
@@ -0,0 +1,5 @@
+[parallelcluster-update]
+triggers=post.update
+path=Resources.<%= @launch_template_resource_id %>.Metadata.AWS::CloudFormation::Init
+action=PATH=/usr/local/bin:/bin:/usr/bin:/opt/aws/bin; . /etc/profile.d/pcluster.sh; cfn-init -v --stack <%= @stack_id %> --resource <%= @launch_template_resource_id %> --configsets update --region <%= @region %> --url <%= @cloudformation_url %> --role <%= @cfn_init_role %>
```
hanwen-pcluste, line 4: Why adding this new parameter "role"?

cookbooks/aws-parallelcluster-environment/templates/cfn_bootstrap/cfn-hup.conf.erb
gmarciani: Very good question :-)= 
This is required by cfn-hup when you configure it to monitor a nested stack. 
Compute fleet and login nodes are defined in nested stacks so it's required for me.
It is optional for the head node, instead, but we prefer to have a single definition that works for all.

We added this role as part of the DFSMv2 feature, but we are not introducing it in this PR specifically. In fact, it was already defined [here](https://github.com/aws/aws-parallelcluster/blob/develop/cli/src/pcluster/resources/compute_node/user_data.sh#L62)

cookbooks/aws-parallelcluster-environment/templates/cfn_bootstrap/cfn-hook-update.conf.erb
gmarciani: Same as above

hanwen-pcluste (approved):

PR #2634, opened 2024-02-01
[DFSM] Make head node check that all compute nodes have deployed the expected cluster config version, on create and update.

gmarciani:
### Description of changes
Make head node check that all compute nodes have deployed the expected cluster config version, on create and update.

This PR depends on this other PR, where the required extra permissions for the head node are added.

### Tests
* Unit tests for the newly introduced utility `check_cluster_ready.py`.
* New and existing spec tests
* Manually created, updated (forced) mounting/unmounting EFS, deleted a cluster.
* Verified that the sequence of events is as expected: stack creation/update waits for the head node to verify cluster config version on DDB from compute nodes before proceeding.
* Integ test `test_update_slurm`.

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2673, opened 2024-03-04
Get instances from DynamoDB in batches of 100

hanwen-pcluste:
DynamoDB batch_get_item API limit the number of input to 100. This change fixes cluster creation with >100 compute nodes

### Tests
Manually tested cluster creation and update with 1000 compute nodes

### References
* This PR fixes a bug from https://github.com/aws/aws-parallelcluster-cookbook/pull/2634

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/files/default/head_node_checks/common/constants.py
+++ b/cookbooks/aws-parallelcluster-slurm/files/default/head_node_checks/common/constants.py
@@ -12,7 +12,7 @@ from botocore.config import Config
 
 # BOTO
 BOTO_CONFIG = Config(retries={"max_attempts": 60})
-BOTO_PAGINATION_CONFIG = {"PageSize": 500}
+BOTO_PAGINATION_CONFIG = {"PageSize": 100}
 
 # TAGS
 CLUSTER_NAME_TAG = "parallelcluster:cluster-name"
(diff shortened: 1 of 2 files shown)


himani2411 (approved):

cookbooks/aws-parallelcluster-slurm/files/default/head_node_checks/common/constants.py
```diff
@@ -12,7 +12,7 @@
 
 # BOTO
 BOTO_CONFIG = Config(retries={"max_attempts": 60})
-BOTO_PAGINATION_CONFIG = {"PageSize": 500}
+BOTO_PAGINATION_CONFIG = {"PageSize": 100}
```
gmarciani, line 15: To address scaling issues caused by DFSMv2, we are decreasing the pagination size from 500 to 100 items, making both EC2 calls and DDB calls retrieve at most 100 items per request. This is fine but can be improved post release. In fact, DDB request BatchGetItem supports at most 100 elements, but the EC2 DescribeInstances can support 500. So with this change we increase by 5x the number of EC2 requests.

We want to improve this approach by making the EC2 DescribeInstances retrieve 500 items per request and execute the DDB request in batch of 100 items each. In this way we would take the most out of each request, minimizing the number of request and the consequent risk of throttling.

gmarciani (commented):
To address scaling issues caused by DFSMv2, we are decreasing the pagination size from 500 to 100 items, making both EC2 calls and DDB calls retrieve at most 100 items per request. This is fine but can be improved post release. In fact, DDB request BatchGetItem supports at most 100 elements, but the EC2 DescribeInstances can support 500. So with this change we increase by 5x the number of EC2 requests.

We want to improve this approach by making the EC2 DescribeInstances retrieve 500 items per request and execute the DDB request in batch of 100 items each. In this way we would take the most out of each request, minimizing the number of request and the consequent risk of throttling.

PR #2691, opened 2024-04-05
Force a string `true` in the update_shared_storages to match the expectations of  _mount_unmount

dreambeyondorange:
… in the efs resource

Updates to a cluster result in boolean values for the iam_authorization and efs_encryption_in_transit objects.

### Tests
* Tested the change locally with an update to a cluster with an EFS fs with encryption in transit.  Adding the boolean check allows the cookbook to use TLS for the fs.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/recipes/config/update_shared_storages.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/config/update_shared_storages.rb
@@ -83,8 +83,9 @@ ruby_block "get storage to mount and unmount" do
         next unless not_in_shared_storages_mapping["efs"].nil? || !not_in_shared_storages_mapping["efs"].include?(storage)
         shared_dir_array.push(storage["mount_dir"])
         efs_fs_id_array.push(storage["efs_fs_id"])
-        efs_encryption_in_transit_array.push(storage["efs_encryption_in_transit"])
-        efs_iam_authorization_array.push(storage["efs_iam_authorization"])
+        # The EFS resource expects strings for these attributes, not booleans
+        efs_encryption_in_transit_array.push(String(storage["efs_encryption_in_transit"]))
+        efs_iam_authorization_array.push(String(storage["efs_iam_authorization"]))
       end
     end
     [shared_dir_array, efs_fs_id_array, efs_encryption_in_transit_array, efs_iam_authorization_array]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@ aws-parallelcluster-cookbook CHANGELOG
 
 This file is used to list changes made in each version of the AWS ParallelCluster cookbook.
 
+3.10.0
+------
+
+**ENHANCEMENTS**
+
+**CHANGES**
+
+**BUG FIXES**
+- Fixed an issue that prevented cluster updates from including EFS filesystems with encryption in transit.
+
 3.9.0
 ------
 


cookbooks/aws-parallelcluster-environment/resources/efs/partial/_mount_umount.rb
```diff
@@ -40,9 +40,9 @@
 
     # See reference of mount options: https://docs.aws.amazon.com/efs/latest/ug/automount-with-efs-mount-helper.html
     mount_options = "_netdev,noresvport"
-    if efs_encryption_in_transit == "true"
+    if efs_encryption_in_transit == "true" || efs_encryption_in_transit == true
```
gmarciani: We introduced encryption and iam authz in 3.4.0. Does this bug means that it has never worked?
What determines if the variable uses the string or the boolean?
Instead of making a double check string&bool I suggest two alternatives:
1. (enough) convert the variable to boolean and check only for the boolean value
2. (the best) fix the code at the source of the issue, so that the value can only be a string or a boolean, and check just for one of them.

Also, I imagine we may have this type of situation also in other places. I suggest to create a task to check cookbook wide to prevent future issues.

dreambeyondorange: I think the problem is just in update, since the dna.json doesn't have booleans, just strings and the update generates the data.  It seems adding an EFS fs with encryption in transit wasn't tested, and is an edge case since no one has reported an issue until now.

dreambeyondorange: Either converting a string to a boolean or checking for the string and the boolean seems the same to me.  Why is the former better than the latter?  Actually I can answer this myself ahaha, but the code generally doesn't allow us to force typing, and that is how we end up in these situations.  We can convert the boolean that is generated from the yaml loading to a string.  Unless we document these types of patterns, this will trip up engineers in the future as well.  Since how would they know what the variable type should be?

dreambeyondorange: But considering we use the data from dna.json throughout the codebase, we should force these to be strings to match that pattern

cookbooks/aws-parallelcluster-environment/recipes/config/update_shared_storages.rb
```diff
@@ -83,8 +83,17 @@ def get_efs(action)
         next unless not_in_shared_storages_mapping["efs"].nil? || !not_in_shared_storages_mapping["efs"].include?(storage)
         shared_dir_array.push(storage["mount_dir"])
         efs_fs_id_array.push(storage["efs_fs_id"])
-        efs_encryption_in_transit_array.push(storage["efs_encryption_in_transit"])
-        efs_iam_authorization_array.push(storage["efs_iam_authorization"])
+        # The EFS resource expects strings for these attributes, not booleans
+        if storage["efs_encryption_in_transit"]
```
gmarciani: What about avoiding if..else block as follows?

```
efs_encryption_in_transit_array.push(String(storage["efs_encryption_in_transit"]))
```

dreambeyondorange: oh thanks this is great!

dreambeyondorange: Although I'm not sure that the object is guaranteed to exist and String cannot handle empty input

gmarciani (approved):

PR #2727, opened 2024-05-20
[Develop] Add data integrity checks before removing backup directories

hehe7318:
### Description of changes
* Updated `restore_home_shared_data.rb` to include a diff check before removing /tmp/home/ directory.
  * Ensures data in the new home directory is the same as the original before deleting the temporary backup.
* Updated `restore_internal_use_shared_data.rb` to include a diff check before removing /tmp directories for each internal shared directory.
  * Ensures data in the new directory is the same as the original before deleting the temporary backup.
* Updated `config_default_user_home.rb` to include data integrity checks during the process of moving the cluster user's default home directory.
* Added detailed comments to explain the new data integrity checks.
* Added new Spec tests.

* This update prevents potential data loss or inconsistency during the restoration process.

### Tests
* Added Spec tests
* Spec tests passed successfully
* Kitchen tests passed successfully
* Manual Test for changes in `config_default_user_home.rb`:
  * Conducted manual tests to verify the new integrity check functionality.
  * Happy Path: Manually run `config_default_user_home` recipe. Ensured the home directory is moved correctly when no data integrity issues are present.
  * Unhappy Path: Simulated a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']` and modifying files in the expected directory to differ from the original. Manually run `config_default_user_home` recipe. Verified that the script fails and outputs the correct error message in chef-client.log.
  * Results:
    * Happy path: Passed.
    * Unhappy path: Passed, with expected failure and error message.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -50,22 +50,31 @@ end
 # 2. Copies the data from the temporary backup directory (/tmp/cluster_user_home) to the new local home directory.
 # 3. Updates the cluster user's home directory path to the new local home directory.
 # 4. Changes the ownership of the new local home directory to the cluster user.
-# 5. Verifies data integrity by comparing the temporary backup directory and the new local home directory.
-# 6. If the data integrity check passes, it removes both the temporary backup directory and the original home directory.
-# 7. If the data integrity check fails, it outputs an error message, restart sshd service and exits with an error code.
-# 8. Unhappy Path Manually test passed successfully:
-#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
-#    Modify files in the expected directory to differ from the original.
-#    Manually run `config_default_user_home` recipe.
-#    Verified that the script fails and outputs the correct error message in chef-client.log.
 bash "Move #{node['cluster']['cluster_user_home']}" do
   user 'root'
   group 'root'
   code <<-EOH
+    set -e
     mkdir -p #{node['cluster']['cluster_user_local_home']}
     rsync -a /tmp#{node['cluster']['cluster_user_home']}/ #{node['cluster']['cluster_user_local_home']}
     usermod -d #{node['cluster']['cluster_user_local_home']} #{node['cluster']['cluster_user']}
     chown -R #{node['cluster']['cluster_user']}: #{node['cluster']['cluster_user_local_home']}
+  EOH
+end
+
+# Data integrity check
+# 1. Verifies data integrity by comparing the temporary backup directory and the new local home directory.
+# 2. If the data integrity check passes, it removes both the temporary backup directory and the original home directory.
+# 3. If the data integrity check fails, it outputs an error message and exits with an error code 1.
+# 4. Unhappy Path Manually test passed successfully:
+#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
+#    Modify files in the expected directory to differ from the original.
+#    Manually run `config_default_user_home` recipe.
+#    Verified that the script fails and outputs the correct error message in chef-client.log.
+bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
+  user 'root'
+  group 'root'
+  code <<-EOH
     diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
     diff_exit_code=$?
     if [ $diff_exit_code -eq 0 ]; then
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -76,6 +76,7 @@ bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
       rm -rf #{node['cluster']['cluster_user_home']}
     else
       echo "Data integrity check failed comparing #{node['cluster']['cluster_user_local_home']} and #{node['cluster']['cluster_user_home']}: $diff_output" >&2
+      systemctl start sshd
       exit 1
     fi
   EOH
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -62,18 +62,17 @@ bash "Move #{node['cluster']['cluster_user_home']}" do
   user 'root'
   group 'root'
   code <<-EOH
-    set -e
     mkdir -p #{node['cluster']['cluster_user_local_home']}
     rsync -a /tmp#{node['cluster']['cluster_user_home']}/ #{node['cluster']['cluster_user_local_home']}
     usermod -d #{node['cluster']['cluster_user_local_home']} #{node['cluster']['cluster_user']}
     chown -R #{node['cluster']['cluster_user']}: #{node['cluster']['cluster_user_local_home']}
     diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
-    if [ $? -eq 0 ]; then
+    diff_exit_code=$?
+    if [ $diff_exit_code -eq 0 ]; then
       rm -rf /tmp#{node['cluster']['cluster_user_home']}
       rm -rf #{node['cluster']['cluster_user_home']}
     else
       echo "Data integrity check failed comparing #{node['cluster']['cluster_user_local_home']} and #{node['cluster']['cluster_user_home']}: $diff_output" >&2
-      systemctl start sshd
       exit 1
     fi
   EOH
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -76,8 +76,7 @@ bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
   group 'root'
   code <<-EOH
     diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
-    diff_exit_code=$?
-    if [ $diff_exit_code -eq 0 ]; then
+    if [ $? -eq 0 ]; then
       rm -rf /tmp#{node['cluster']['cluster_user_home']}
       rm -rf #{node['cluster']['cluster_user_home']}
     else
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -52,7 +52,12 @@ end
 # 4. Changes the ownership of the new local home directory to the cluster user.
 # 5. Verifies data integrity by comparing the temporary backup directory and the new local home directory.
 # 6. If the data integrity check passes, it removes both the temporary backup directory and the original home directory.
-# 7. If the data integrity check fails, it outputs an error message and exits with an error code.
+# 7. If the data integrity check fails, it outputs an error message, restart sshd service and exits with an error code.
+# 8. Unhappy Path Manually test passed successfully:
+#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
+#    Modify files in the expected directory to differ from the original.
+#    Manually run `config_default_user_home` recipe.
+#    Verified that the script fails and outputs the correct error message in chef-client.log.
 bash "Move #{node['cluster']['cluster_user_home']}" do
   user 'root'
   group 'root'
@@ -67,7 +72,8 @@ bash "Move #{node['cluster']['cluster_user_home']}" do
       rm -rf /tmp#{node['cluster']['cluster_user_home']}
       rm -rf #{node['cluster']['cluster_user_home']}
     else
-      echo "Data integrity check failed comparing #{node['cluster']['cluster_user_home']} and /tmp#{node['cluster']['cluster_user_home']}: $diff_output"
+      echo "Data integrity check failed comparing #{node['cluster']['cluster_user_home']} and /tmp#{node['cluster']['cluster_user_home']}: $diff_output" >&2
+      systemctl start sshd
       exit 1
     fi
   EOH
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -62,15 +62,10 @@ bash "Move #{node['cluster']['cluster_user_home']}" do
   EOH
 end
 
-# Data integrity check
+# Data integrity check and cleanup for temporary backup and original home directory
 # 1. Verifies data integrity by comparing the temporary backup directory and the new local home directory.
 # 2. If the data integrity check passes, it removes both the temporary backup directory and the original home directory.
 # 3. If the data integrity check fails, it outputs an error message and exits with an error code 1.
-# 4. Unhappy Path Manually test passed successfully:
-#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
-#    Modify files in the expected directory to differ from the original.
-#    Manually run `config_default_user_home` recipe.
-#    Verified that the script fails and outputs the correct error message in chef-client.log.
 bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
   user 'root'
   group 'root'
--- a/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
@@ -44,7 +44,15 @@ bash "Backup #{node['cluster']['cluster_user_home']}" do
   EOH
 end
 
-# move the cluster user's default home directory
+# Move the cluster user's default home directory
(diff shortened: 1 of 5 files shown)


cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -19,13 +19,22 @@
...
     user 'root'
     group 'root'
     code <<-EOH
       rsync -a --ignore-existing /tmp/home/ /home
-      rm -rf /tmp/home/
+      diff -r /tmp/home/ /home
+      if [ $? -eq 0 ]; then
+        rm -rf /tmp/home/
+      else
+        echo "Data integrity check failed for /home"
```
gmarciani: It may be helpful to have an error message that mentions both the directories and the detected diff.
For example: "Data integrity check failed comparing /home and /tmp/home: $diff"

cookbooks/aws-parallelcluster-environment/recipes/init/restore_internal_use_shared_data.rb
```diff
@@ -19,14 +19,23 @@
...
       user 'root'
       group 'root'
       code <<-EOH
         rsync -a --ignore-existing /tmp#{dir}/ #{dir}
-        rm -rf /tmp#{dir}/
+        diff -r /tmp#{dir}/ #{dir}
+        if [ $? -eq 0 ]; then
+          rm -rf /tmp#{dir}/
+        else
+          echo "Data integrity check failed for #{dir}"
```
gmarciani: Same as above

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -19,13 +19,22 @@
   # This is necessary to preserve any data in these directories that was
   # generated during the build of ParallelCluster AMIs after converting to
   # shared storage and backed up to a temporary location previously
-  # Remove the backup after the copy is done
+  # Before removing the backup, ensure the data in the new home is the same
+  # as the original to avoid any data loss or inconsistency. This is done
+  # by using rsync to copy the data and diff to check for differences.
+  # Remove the backup after the copy is done and the data integrity is verified.
   bash "Restore /home" do
```
gmarciani, line 26: It's important to capture this change in a test. 
It may be enough to verify in a spec test that the body of the script contains the expected logic.

gmarciani (changes requested):

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
dreambeyondorange: I think we need to test the error path as well, to make sure it works as expected

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Great suggestion! Done!

cookbooks/aws-parallelcluster-environment/recipes/init/restore_internal_use_shared_data.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: How to test the error path? Do we want to mock the error? I didn't see Ryan's comment until just now. So for now it has the spec test that the body of the script contains the expected logic. But only for `config_default_user_home.rb` and `restore_internal_use_shared_data.rb`, because I can not find a existing spec test for `restore_home_shared_data.rb` or `mount_home`. Do you know where is it?

dreambeyondorange: They may not exist.

dreambeyondorange: You'll definitely need to mock the error, but I'm not sure its possible with everything combined into one bash command.  You might not be able to add the error path to the unit tests, but I would say you should at least test it manually, if you haven't already.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
```diff
@@ -22,7 +22,31 @@
...
           is_expected.to start_service("sshd")
         end
+
+        it 'moves the cluster user home directory with data integrity check' do
+          user_home = "/home/user"
+          user_local_home = "/local/home/user"
+          cluster_user = chef_run.node['cluster']['cluster_user']
+          expect(chef_run).to run_bash("Move #{user_home}").with(
+            code: <<-CODE
+    set -e
```
gmarciani: Minor: is this indentation required? wCan't we place the code with the indentation aligned to `code:`?

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
```diff
@@ -22,7 +22,31 @@
           expect(chef_run.node['cluster']['cluster_user_home']).to eq('/local/home/user')
           is_expected.to start_service("sshd")
         end
+
+        it 'moves the cluster user home directory with data integrity check' do
+          user_home = "/home/user"
+          user_local_home = "/local/home/user"
```
gmarciani, line 28: *Minor on CodeStyle / Knowledge Sharing*
It's a good practice in testing code to use values conveying the assumptions you have on such variables.
This is a good practice to leverage test code as a complementary form of documentation (not an absolute one of corse).
https://softwareengineering.stackexchange.com/questions/154615/are-unit-tests-really-used-as-documentation

Example:
`user_local_home` can be whatever absolute path in the system. 
If you set it to `/local/home/user` it  seems that there is a requirement for that directory to be within `/local`, since it is a notable folder in UNIX, but this is not the case.
If you set it to `/a/local/user/home` or `/whatever/local/user/home` or `/WHATEVER_DIR`, you convey the right assumptions: it can be whatever absolute path.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
gmarciani: Please, mention the manual test of the unhappy case in the description, when done.
PR LGTM now, will approve after the manual test.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
hehe7318: Unfortunately it can't. Or the test will fail. But we can do that use Regular Expression. But it needs time and is not necessary, and will make the code hard to read.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
hehe7318: Thank you for your sharing. I will remember. 
But for this case, cause we already have these codes:
```
          runner = runner(platform: platform, version: version) do |node|
            node.override['cluster']['default_user_home'] = "local"
            node.override['cluster']['cluster_user_home'] = "/home/user"
            node.override['cluster']['cluster_user_local_home'] = "/local/home/user"
          end
```
So it need to be `/local/home/user`

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Done

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
gmarciani: np, let's leave it as it is.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
gmarciani: I see, let's keep it as it is as it's not worth it to change parts of the code that are out of the PR scope.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -54,8 +67,15 @@
...
     chown -R #{node['cluster']['cluster_user']}: #{node['cluster']['cluster_user_local_home']}
-    rm -rf /tmp#{node['cluster']['cluster_user_home']}
-    rm -rf #{node['cluster']['cluster_user_home']}
+    diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
+    if [ $? -eq 0 ]; then
+      rm -rf /tmp#{node['cluster']['cluster_user_home']}
+      rm -rf #{node['cluster']['cluster_user_home']}
+    else
+      echo "Data integrity check failed comparing #{node['cluster']['cluster_user_local_home']} and #{node['cluster']['cluster_user_home']}: $diff_output" >&2
+      sudo systemctl start sshd
```
gmarciani: Why do we need to restart sshd?

gmarciani: Got it. As part of default user home setting we stop sshd to prevent unexpected access during the home configuration and we do not restart it in case of failures.

https://github.com/aws/aws-parallelcluster-cookbook/blob/d85a09c20b06ef393ef8027c03cd1f72026e8e4c/cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb#L17-L29

hehe7318: Done. Removed it. Also for preventing the `set -e` ends the script at diff_output=$(diff -r #{user_home} #{user_local_home}) so that the script will not reach the if else part, made the data integrity check a separate script without `set -e`.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -44,18 +44,37 @@
...
+# 8. Unhappy Path Manually test passed successfully:
+#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
+#    Modify files in the expected directory to differ from the original.
+#    Manually run `config_default_user_home` recipe.
+#    Verified that the script fails and outputs the correct error message in chef-client.log.
 bash "Move #{node['cluster']['cluster_user_home']}" do
   user 'root'
   group 'root'
   code <<-EOH
-    set -e
```
dreambeyondorange: By removing this setting, we allow the process to continue even there are failures in the mkdir, rsync, usermod, chown, etc.  Does this behave properly in those cases?

It might be better in this case to move the diff check to a separate bash resource.  This might also make it easier to unit test

hehe7318: Great catch and advice, thank you! Have corrected it.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -54,8 +59,31 @@
...
+#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
+#    Modify files in the expected directory to differ from the original.
+#    Manually run `config_default_user_home` recipe.
+#    Verified that the script fails and outputs the correct error message in chef-client.log.
+bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
+  user 'root'
+  group 'root'
+  code <<-EOH
+    diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
+    diff_exit_code=$?
```
dreambeyondorange: why save the diff exit code here?

hehe7318: Previously is to see if it can prevent the script from being ended by `set -e`. But for now it's not necessary. Do we want to remove it?

hehe7318: Have removed it!

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -54,8 +59,30 @@
...
-    rm -rf /tmp#{node['cluster']['cluster_user_home']}
-    rm -rf #{node['cluster']['cluster_user_home']}
+  EOH
+end
+
+# Data integrity check
+# 1. Verifies data integrity by comparing the temporary backup directory and the new local home directory.
+# 2. If the data integrity check passes, it removes both the temporary backup directory and the original home directory.
+# 3. If the data integrity check fails, it outputs an error message and exits with an error code 1.
+# 4. Unhappy Path Manually test passed successfully:
```
gmarciani: No need to specify the tests you made in a comment. It's enough to describe it in the PR description.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -54,8 +59,30 @@
...
+#    Simulate a failure in the data integrity check by creating expected `node['cluster']['cluster_user_local_home']`
+#    Modify files in the expected directory to differ from the original.
+#    Manually run `config_default_user_home` recipe.
+#    Verified that the script fails and outputs the correct error message in chef-client.log.
+bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
```
gmarciani, line 69: Minor: The resource not only does a verification, but also a removal, so this should be reflected in the resource title. Something like "Data integrity check and cleanup for ..."

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -54,8 +59,30 @@
...
+  user 'root'
+  group 'root'
+  code <<-EOH
+    diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
+    if [ $? -eq 0 ]; then
+      rm -rf /tmp#{node['cluster']['cluster_user_home']}
+      rm -rf #{node['cluster']['cluster_user_home']}
+    else
+      echo "Data integrity check failed comparing #{node['cluster']['cluster_user_local_home']} and #{node['cluster']['cluster_user_home']}: $diff_output" >&2
+      exit 1
```
gmarciani, line 80: In previous revision you used to restart sshd because if the integrity check fails, sshd is not restarted by the resource at line 92. Now that it's removed, how can a user login in case of failure?

hehe7318: You said we do not restart it in case of failures. So I removed it. For now, customers can login by run this command in terminal:
```
aws ssm send-command \
    --instance-ids "i-0c806e61b7db08afe" \
    --document-name "AWS-RunShellScript" \
    --comment "Start SSHD service" \
    --parameters commands="sudo systemctl start sshd"
```
Or we can add back `systemctl start sshd` in the bash script. It has been tested and worked.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
gmarciani: Good point on the `set -e`.
Anyway, we cannot have a situation where sshd is not restarted, so my suggestion is to keep the sshd restart.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
gmarciani: > You said we do not restart it in case of failures.

I think there was a misunderstanding. I was explaining why we do stop at the beginning of the recipe, I didn't mean that we should not restart :-) 

Anyway, let's keep the restart

hehe7318: Done, added back.

dreambeyondorange (approved):
LGTM!

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -19,13 +19,22 @@
...
     group 'root'
     code <<-EOH
       rsync -a --ignore-existing /tmp/home/ /home
-      rm -rf /tmp/home/
+      diff_output=$(diff -r /tmp/home/ /home)
+      if [ $? -eq 0 ]; then
+        rm -rf /tmp/home/
+      else
+        echo "Data integrity check failed comparing /home and /tmp/home: $diff_output"
+        exit 1
```
gmarciani, line 36: Don't we need to restart sshd here as well? If not, why not?

hehe7318: Because we don't shut down the sshd service in this logic.

gmarciani (approved):

PR #2742, opened 2024-05-29
Renaming pcluster.sh profile to cookbook_environment.sh profile

himani2411:
### Description of changes
* Renaming pcluster.sh profile to cookbook_environment.sh profile which will be using cfn_bootstrap_virtualenv_path for all cfn-init, cfn-signal, cfn-hup commands

* This is not to pollute the default PATH

CLI https://github.com/aws/aws-parallelcluster/pull/6266

### Tests
* Created AMI's
* tests
```
  update:
    test_update.py::test_update_slurm:
      dimensions:
        - regions: ["eu-central-1"]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["ubuntu2004"]
    test_update.py::test_dynamic_file_systems_update:
      dimensions:
        - regions: ["eu-west-2"]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["centos7"]
          schedulers: ["slurm"]
    test_update.py::test_dynamic_file_systems_update_rollback:
      dimensions:
        - regions: [ "eu-west-2" ]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: [ "centos7" ]
          schedulers: [ "slurm" ]
```

### References

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   EFS is used for shared internal data. 
 - On Ubuntu systems, remove default logrotate configuration for cloud-init log files that clashed with the
   configuration coming from Parallelcluster.
+- Removing `/etc/profile.d/pcluster.sh` so that it's not executed at every user login and 
+  `cfn_bootstrap_virtualenv` is not added in PATH environment variable.
 
 3.9.1
 ------
--- a/cookbooks/aws-parallelcluster-environment/recipes/install/cfn_bootstrap.rb
+++ b/cookbooks/aws-parallelcluster-environment/recipes/install/cfn_bootstrap.rb
@@ -53,9 +53,9 @@ bash "Install CloudFormation helpers from #{cfnbootstrap_package}" do
   creates "#{virtualenv_path}/bin/cfn-hup"
 end
 
-# Add cfn_bootstrap virtualenv to default path
-template "/etc/profile.d/pcluster.sh" do
-  source "cfn_bootstrap/pcluster.sh.erb"
+# Add cfn_bootstrap virtualenv to a profile to be used only for cookbook
+template "#{node['cluster']['etc_dir']}/pcluster_cookbook_environment.sh" do
+  source "cfn_bootstrap/pcluster_cookbook_environment.sh.erb"
   owner 'root'
   group 'root'
   mode '0644'
--- a/cookbooks/aws-parallelcluster-environment/spec/unit/recipes/cfn_bootstrap_spec.rb
+++ b/cookbooks/aws-parallelcluster-environment/spec/unit/recipes/cfn_bootstrap_spec.rb
@@ -51,9 +51,9 @@ describe 'aws-parallelcluster-environment::cfn_bootstrap' do
           )
         end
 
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   EFS is used for shared internal data. 
 - On Ubuntu systems, remove default logrotate configuration for cloud-init log files that clashed with the
   configuration coming from Parallelcluster.
+- Renaming pcluster.sh profile to cookbook_environment.sh profile which will be used only for cfn commands.
```
gmarciani: Here we should explain the change from the user perspective. What's the consequence of this change for the user?

cookbooks/aws-parallelcluster-environment/recipes/install/cfn_bootstrap.rb
```diff
@@ -53,9 +53,9 @@
   creates "#{virtualenv_path}/bin/cfn-hup"
 end
 
-# Add cfn_bootstrap virtualenv to default path
-template "/etc/profile.d/pcluster.sh" do
-  source "cfn_bootstrap/pcluster.sh.erb"
+# Add cfn_bootstrap virtualenv to a profile to be used only for cookbook
+template "/etc/profile.d/cookbook_environment.sh" do
```
gmarciani: If we define the cookbook environment as a script in profile.d, then it gets executed at every user login before the execution of pcluster.sh.  Is this the expected behaviour?

Correct me if I'm wrong but I think the goal here is to define the environment only for the cookbook, so this script should be placed outside profile.d and sourced at the beginning of the user data.

Also, since this is a pcluster specific file, I suggest to name it `pcluster_cookbook_environment.sh`.

cookbooks/aws-parallelcluster-environment/templates/cfn_bootstrap/cookbook_environment.sh.erb
```diff
@@ -0,0 +1,8 @@
+#
+# cookbook_environment.sh:
+#   Setup ParallelCluster environment variables
```
gmarciani: Let's be a bit more specific: here we define only the enavrs that are expected to be execute by the cookbook only. Is that right?

cookbooks/aws-parallelcluster-environment/recipes/install/cfn_bootstrap.rb
himani2411: We are okay with the `cookbook_environment.sh` being in the profile.d as per the suggestion here https://github.com/aws/aws-parallelcluster/pull/6266#discussion_r1619169039

However, we want the PATH not being polluted as we were doing which affected every user. So we are now not changing the PATH but using a different variable

gmarciani: Thanks for the clarification. 
I still think that this is not a good security practice. A script must be added to profile.d if and only if it must be loaded at every user login; otherwise it must be moved out of that directory.

Even if the script is currently safe to be loaded for every user it is still an unneeded vector of risks.

gmarciani (approved):

PR #2748, opened 2024-06-10
[develop] Add logic to assert that the output of the diff command must do not contain lines starting with "Only in /tmp/home"

hehe7318:
### Description of changes
* The `test_shared_home` test is failing caused by integrity check process in `restore_home_shared_data.rb` before delete the backup directory. Now fix the test.
    - Add logic to assert that the output of the `diff` command must do not contain lines starting with "Only in /tmp/home"
    - This fix addresses issues where the /home directory contains pre-existing files such as slurm-*.out generated by running SLURM jobs, the automatically created lost+found directory and etc.
* Apply the same logic to `restore_internal_use_shared_data.rb` and `config_default_user_home.rb` to avoid any potential risks. Also modified their spec test to adopt the new changes.
* We decided to keep node['cluster']['cluster_user_home'] to avoid any potential file system data loss risks.

### Tests
* Failed `test_shared_home`(("FsxOpenZfs", "Efs"), ("FsxOntap", "Efs"), ("Ebs", "Efs")) integ tests can now successfully pass

### References
* PR that caused this issue: https://github.com/aws/aws-parallelcluster-cookbook/pull/2727

### Checklist


cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
...
+          fi
+        done
+      done < /tmp/home_existing_files.txt
+
+      # Sync data from /tmp/home to /home
       rsync -a --ignore-existing /tmp/home/ /home
-      diff_output=$(diff -r /tmp/home/ /home)
+
+      # Perform the diff check, excluding the original files
+      diff_output=$(eval diff -r $exclude_options /tmp/home /home)
```
gmarciani: why do we need to use eval here? Can't we cal the diff command directly?

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
...
     group 'root'
     code <<-EOH
+      # Generate a list of existing files and dirs in /home before the sync
+      find /home -mindepth 1 > /tmp/home_existing_files.txt
+
+      # Initialize an empty set for exclude options and directories to exclude
+      exclude_options=""
+      declare -A exclude_dirs
+
+      # Process each file in the list
```
gmarciani: It's a good practice to add a comment that explain what a more complex piece of code is doing. However this comment is too generic atmand it's not providing valuable info as it is. In general, if a comment does not provide valuable info, or it does provide info that are too obvious, then better to not have the comment at all.

Here I suggest to make it more specific: what's the processing is about?

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
...
+  # the newly synchronized files. This approach is necessary because the /home
+  # directory may contain pre-existing files such as slurm-*.out generated by
+  # running SLURM jobs and the automatically created lost+found directory.
   # Remove the backup after the copy is done and the data integrity is verified.
   bash "Restore /home" do
     user 'root'
     group 'root'
     code <<-EOH
+      # Generate a list of existing files and dirs in /home before the sync
+      find /home -mindepth 1 > /tmp/home_existing_files.txt
```
gmarciani: Why mindepth=1?

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
   # Restore the shared storage home data if it doesn't already exist
   # This is necessary to preserve any data in these directories that was
   # generated during the build of ParallelCluster AMIs after converting to
```
gmarciani, line 20: Why during the Pcluster AMI build? This recipe is executed at node bootstrap time.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
...
-  # shared storage and backed up to a temporary location previously
+  # shared storage and backed up to a temporary location previously.
   # Before removing the backup, ensure the data in the new home is the same
   # as the original to avoid any data loss or inconsistency. This is done
   # by using rsync to copy the data and diff to check for differences.
+  # The diff command excludes files that were originally in /home to focus on
+  # the newly synchronized files. This approach is necessary because the /home
+  # directory may contain pre-existing files such as slurm-*.out generated by
+  # running SLURM jobs and the automatically created lost+found directory.
```
gmarciani: why should we exclude these files form the integrity checks?

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
...
+      find /home -mindepth 1 > /tmp/home_existing_files.txt
+
+      # Initialize an empty set for exclude options and directories to exclude
+      exclude_options=""
+      declare -A exclude_dirs
+
+      # Process each file in the list
+      while IFS= read -r file; do
+        # Remove the /home/ prefix
+        relative_path=${file#/home/}
```
gmarciani: FYI there is a linux command to get the relative path: `realpath `

See example of usage: https://stackoverflow.com/questions/2564634/convert-absolute-path-into-relative-path-given-a-current-directory-using-bash

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -18,19 +18,65 @@
...
+        # Split the relative path by /
+        IFS='/' read -ra parts <<< "$relative_path"
+
+        for part in "${parts[@]}"; do
+          current_path="$current_path/$part"
+          if [ ! -e "$current_path" ]; then
+            # If the path does not exist in /tmp/home, add the last part of path to the exclude list
+            if [ -z "${exclude_dirs[$part]}" ]; then
+              exclude_dirs[$part]=1
+              exclude_options="$exclude_options --exclude=$part"
```
gmarciani: What about using `-X, --exclude-from=FILE` rather than per file exclusion?
You already have a file of paths to exclude so you can process it.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: I didn't modify it, it was like this previously. But I agree with you. Should we modify this comments?

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: The -mindepth 1 is used to avoid including the /home directory itself. This will only list the files and subdirectories under the /home directory.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: The diff command will check if the two directories are totally the same. So if the target dir(/home) has existing files which are different from backup dir(/tmp/home), the diff command will return exit code 1. That's the reason why integ test `test_shared_home` failed.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: The reason for using `eval` is to correctly handle variable substitution in the `diff` command. Calling `diff` directly can cause problems because `$exclude_options` contains multiple `--exclude` parameters, and passing them directly to `diff` may cause these options to be processed as a whole instead of as separate options. `eval` ensures that each option is parsed correctly.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: In this case, it's simpler and more efficient to manipulate the string directly, so why don't we keep the existing logic.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: But if we use `-X, --exclude-from=FILE` you mentioned rather than per file exclusion, we can remove `eval`.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Of course we can use it. Done!

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Also removed some not necessary comment.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
gmarciani: Yes, but not a blocker for the PR

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
gmarciani: sure, it was just for knowledge sharing (see FYI). I think it's useful to take the chance of a PR to share insights on alternative approaches. Either ways are fine.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
```diff
@@ -17,20 +17,65 @@
...
+      touch /tmp/exclude_options.txt
+
+      # Process each file and directory in the list to determine which paths should be excluded from the diff check
+      while IFS= read -r file; do
+        # Remove the /home/ prefix
+        relative_path=${file#/home/}
+        current_path="/tmp/home"
+
+        # Split the relative path by /
+        IFS='/' read -ra parts <<< "$relative_path"
```
gmarciani: We can avoid this nested loop.
Let's compare the content of /home and /tmp/home with `diff -r` assert that the output of the command must contain only lines starting with "Only in /home"

gmarciani (changes requested):

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Just understand the previous comment. It's saying the data generated during the build of ParallelCluster AMIs. Not this recipe is executed during the build of ParallelCluster AMIs.

hehe7318: Modified the comment to the original one.

cookbooks/aws-parallelcluster-environment/recipes/init/restore_home_shared_data.rb
hehe7318: Done! Thank you for the clever approach you provided.

hehe7318: Also applied the same logic to the other 2 recipes modified in the previous PR to avoid any potential error.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
```diff
@@ -29,11 +29,13 @@
           expect(chef_run).to run_bash("Verify data integrity for #{user_home}").with(
             code: <<-CODE
     diff_output=$(diff -r #{user_home} #{user_local_home})
-    if [ $? -eq 0 ]; then
+    if [[ $diff_output != *"Only in #{user_home}"* ]]; then
```
gmarciani: The integrity check should fail if there is at least one occurrence of the substring "Only in /tmp#{user_home}".
Here the check is doing the opposite: it fails if there is at least an occurrence of "Only in #{user_home}".

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
```diff
@@ -63,19 +63,21 @@
...
+# 1. Verifies data integrity by comparing the original home directory and the new local home directory.
 # 2. If the data integrity check passes, it removes both the temporary backup directory and the original home directory.
 # 3. If the data integrity check fails, it outputs an error message and exits with an error code 1.
 bash "Verify data integrity for #{node['cluster']['cluster_user_home']}" do
   user 'root'
   group 'root'
   code <<-EOH
     diff_output=$(diff -r #{node['cluster']['cluster_user_home']} #{node['cluster']['cluster_user_local_home']})
-    if [ $? -eq 0 ]; then
+    if [[ $diff_output != *"Only in #{node['cluster']['cluster_user_home']}"* ]]; then
```
gmarciani, line 76: The integrity check is supposed to fail if the tmp dir contains something that the original dir doesn't.
Why are you checking the original home and not the tmp folder?

gmarciani, line 76: *Best Practice (not a blocker, but to keep in mind)*
We should always refer positive check over negative checks because it improves readability and, as a consequence, maintainability.

It's easier to understand a positive check than a negative one.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
hehe7318: No, I would disagree. "Only in #{user_home}" is okay in this use case. We sync data from /tmp#{user_home} to #{user_home}, and we want all data in /tmp#{user_home} have been sync to #{user_home}. But there maybe are some pre-existing files in #{user_home}. So we can accept "Only in #{user_home}". But if "Only in /tmp#{user_home}" appear, that means some data sync failed.

hehe7318: And FYI, the #{user_home} in this case is not the original one. It has been replaced. So the /tmp#{user_home} contains the original data.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
hehe7318: In this case, #{node['cluster']['cluster_user_home']} is the original dir. And the temp dir has the same data as the original dir.
#{node['cluster']['cluster_user_local_home']} is the replaced dir.
We can accept the replaced dir has some pre-existing data. But we can not accept data in the original dir hasn't been synced to the replaced dir, which means we can not accept "Only in 'original dir'".

hehe7318: I would keep in mind, thank you! Should we modify the existing logic?

gmarciani: no because it's not worth it considering the deadlines. However, please keep track of this improvement in our backlog for the next release.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
gmarciani: > if "Only in /tmp#{user_home}" appear, that means some data sync failed.

Exactly, this is what I'm saying: The integrity check should fail if there is at least one occurrence of the substring "Only in /tmp#{user_home}" in the diff output.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
```diff
@@ -29,11 +29,13 @@
...
       rm -rf /tmp#{user_home}
       rm -rf #{user_home}
```
gmarciani: Also I'm just noticing this `rm -rf #{user_home}`. This is very risky and should never be done.
We should only delete temporary folders.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
hehe7318: Sorry, my previous explain is wrong, ignore them. I thought it was the `restore_home_shared_data` recipe.

This is the spec test for `config_default_user_home.rb`. Please look at the explain below in `config_default_user_home.rb`.

The goal of this spec test is to check if the logic existed. #{user_home} here is the original dir, we sync data from it. The #{user_local_home} is the new replaced dir. So we can not accept "Only in #{user_home}".

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
hehe7318: We had a new `user_local_home`. Why we still need `user_home` which has the data `user_local_home` already have? This `rm -rf #{user_home}` is the existing logic before I created the data integrity check PR. And we have used it for a long time. So it should be okay.

Also, it's not a recipe, it's a spec test.

My concern is maybe the temp folder is useless. In this case, it's not like the `restore_home_shared_data.rb`, in `restore_home_shared_data.rb` we replace /home dir, so we need a temp folder to store data. But in this case, we can directly sync data from #{node['cluster']['cluster_user_home'] to #{node['cluster']['cluster_user_local_home']. But also the temp folder is the existing logic before I created the data integrity check PR.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
gmarciani: I see, so I think there is a bit of confusion as a reviewer opf the code about what is the original dir and what is the destination dir. I suggest to make it clear with a comment or with variable names (even better).

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
gmarciani: > We had a new user_local_home. Why we still need user_home which has the data user_local_home already have?

We should avoid to do `rm -rf` on home directories because we do not want to take any risk of data loss, regardless if we need the data or not. We should not take the responsibility of removing /home data.

What if, for whatever race condition, a remote file system is still attached to the original home directory? 
In that case that rm -rf would remove everything from the remote file system and we would cause data loss.
We can say that there exist already a mitigation to this case: this recipe is executed before the recipes that mount the remote file systems, so we may be confident that at this point of the execution, no custom data exist yet in the home directories.

This is a good mitigation, but it's still not enough:
1. What if there exist a corner case where a remote file system is mounted before it?
A very improbable, but still possible, example may be a custom AMI with a init.d script that mounts a remote file system. I agree that it's really really improbable, is the value worth the risk?
2. Furthermore, at some point we may reorder our recipes forgetting about the requirement for this recipe to be executed at the beginning, making the recursive deletion a potential source of data loss.  

To sum up, we have a 99% mitigation of the data loss risk, but the potential cost of that 1% is disruptive.

In general, we should reason in terms of cost/opportunity.
1. What the worst case cost of doing rm -rf? In the worst case it causes data loss, so potentially millions of dollars in data and reputation.
2. What's the cost of keeping the original home directory?  A potential confusion in the user that may not understand why the original home is kept. This would cost barely nothing to the cx and only potential tickets to us that could be resolved with documentation.

> Also, it's not a recipe, it's a spec test.

A spec test typically reflects what the recipe is doing. If it does reflect it, then my comment applies to both the test and the recipe. If the test does not reflect the recipe, then we have a problem in the test.

cookbooks/aws-parallelcluster-environment/recipes/init/config_default_user_home.rb
hehe7318: Done with both the recipe and its spec test!

hehe7318: For the recipe, add a new comment. For the spec test, both modify the variable names and add a comment.

cookbooks/aws-parallelcluster-environment/spec/unit/recipes/config_default_user_home_spec.rb
hehe7318: Yes, that explain my confusion, thank you Giacomo. I totally agree with the potential risks you mentioned. 

> Furthermore, at some point we may reorder our recipes forgetting about the requirement for this recipe to be executed at the beginning, making the recursive deletion a potential source of data loss.

You are correct! Can not agree more.

> A spec test typically reflects what the recipe is doing. If it does reflect it, then my comment applies to both the test and the recipe. If the test does not reflect the recipe, then we have a problem in the test.

I know! I was just trying to say we should look at the recipe.

hehe7318: Done and add a comment for it.

gmarciani (approved):

PR #2776, opened 2024-07-12
RHEL8: switch to gem::version check, not float

nyetsche:
The code works for 8.9, but once you reach 8.10, the `.to_f()` calls turns 8.10 into a float, which compared to 8.4 is indeed "less than".


nyetsche:
I'm not a superb Rubyist, so I came up with the `Gem::Version` object comparison. Perhaps there's an easier way or it's bad form to `require rubygems` inside a cookbook?

hanwen-pcluste (approved):
Thank you for the contribution.

I will approve and merge this PR.

We created another PR for further improvement https://github.com/aws/aws-parallelcluster-cookbook/pull/2777

PR #2788, opened 2024-08-14
[ADC] Set default node package for build image command

hgreebe:
### Description of changes
* Set default node package for build image command
* So that `build-image` command can be run without having to specify custom packages in ADC

### DIFF
--- a/cookbooks/aws-parallelcluster-awsbatch/recipes/install.rb
+++ b/cookbooks/aws-parallelcluster-awsbatch/recipes/install.rb
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 return if platform?('redhat')
+return if aws_region.start_with?("us-iso")
 
 include_recipe "::awsbatch_virtualenv"
 
@@ -28,9 +29,6 @@ end
 
 # Check whether install a custom aws-parallelcluster-awsbatch-cli package or the standard one
 # Install awsbatch cli into awsbatch virtual env
-if aws_region.start_with?("us-iso") && !node['cluster']['custom_awsbatchcli_package'].empty?
-  node.default['cluster']['custom_awsbatchcli_package'] = "#{node['cluster']['artifacts_s3_url']}/dependencies/awsbatch/aws-parallelcluster.tgz"
-end
 if !node['cluster']['custom_awsbatchcli_package'].nil? && !node['cluster']['custom_awsbatchcli_package'].empty?
   # Install custom aws-parallelcluster package
   bash "install aws-parallelcluster-awsbatch-cli" do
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-awsbatch/recipes/install.rb
```diff
@@ -28,9 +29,6 @@
 
 # Check whether install a custom aws-parallelcluster-awsbatch-cli package or the standard one
 # Install awsbatch cli into awsbatch virtual env
-if aws_region.start_with?("us-iso") && !node['cluster']['custom_awsbatchcli_package'].empty?
-  node.default['cluster']['custom_awsbatchcli_package'] = "#{node['cluster']['artifacts_s3_url']}/dependencies/awsbatch/aws-parallelcluster.tgz"
-end
```
hanwen-pcluste, line 33: Is only removing the code enough? In the below code, it will try to install from pip and will fail, right

himani2411: From what I know we dont even support Batch scheduler in ADC

cookbooks/aws-parallelcluster-computefleet/recipes/install/parallelcluster_node.rb
```diff
@@ -34,7 +34,9 @@
 end
 
 if aws_region.start_with?("us-iso") && !is_custom_node?
-  node.default['cluster']['custom_node_package'] = "#{node['cluster']['artifacts_s3_url']}/dependencies/node/aws-parallelcluster-node.tgz"
+  node_package = "aws-parallelcluster-node-#{node['cluster']['parallelcluster-node-version']}.tgz"
+
+  node.default['cluster']['custom_node_package'] = "#{node['cluster']['s3_url']}/parallelcluster/#{node['cluster']['parallelcluster-node-version']}/node/#{node_package}"
```
himani2411, line 39: Are we making changes in the S3 location to upload default Node package in this location. As I dont see node package in s3 for other regions?

cookbooks/aws-parallelcluster-awsbatch/recipes/install.rb
hgreebe: it skips the whole install part for aws batch in ADC due to this line: `return if aws_region.start_with?("us-iso")`

cookbooks/aws-parallelcluster-computefleet/recipes/install/parallelcluster_node.rb
hgreebe: when we release in adc, we will also have to release a node and cookbook package so that they can use those for build image. I think we have a similar path where we have been publish the cookbook in s3 in commercial.

hanwen-pcluste (approved):

PR #2797, opened 2024-08-29
Prevent pcluster from changing default permissions on /home directory

hgreebe:
### Description of changes
* Prevents the permissions of the `/home` directory from being reset every time a compute node or login node is launched
* The old behavior works as follows:
  * When a cluster is launched with 0 static nodes, the `/home` directory on the head node has permission 755
    * When a dynamic node is launched, the `/home` directory on the head node has its permissions changed to 1777
  * When a cluster with at least 1 static node is launched, the `/home` directory on the head node has permission 1777 because the launching of the static node changes the permissions
  * When the permissions of the `/home` directory is changed, and a dynamic node is launched, the permission of `/home` is changed back to 1777
* The new behavior works as follows:
  * When a cluster is launched with 0 or more static nodes, the `/home` directory on the head node has permission 755
  * When a dynamic node is launched, the permissions of `/home` does not change
  * When the permissions of the `/home` directory is changed, and a dynamic node is launched, the permission of `/home` is not changed

### Tests
* Tried all the cases mentioned in the description
* Ran an integration test with the following test configuration:
```
test-suites:
  storage:
    test_shared_home.py::test_shared_home:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["c5.xlarge"]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
```

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/resources/manage_ebs.rb
+++ b/cookbooks/aws-parallelcluster-environment/resources/manage_ebs.rb
@@ -13,6 +13,7 @@ unified_mode true
 
 property :shared_dir_array, Array, required: %i(mount export unmount unexport)
 property :vol_array, Array, required: %i(mount unmount)
+property :mode, String, default: "1777"
 
 default_action :mount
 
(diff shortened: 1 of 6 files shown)


cookbooks/aws-parallelcluster-environment/resources/manage_ebs.rb
```diff
@@ -51,6 +51,12 @@
       retries 10
       retry_delay 6
     end
+
+    directory shared_dir_array[index] do
+      owner 'root'
+      group 'root'
+      mode '1777'
+    end
```
hanwen-pcluste: why is this change necessary?

dreambeyondorange (approved):

PR #2852, opened 2024-12-11
On Ubuntu22, install the NVIDIA driver using the gcc version used to compile the kernel.

gmarciani:
### Description of changes
On Ubuntu22, install the NVIDIA driver using the gcc version used to compile the kernel.
    
This is required because, NVIDIA driver must be compiled with the same gcc version used by the kernel.
If this is not the case, the NVIDIA driver installation would fail a compiler version check.
On newer version of Ubuntu22.04 (kernel 6.8+), the kernel is compiled with gcc-12, however gcc-11 is installed as default version by build-essentials, making this change necessary.

* Pinned Cinc to v23 for the execution of spec tests in the PR. We use Cinc v23 locally. With Cinc v24 the newly introduced unit test fails because the test cannot find the resource for ubuntu22. this does not happen with Cinc v23 and does not happen during cookbook execution (which uses Cinc v18)

### Tests
* Build image Ubuntu 22.04 arm (x86 affected by another issue, will be tested on pipeline)
* test_efa (which is part of the suite used to validate gpus) succeeded


demartinofra (dismissed):

hanwen-cluster (approved):
Reapprove based on [demartinofra](https://github.com/demartinofra)'s approval

PR #2855, opened 2024-12-18
Support attaching multiple network interfaces to the same network card

hanwen-cluster:
Prior to this commit, the code had two assumptions:
1. Single network card instances can have only one network interface 
1.1. therefore, when there are more than one network interface, the code made a IMDS call to retrieve device index (network interface index), which is only available on instances with multiple network cards. Having a secondary network interface on single network card instance failed the code and caused instance launch failures.
3. Each network card can have only one network interface 
2.1. therefore, the route table is unique to each network card. Having multiple network interfaces on a network card confused the code and generated wrong route tables.

To fix (1), this commit uses a fallback value 0 when retrieval of device index fails. 
To fix (2), this commit names the route tables in a way that is unique to network interface and network card.

FYI:
1. Network card is the physical card. Network interface is the virtual concept. Each network card can have multiple network interfaces (which in AWS is 1 or 2)
2. "Network interface" is synonym to "device"

### Tests
* Manually modified launch template to add secondary network interface to compute nodes. Jobs ran successfully. SSH works on both interfaces.

### Checklist


hgreebe (approved):

cookbooks/aws-parallelcluster-environment/files/amazon-2023/network_interfaces/configure_nw_interface.sh
```diff
@@ -23,12 +24,13 @@ fi
 
 cd "$configuration_directory"
 
-ROUTE_TABLE=100${DEVICE_NUMBER}
+SUFFIX=$(printf "%03d" $NETWORK_CARD_INDEX)$(printf "%03d" $DEVICE_NUMBER)
```
enrico-usai, line 27: it would be good to add an example of the expected output in a comment

cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
```diff
@@ -20,9 +21,11 @@ then
   exit 1
 fi
 
-ROUTE_TABLE="100${DEVICE_NUMBER}"
+SUFFIX=$(printf "%03d" $NETWORK_CARD_INDEX)$(printf "%03d" $DEVICE_NUMBER)
 
-echo "Configuring ${DEVICE_NAME} with IP:${DEVICE_IP_ADDRESS} CIDR_PREFIX:${CIDR_PREFIX_LENGTH} NETMASK:${NETMASK} GW:${GW_IP_ADDRESS} ROUTING_TABLE:${ROUTE_TABLE}"
+ROUTE_TABLE="1${SUFFIX}"
+
+echo "Configuring device name: ${DEVICE_NAME} with IP:${DEVICE_IP_ADDRESS} CIDR_PREFIX:${CIDR_PREFIX_LENGTH} NETMASK:${NETMASK} GW:${GW_IP_ADDRESS} ROUTING_TABLE:${ROUTE_TABLE}"
```
enrico-usai, line 28: what about adding network card index too? ${NETWORK_CARD_INDEX}

cookbooks/aws-parallelcluster-environment/files/redhat-8.network_interfaces/configure_nw_interface.sh
```diff
@@ -23,9 +24,10 @@ then
 fi
 
 con_name="System ${DEVICE_NAME}"
-route_table="100${DEVICE_NUMBER}"
-priority="100${DEVICE_NUMBER}"
-metric="100${DEVICE_NUMBER}"
+SUFFIX=$(printf "%03d" $NETWORK_CARD_INDEX)$(printf "%03d" $DEVICE_NUMBER)
```
enrico-usai, line 27: To be consistent with existing code we should use lowercase variable names for internal variables, so we should have `suffix` rather than `SUFFIX`.

cookbooks/aws-parallelcluster-environment/recipes/config/network_interfaces.rb
```diff
@@ -16,7 +16,13 @@
 return if on_docker?
 
 def network_card_index(mac, token)
+  # This IMDS call is not available on single NIC instance, therefore fallback to 0
```
enrico-usai, line 19: It would be good to link in a comment the public documentation where this behavior is described.

cookbooks/aws-parallelcluster-environment/files/ubuntu/network_interfaces/configure_nw_interface.sh
```diff
@@ -40,7 +41,8 @@ if [ "${STATIC_IP_CONFIG}" = "" ]
 fi
 
 FILE="/etc/netplan/${DEVICE_NAME}.yaml"
-ROUTE_TABLE="100${DEVICE_NUMBER}"
+SUFFIX=$(printf "%03d" $NETWORK_CARD_INDEX)$(printf "%03d" $DEVICE_NUMBER)
+ROUTE_TABLE="1${SUFFIX}"
```
enrico-usai, line 45: I would have expected kitchen tests to fail in the https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-environment/test/controls/network_interfaces_spec.rb after this changes.

Could you please run them and double check if we can improve existing coverage?

PR #2872, opened 2025-02-05
Download http-parser from s3

hgreebe:
### Description of changes
* Download http-parser from s3
* Modify patching script for ADC to patch al2023

### Tests
* Build image for al2023 in ADC

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/templates/isolated/patch-iso-instance.sh.erb
+++ b/cookbooks/aws-parallelcluster-environment/templates/isolated/patch-iso-instance.sh.erb
@@ -19,15 +19,41 @@ REGION="$(get_instance_region)"
 
 source /etc/os-release
 OS="${ID}${VERSION_ID}"
-[[ "${OS}" != "amzn2" ]] && echo "[ERROR] Unsupported OS '${OS}'. Configuration supported only on Amazon Linux 2." && exit 1
+
+if [[ "${OS}" != "amzn2" && "${OS}" != "amzn2023" ]]; then
+  echo "[ERROR] Unsupported OS '${OS}'. Configuration supported only on Amazon Linux 2 and Amazon Linux 2023."
+  exit 1
+fi
 
 echo "[INFO] Starting: instance configuration for US isolated region"
 
 echo "[INFO] Starting: installation of packages from amazon Linux 2 repository for US isolated region"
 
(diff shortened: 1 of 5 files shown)


cookbooks/aws-parallelcluster-environment/templates/isolated/patch-iso-instance.sh.erb
```diff
@@ -39,30 +65,35 @@ mirrorlist_expire=300
 report_instanceid=yes
 REPO_DEFINITION
 
-yum --disablerepo="*" --enablerepo="amzn2-iso" install -y "*-${REGION}"
+  yum --disablerepo="*" --enablerepo="${OS}-iso" install -y "*-${REGION}"
```
hanwen-cluster, line 68: Why is this command only necessary on amzn2?

cookbooks/aws-parallelcluster-environment/templates/isolated/patch-iso-instance.sh.erb
```diff
@@ -39,30 +65,35 @@ mirrorlist_expire=300
...
 
-USERS=(<%= @users %>)
 CA_BUNDLE="/etc/pki/${REGION}/certs/ca-bundle.pem"
 
-for user in "${USERS[@]}"; do
-  echo "[INFO] Setting CA bundle ${CA_BUNDLE} for user ${user}"
-  sudo mkhomedir_helper $user
-  sudo -u $user aws configure set ca_bundle "$CA_BUNDLE"
-done
+sudo aws configure set ca_bundle "$CA_BUNDLE"
```
hanwen-cluster, line 79: Why is this change?

hgreebe: Because setting the system wide environment variables make them apply for each user, so it is unneccessary to loop through each user

cookbooks/aws-parallelcluster-environment/templates/isolated/patch-iso-instance.sh.erb
hgreebe: Because for amzn2023 we do `dnf install -y amazon-linux-repo-iso ca-certificates-isob` which is what is in the documentation. But I kept what we do for amzn2 the same, so that is why they install the region-specific packages differently

hanwen-cluster (approved):

PR #2994, opened 2025-07-15
Delete cluster name state file whenever slurm accounting is configured or updated

hgreebe:
### Description of changes
* Covers slurm known error where cluster id saved in `/var/spool/slurm.state/clustername` does not match the cluster id that slurm dbd has 
* This fix needs to be in both `clear_slurm_accounting` and `config_slurm_accounting`
  * If a cluster is created without slurm accounting and then updated to have slurm accounting it will run `config_slurm_accounting` and have the cluster id mismatch
  * If a cluster is created with slurm accounting and then updated, it will run `clear_slurm_accounting` and have the cluster id mismatch
* Example error message:
```
[2025-07-10T00:54:46.362] fatal: CLUSTER ID MISMATCH.
slurmctld has been started with "ClusterID=4018"  from the state files in StateSaveLocation, but the DBD thinks it should be "3073".
Running multiple clusters from a shared StateSaveLocation WILL CAUSE CORRUPTION.
Remove /var/spool/slurm.state/clustername to override this safety check if this is intentional.
```

### Tests
* Created a new AMI with the cookbook changes and ran the `test_slurm_accounting` and `test_slurm integ` tests.
* Tested the if I have one job running and one job pending and then stop slurmctld and delete `/var/spool/slurm.state/clustername`, once I restart slurmctld, the slurm state remains the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 **CHANGES**
 - Ubuntu 20.04 is no longer supported.
 - Upgrade Slurm to version 24.11.5.
+- Addressed cluster id mismatch known issue by deleting the file `/var/spool/slurm.state/clustername` before configuring Slurm accounting.
 - Upgrade DCV to version 2024.0-19030.
 - Remove `berkshelf`. All cookbooks are local and do not need `berkshelf` dependency management.
 
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
@@ -76,6 +76,10 @@ service "slurmdbd" do
   action action
 end unless on_docker?
 
+file "/var/spool/slurm.state/clustername" do
+  action "delete"
+end
+
 if node['cluster']['slurmdbd_service_enabled'] == "true"
   # After starting slurmdbd the database may not be fully responsive yet and
   # its bootstrapping may fail. We need to wait for sacctmgr to successfully
@@ -87,11 +91,6 @@ if node['cluster']['slurmdbd_service_enabled'] == "true"
     retries node['cluster']['slurmdbd_response_retries']
     retry_delay 10
   end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
-
-
-  file '/var/spool/slurm.state/clustername' do
-    action :delete
-  end
 
   bash "bootstrap slurm database" do
     user 'root'
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
@@ -88,13 +88,9 @@ if node['cluster']['slurmdbd_service_enabled'] == "true"
     retry_delay 10
   end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
 
-  bash "Remove existing cluster name state file" do
-    user 'root'
-    group 'root'
-    code <<-CLUSTERSTATE
-      rm /var/spool/slurm.state/clustername
-    CLUSTERSTATE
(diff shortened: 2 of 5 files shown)


gmarciani:
> if I have one job running and one job pending and then stop slurmctld and delete /var/spool/slurm.state/clustername, once I restart slurmctld, the slurm state remains the same.

Do we have an integ test capturing the same scenario when we execute the cluster update?
If not, it would be a nice to have to capture it

cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
```diff
@@ -88,6 +88,15 @@
     retry_delay 10
   end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
 
+  bash "Remove existing cluster name state file" do
```
gmarciani: Could you please cover this new logic with a spec test?

cookbooks/aws-parallelcluster-slurm/recipes/update/clear_slurm_accounting.rb
```diff
@@ -23,3 +23,12 @@
   supports restart: false
   action %i(disable stop)
 end
+
+bash "Remove existing cluster name state file" do
```
gmarciani: Could you please cover this new logic with a spec test?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
```diff
@@ -88,6 +88,15 @@
...
+    user 'root'
+    group 'root'
+    code <<-CLUSTERSTATE
+      rm /var/spool/slurm.state/clustername
+    CLUSTERSTATE
+    only_if { ::File.exist?('/var/spool/slurm.state/clustername') }
```
gmarciani: If we want to use bash, why using an `only_if` rather than having a `rm -f`?
Also, why using bash and not using the file resource with action delete?

cookbooks/aws-parallelcluster-slurm/recipes/update/clear_slurm_accounting.rb
```diff
@@ -23,3 +23,12 @@
...
+  user 'root'
+  group 'root'
+  code <<-CLUSTERSTATE
+      rm /var/spool/slurm.state/clustername
+    CLUSTERSTATE
+  only_if { ::File.exist?('/var/spool/slurm.state/clustername') }
```
gmarciani: Same as https://github.com/aws/aws-parallelcluster-cookbook/pull/2994/files#r2208466853

gmarciani:
With this change we are fixing a bug. Can we surface it in the changelog?

cookbooks/aws-parallelcluster-slurm/recipes/update/clear_slurm_accounting.rb
gmarciani: This logic is duplicated. What about reducing code duplication by defining this logic into a function and call that function or use the file resources to delete the file?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
gmarciani: thanks for adding unit tests!

cookbooks/aws-parallelcluster-slurm/recipes/update/clear_slurm_accounting.rb
gmarciani: Thanks for adding unit tests!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 **CHANGES**
 - Ubuntu 20.04 is no longer supported.
 - Upgrade Slurm to version 24.11.5.
+  - Addressed cluster id mismatch known issue by deleting the file `/var/spool/slurm.state/clustername` before configuring slurm accounting.
```
gmarciani: Typo: extra indent, and Slurm must be capitalized

gmarciani (approved):

PR #2887, opened 2025-02-25
Upgrade NVIDIA driver and cuda version

hgreebe:
### Description of changes
* Upgrade NVIDIA driver version from 550.127.08 to 570.86.15 for all OSs except AL2
* Upgrade CUDA version from 12.4.1 to 12.8.0 for all OSs except AL2

### Tests
* Ran test_build_image for all OSs

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2987, opened 2025-07-03
Add NVSwitch device ID for p6 instance type

himani2411:
### Description of changes
* Add NVSwitch device ID for p6 instance type as NVIDIA Fabric manager needs to be enabled for GPU Health Checks to be invoked.

Steps for Device ID: https://nvidia.custhelp.com/app/answers/detail/a_id/2040/~/identifying-the-graphics-card-model-and-device-id-in-a-pc

* This PR addresses a technical blocker and does not provide full support of p6 instance type.

### Tests
* Cluster launch with p4d instance type and Log line (which is now removed)
```
[2025-07-03T18:27:07+00:00] INFO: NVSwitch works 6

    * service[nvidia-fabricmanager] action start[2025-07-03T18:27:07+00:00] INFO: Processing service[nvidia-fabricmanager] action start (aws-parallelcluster-platform::test line 36)
 (up to date)
```

### References
* Link to impacted open issues.
* Link to related PRs in other packages (i.e. cookbook, node).
* Link to documentation useful to understand the changes.

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2996, opened 2025-07-29
[Nvidia-Imex] Installing Nvidia-imex as part of ParallelCluster Build Image 

himani2411:
### Description of changes

* Adding resource which installs nvidia-imex from S3, which is installed only in commercial regions and ignored in isolated regions
*  * Version of nvidia-imex should be same as nvidia version
* * We skip installing nvidia-imex for AL2 instance as NVIDIA Driver 570+ are not supported with AL2 https://github.com/aws/aws-parallelcluster-cookbook/pull/2887
*  * We skip installation if nvidia-imex is already installed or if [Installation/NvidiaSoftware/Enabled](https://docs.aws.amazon.com/parallelcluster/latest/ug/Build-v3.html#Build-v3-Installation) is kept False.
*  * nvidia-imex configuration files are kept in a shared location `/opt/parallelcluster/shared/nvidia-imex`
*  * We overwrite nvidia-imex files are configured only during cluster creation phase and only on gb200.
*  * NVIDIA Configuration files contains **default values** except for where we re-direct nvidia-imex logs to system logs which are pushed in CW Log group
*  * We keep `/opt/parallelcluster/shared/nvidia-imex/nodes_config_<LaunchTemplateID>.cfg` containing fake IP addresses for being able to start enable nvidia-imex service during configuration Phase of cluster creation.
*. * We use `lt-123456789012` Launch template Id suffix to adhere to `max length of 256 characters.
* Similarly configured changes in NVIDIA-Fabric Manager to be enabled when p6e instance is used as a compute fleet.
* Add Kitchen and InSpec tests

### Tests
* Inpsec tests
* Build AMI for all OSSes
 ** Tested by HARDCODing NVIDIA Driver and Fabric Manager BUILD IMAGE as nvidia-imex was not available for older version.** 

### References

### Checklist


cookbooks/aws-parallelcluster-platform/resources/fabric_manager/partial/_fabric_manager_install_debian.rb
```diff
@@ -35,9 +35,14 @@
...
+end
+
+def name_suffix
   arm_instance? ? 'arm64' : 'amd64'
 end
 
 def fabric_manager_url
-  "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_fabric/#{platform}/#{fabric_manager_package}_#{fabric_manager_version}-1_#{arch_suffix}.deb"
+  # "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_fabric/#{platform}/#{fabric_manager_package}_#{fabric_manager_version}-1_#{arch_suffix}.deb"
+  "https://developer.download.nvidia.com/compute/cuda/repos/#{platform}/#{arch_suffix}/#{fabric_manager_package}_#{fabric_manager_version}-1_#{name_suffix}.deb"
```
gmarciani: [BLOCKING] Remove hard coded values after testing

cookbooks/aws-parallelcluster-platform/attributes/platform.rb
```diff
@@ -16,7 +16,7 @@
 
 # NVidia
 default['cluster']['nvidia']['enabled'] = 'no'
-default['cluster']['nvidia']['driver_version'] = '570.86.15'
+default['cluster']['nvidia']['driver_version'] = '570.158.01'
```
himani2411: Same as https://github.com/aws/aws-parallelcluster-cookbook/pull/2996#discussion_r2240795077. Added only as part of testing and will be removed in this PR.

cookbooks/aws-parallelcluster-platform/resources/fabric_manager/partial/_fabric_manager_install_rhel.rb
```diff
@@ -40,5 +40,6 @@ def arch_suffix
 end
 
 def fabric_manager_url
-  "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_fabric/#{platform}/#{fabric_manager_package}-#{fabric_manager_version}-1.#{arch_suffix}.rpm"
+  # "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_fabric/#{platform}/#{fabric_manager_package}-#{fabric_manager_version}-1.#{arch_suffix}.rpm"
```
himani2411: Same as https://github.com/aws/aws-parallelcluster-cookbook/pull/2996#discussion_r2240795077. Added only as part of testing and will be removed in this PR.

cookbooks/aws-parallelcluster-platform/resources/nvidia_driver/partial/_nvidia_driver_common.rb
```diff
@@ -96,7 +96,8 @@ def _nvidia_driver_version
 end
 
 def nvidia_driver_url
-  "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_driver/NVIDIA-Linux-#{nvidia_arch}-#{_nvidia_driver_version}.run"
+  # "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_driver/NVIDIA-Linux-#{nvidia_arch}-#{_nvidia_driver_version}.run"
```
himani2411: Same as https://github.com/aws/aws-parallelcluster-cookbook/pull/2996#discussion_r2240795077. Added only as part of testing and will be removed in this PR.

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
```diff
@@ -18,3 +18,21 @@ def is_process_running(process_name)
...
 end
+
+#
+# Get Count of GPUs in instance
+#
+def get_nvswitch_count(device_id)
+  shell_out("lspci -d #{device_id} | wc -l").stdout.strip.to_i
+end
+
+def get_device_ids
```
gmarciani, line 29: [minor] Did you find public documentation specifying these device ids as reference?
If you do, then let's put a link here for reference

cookbooks/aws-parallelcluster-platform/recipes/install/nvidia_install.rb
```diff
@@ -24,3 +24,5 @@
 fabric_manager 'Install Nvidia Fabric Manager'
 
 nvidia_dcgm 'install Nvidia datacenter-gpu-manager'
+
+nvidia_imex 'Install Nvidia-imex'
```
gmarciani: [minor] Sometimes you refere to it as `nvidia-imex`, sometimes as `Nvidia-imex`. Let's use a consistent naming. I'm more in favor of `nvidia-imex`

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
```diff
@@ -18,3 +18,21 @@ def is_process_running(process_name)
...
+  #  A100 (P4), H100(P5), B200(P6) and GB200()p6e) systems have NVSwitches
+  # NVSwitch device id is 10de:1af1 for P4 instance
+  # NVSwitch device id is 10de:22a3 for P5 instance
+  # NVSwitch device id is 10de:2901 for P6 instance
+  # NVSwitch device id is 10de:2941 for P6e instance
+  # We sum the count for all these deviceIds as output of lscpi command will be >0
+  # for only one device ID based on the instance type
```
gmarciani: You're not summing here, but in `get_nvswitches` function. Let's move this comment there

cookbooks/aws-parallelcluster-platform/attributes/platform.rb
```diff
@@ -16,12 +16,15 @@
...
 default['cluster']['nvidia']['dcgm_version'] = '3.3.6'
 if platform?('amazon') && node['platform_version'] == "2"
   default['cluster']['nvidia']['driver_version'] = '550.127.08'
 end
 
+# Nvidia-imex
+default['cluster']['nvidia']['imex']['shared_dir'] = "#{node['cluster']['shared_dir']}/nvidia-imex"
```
gmarciani, line 26: This folder is shared across the whole cluster. Do we need to share ti with login nodes as well and why?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_alinux2023.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Copyright:: 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: Here and other new files: fix copyright year

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_amazon2.rb
```diff
@@ -0,0 +1,27 @@
...
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+provides :nvidia_imex, platform: 'amazon', platform_version: '2'
+
+use 'partial/_nvidia_imex_common.rb'
+use 'partial/_nvidia_imex_rhel.rb'
+
+def imex_installed
```
gmarciani: The name of this function is misleading because nvidia-imex is actually not installed on AL2.
If you need a function that decides whether the package must be installed or not then I suggest to call it `install_nvidia_imex?`

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_rocky8.rb
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and limitations under the License.
+
+provides :nvidia_imex, platform: 'rocky' do |node|
+  node['platform_version'].to_i >= 8
+end
+
+use 'partial/_nvidia_imex_common.rb'
+use 'partial/_nvidia_imex_rhel.rb'
+
+def _nvidia_imex_version
```
gmarciani: Here and other places: misleading function name: `_nvidia_imex_version` contains `nvidia_imex_full_version` plus something more, so the naming should be the other way around.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_alinux2023.rb
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and limitations under the License.
+
+provides :nvidia_imex, platform: 'amazon' do |node|
+  node['platform_version'].to_i == 2023
+end
+
+use 'partial/_nvidia_imex_common.rb'
+use 'partial/_nvidia_imex_rhel.rb'
+
+def _nvidia_imex_version
```
gmarciani: `_nvidia_imex_version` returns the same value for every Os, but ubuntu22+ What about putting it in the common logic and override only for Ubuntu?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+    action :create
+  end
+
+  action_install_imex
+  # Save Imex version in Node Attributes for InSpec Tests
+  node.default['cluster']['nvidia']['imex']['version'] = nvidia_imex_full_version
+  node.default['cluster']['nvidia']['imex']['package'] = "#{nvidia_imex_service}-#{nvidia_driver_major_version}"
+  node_attributes 'dump node attributes'
+
+  nvidia_repo 'remove nvidia repository' do
```
gmarciani: Why removing the nvidia repo after the installation?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+    action :add
+  end
+
+  directory node['cluster']['nvidia']['imex']['shared_dir']
+
+  template "#{node['cluster']['nvidia']['imex']['shared_dir']}/config.cfg" do
+    source 'nvidia-imex/nvidia-imex-config.erb'
+    owner 'root'
+    group 'root'
+    mode '0755'
```
gmarciani: Why 0755 on a config file? What about `0644`?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  template "#{node['cluster']['nvidia']['imex']['shared_dir']}/nodes_config.cfg" do
+    source 'nvidia-imex/nvidia-imex-nodes.erb'
+    owner 'root'
+    group 'root'
+    mode '0755'
```
gmarciani: Why 0755 on a config file? What about `0644`?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  nvidia_repo 'remove nvidia repository' do
+    action :remove
+  end
+end
+
+action :configure do
+  return unless imex_installed && node['cluster']['node_type'] == "ComputeFleet"
+  # Start nvidia-imex on p6e-gb200 and only on ComputeFleet
+  if get_nvswitch_count(get_device_ids['gb200']) > 1
```
gmarciani, line 34: In general, it is safer to:
1. get values from a dictionary providing a default value (eg: `hash.fetch("d", "default_value")`); OR
2. make the receiving function `get_nvswitch_count` able to handle nil values
In this specific case (2) is preferable because a failed fetching in ruby does not raise an exception, but returns a nil value.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+    service nvidia_imex_service do
+      action %i(start enable)
```
gmarciani: Why starting before enabling it, rather than the other way around?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+
+def nvidia_driver_major_version
+  node['cluster']['nvidia']['driver_version'].split('.')[0]
+end
+
+def nvidia_imex_service
+  'nvidia-imex'
+end
+
+def nvidia_imex_full_version
```
gmarciani, line 84: Misleading function name: the function is expected to return the imex full version, but it does contain the imex versio. suffix, as the full imex version is made of ${nvidia_driver_major_version}-${nvidia_driver_version}-1

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+  "#{node['cluster']['nvidia']['driver_version']}-1"
+end
+
+def imex_installed
```
gmarciani: Functions returning boolean values typically have names ending with `?`

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,91 @@
...
+  ::File.exist?("/usr/bin/#{nvidia_imex_service}") || ::File.exist?("/usr/bin/#{nvidia_imex_service}-ctl")
+end
+
+def nvidia_enabled_or_installed?
```
gmarciani, line 92: This function seems general, not specific to imex. What about moving this function to the shared helpers where the other functions are defined?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_debian.rb
```diff
@@ -0,0 +1,22 @@
...
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+action :install_imex do
+  apt_package "Install nvidia-imex" do
+    package_name "#{nvidia_imex_service}-#{nvidia_driver_major_version}"
```
gmarciani: Package name is OS dependent, same as the package version. What about defining the package name in the resource entry point rather than in the partial (the same way you're doing for the version)? In this way we just need to take a look at the entry point to spot the peculiarities of each OS.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_rhel.rb
```diff
@@ -0,0 +1,20 @@
...
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+action :install_imex do
+  install_packages 'Install nvidia-imex' do
```
gmarciani: Why we do not need to specify the version here and in debiam we do?

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
```diff
@@ -0,0 +1,26 @@
...
+Requires=network-online.target
+
+[Service]
+Environment="KRB5_CLIENT_KTNAME=/etc/krb5.keytab"
+User=root
+PrivateTmp=false
+Type=forking
+TimeoutStartSec=infinity
+
+ExecStart=/usr/bin/nvidia-imex -c <%= node['cluster']['nvidia']['imex']['shared_dir'] %>/config.cfg
```
gmarciani: We reference the path to nvidia-imex binary and config file in different places. What about injecting the variables directly?

`ExecStart=<%= node['cluster']['nvidia']['imex']['binary'] %>` -c <%= node['cluster']['nvidia']['imex']['config'] %>`

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
```diff
@@ -0,0 +1,26 @@
+[Unit]
```
gmarciani, line 4: If this file is taken directly from the nvidia-imex documentation, then link to the official doc documenting it. In this way if we face issues at some point with the service definition we have a quick reference to the suggested values from nvidia and also spot quickly the changes we made on top of it.

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
```diff
@@ -0,0 +1,26 @@
...
+User=root
+PrivateTmp=false
+Type=forking
+TimeoutStartSec=infinity
+
+ExecStart=/usr/bin/nvidia-imex -c <%= node['cluster']['nvidia']['imex']['shared_dir'] %>/config.cfg
+
+LimitCORE=infinity
+
+Restart=on-failure
```
gmarciani, line 19: Why restarting on failure rather than restarting always?
One of the advantage of restarting always is that we capture whatever termination scenario

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
```diff
@@ -0,0 +1,26 @@
+[Unit]
+Description=NVIDIA IMEX service
+After=network-online.target
+Requires=network-online.target
+
+[Service]
+Environment="KRB5_CLIENT_KTNAME=/etc/krb5.keytab"
```
gmarciani: Why do we need kerberos configuration?
What if the user has configured kerberos to point to a different principal file overriding `default_keytab_name` in `/etc/krb5.conf`?

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-nodes.erb
```diff
@@ -0,0 +1,3 @@
+## Please replace below fake IP's
+172.31.51.93
```
gmarciani, line 2: How did you come with these values?
If they are placeholder just to get imex started, what about using 0.0.0.0?
See https://www.techtarget.com/searchnetworking/answer/What-is-the-IP-address-0000-used-for#:~:text=In%20the%20simplest%20terms%2C%20the,the%20destination%20address%20is%20unavailable.

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-nodes.erb
```diff
@@ -0,0 +1,3 @@
+## Please replace below fake IP's
```
gmarciani: We need to write a bit more here:
1. Replace with what?
2. Why fake?

(I'm aware of it, but a user first approaching this file may be confused about its role and the next steps)

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
```diff
@@ -0,0 +1,217 @@
...
+
+#   Description: IMEX logging levels
+#   Possible Values:
+#           0  - All the logging is disabled
+#           1  - Set log level to CRITICAL and above
+#           2  - Set log level to ERROR and above
+#           3  - Set log level to WARNING and above
+#           4  - Set log level to INFO and above
+#       Default Value: 4
+LOG_LEVEL=4
```
gmarciani, line 13: Here and for other parameters:
Why do we need to override the parameter with its default value?
What about commenting it out the same way we did for other params?

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
```diff
@@ -0,0 +1,217 @@
...
+
+#   Description: Filename for IMEX logs
+#   Possible Values:
+#       Full path/filename string (max length of 256). Logs will be redirected
+#       to console(stderr). If the specified log file can't be opened or the
+#       path is empty.
+#   Default Value: /var/log/nvidia-imex.log
+# LOG_FILE_NAME=/var/log/nvidia-imex.log
```
gmarciani, line 21: does it make sense to push nvidia-imex logs to cloudwatch and configure log rotation?

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
```diff
@@ -0,0 +1,217 @@
...
+#   Possible Values:
+#           Any value between 0 and 65535
+#   Default Value: 50000
+SERVER_PORT=50000
+
+#   Description: Name of file containing IP addresses of nodes
+#   Possible Values:
+#       Full path/filename string (max length of 256).
+#       Default Value: /etc/nvidia-imex/nodes_config.cfg
+IMEX_NODE_CONFIG_FILE=<%= node['cluster']['nvidia']['imex']['shared_dir'] %>/nodes_config.cfg
```
gmarciani: Mentioned also in another comment. What about injecting directly the attribute containing the path to the config file rather than redefining it?

I mean:

```
IMEX_NODE_CONFIG_FILE=<%= node['cluster']['nvidia']['imex']['config'] %>
```

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
```diff
@@ -0,0 +1,217 @@
...
+#             have connected, allowing them time to safely clean up any potentially hanging references
+#  Default value: RECOVERY
+IMEX_WAIT_FOR_QUORUM=RECOVERY
+
+#  Description:  Enable authentication and encryption between nodes.
+#  Possible Values:
+#      0:  Disable encryption and authentication
+#      1:  Enable encryption and authentication
+#  Default value: 0
+IMEX_ENABLE_AUTH_ENCRYPTION=0
```
gmarciani, line 113: why not enabling encryption?

cookbooks/aws-parallelcluster-platform/resources/fabric_manager/partial/_fabric_manager_install_debian.rb
himani2411: Yes this is same as https://github.com/aws/aws-parallelcluster-cookbook/pull/2996#discussion_r2240969910

cookbooks/aws-parallelcluster-platform/recipes/install/nvidia_install.rb
himani2411: will update

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
himani2411: I do not have a doc but its the same procedure as mentioned https://github.com/aws/aws-parallelcluster-cookbook/pull/2987

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
himani2411: will do

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_alinux2023.rb
himani2411: Done

cookbooks/aws-parallelcluster-platform/attributes/platform.rb
himani2411: we dont have to share it across to login nodes. Since we do not have a compute fleet specific folder I kept it in this location

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
himani2411: We can, I wanted to make minimal changes in this PR which is why i have re-directed it to syslog

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_amazon2.rb
himani2411: The name is for checking if `imex_installed` is installed or not in common.rb.
Will add the `?`
I am just overriding this for AL2

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_rocky8.rb
himani2411: Removed this as it is no longer needed!

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_alinux2023.rb
himani2411: removed as it was no longer needed

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: Chef is idempotent so the action sequence dose not matter, it will enable and start the service. 
 but for the sake of readability will updated.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: The full version is not necessarily the one you refer. 

When we do a apt/dnf list installed packages we see the `"#{node['cluster']['nvidia']['driver_version']}-1"` which is why I set the `node.default['cluster']['nvidia']['imex']['version']` as part of node attributes which I use in Inspec tests.
But the package naming convention requires during installation me to add the ` ${nvidia_driver_major_version}-${nvidia_driver_version}-1` depending on the platform, so I install using the full name of the package with the exact version so there is no mismatch.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: Updated

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: We can move this later when we actually use it for other resources which is not required for this PR

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_rhel.rb
himani2411: `install_packages` is an internal resource ParallelCluster has created, so it does not have a version specified.
we use `package` chef resource in `install_packages` which is also what I was initially using for both Debian and redhat.
but the naming convention did not match for Debian when I specify using the same format as redhat. So I 
separated them

The version is added as part of the package name itself for redhat installation

himani2411: So For debian I use `apt_package` and I tie the version as the naming convetion differ

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
himani2411: So you want a variable for the binary? Why? we dont decide this path and I dont see the point of add an attribute if its not something i use anywhere else

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
himani2411: I picked the file from an instance in which i downloaded it. Will add the official doc in the comment

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
himani2411: I kept it as is due to the installation from the instance itself that I was working with. But I am going to remove it as official doc does not keep it https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/gettingstarted.html#on-linux-based-systems

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
himani2411: not sure I understand why we want to keep it `always` especially when this should not be enabled or running on all the instance types.

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-nodes.erb
himani2411: I can keep it and was initially going to do so. This is what I got from DLAMI that i used for testing

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
himani2411: I keep the file as it is generated by nvidi-imex installation. So even if the default value is not necessary, the installation itself keeps it, so I followed the same

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
himani2411: I can keep a backlog item to keep them separted and move them to CW

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
himani2411: I kept the file as it is when it is generated by nvidia-imex installation, plus I would need to setup other values for enabling the auth either through SSL/TLS and other options like the SOURCE, TARGET_OVERRIDE,etc as per https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/config.html. i think this can be done in the next iteration and make the minimal chnages for now

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_debian.rb
himani2411: Ok got it

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_rhel.rb
himani2411: Since I cannot tie in the version in package name as I can for redhat and `  install_packages` as well as `package` both do not have `version` attribute, I directly use `apt_package` for debian

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: I am following the same procedure we did for Fabric Manager. Likely the reason for removing the repo could potentially open the upgrade of Nvidia and other nvidia supported software which we do not want. So either I version lock these nvidia packages or i remove the repo. And I dont want to add to the list of package inconsistencies by keeping the repo

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_rhel.rb
```diff
@@ -0,0 +1,20 @@
...
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+action :install_imex do
```
gmarciani, line 15: Installing imex from nvidia public repo rather than our bucket can compromise build image in isolated environments. We accepted this limitation for US ISO regions, but we support isolated environments everywhere else.

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-nodes.erb
himani2411: Updated

cookbooks/aws-parallelcluster-platform/attributes/platform.rb
mjkoop: You do not want to share this beyond a IMEX domain boundary -- e.g. each ultraserver should have a separate config as it should only have IPs of nodes that it can actually share GPU memory over NVLink.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -0,0 +1,106 @@
...
+  node_attributes 'dump node attributes'
+end
+
+action :configure do
+  return unless imex_installed? && node['cluster']['node_type'] == "ComputeFleet"
+  # Start nvidia-imex on p6e-gb200 and only on ComputeFleet
+  if get_nvswitch_count(get_device_ids['gb200']) > 1
+    # For each Compute Resource, we generate a unique NVIDIA IMEX configuration file,
+    # if one doesn't already exist in a common, shared location.
+    template nvidia_imex_nodes_conf_file do
```
gmarciani, line 37: The overall design assumes the imex nodes config file to be shared in the cluster.
This is helpful to centralize the orchestration from the head node and also to simplify the torubleshooting. 
However it would be a blocker for the per-job deployment type(https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/deployment.html#per-job-wide). Avoid blockers for such deployment would be beneficial for not only for our users, but also to us, because it would allow us to vend the automated configuraiton of imex following the NVIDIA example [SLURM Scheduler Integration](https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/deployment.html#per-job-wide)

himani2411: Lets have an offline discussion as there would be blockers on the Job wide deployment model

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_rhel.rb
himani2411: Resolved by using our s3 buckets

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: As discussed we will keep the exiting changes and we can later make changes as per naming convention or whichever is easier for the design of Custom Actions we recommend

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
himani2411: I have kept it specific to each node and dont see the point of making another attribute. But if later we can find that it makes the implementation of recommended custom actions easier we can chnage in next iteration

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_rhel.rb
himani2411: No longer relevant as we download it from S3

cookbooks/aws-parallelcluster-platform/attributes/platform.rb
himani2411: Hi @mjkoop 

I think we will re-iterate on the shared location or not as per the deployment Model we choose for the recommended custom actions that we will suggest to CX for management of the IP addresses. https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/deployment.html#per-job-wide

For now we will keep it shared

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex-config.erb
himani2411: @mjkoop 
Do you have any recommendations on this?

mjkoop: I'd recommend in the future using encryption, but don't see it as a blocker as initial release is a guide vs full-implementation. More importantly, I'd recommend we let the customer know about how the IMEX-channels work so they understand in a single channel mode there is no protection across users on the same node.

himani2411: Just to be sure I wasnt planning on making changes for IMex channel 's in the next iteration either as that would be more of a customization w.r.t DirectoryServices.

cookbooks/aws-parallelcluster-platform/templates/nvidia-imex/nvidia-imex.service.erb
gmarciani: On those instance types where it should not run, the service is disabled, so the restart policy is not even taken into account.
The advantage of always compared to on-failure is that the service gets restarted regardless the exit code. This is just a safety net. 

However since you took the unit definition from NVIDIA recommedation, they probably have good reasons to prefer on-failure rather than always.
Let's go with on-failure.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/nvidia_imex_amazon2.rb
```diff
@@ -0,0 +1,27 @@
...
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+provides :nvidia_imex, platform: 'amazon', platform_version: '2'
+
+use 'partial/_nvidia_imex_common.rb'
+use 'partial/_nvidia_imex_rhel.rb'
+
+def imex_installed?
```
gmarciani, line 20: Not a blocker but this function name is misleading. 
IMEx is never installed on AL2. You need a function to determine wheteher or not imex should be installed. So I would rename this to `install_imex?` or `skip_imex_installation?`

gmarciani (approved):

PR #2999, opened 2025-08-06
Remove device name and ip address from required parameters of configure_nw_interfaces.sh

hgreebe:
### Description of changes
* Remove device name and ip address from required parameters of `configure_nw_interfaces.sh`
* Device name and ip address do not exist for efa-only nw interfaces.
* This adds a check that is device name and ip address are not included, then it is an efa-only interface so the rest of the script is skipped.
* We use [ec2 instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) to get the parameters like ip address, network card index etc. This instance metadata does not have interface_type so we can not use it to see if it is efa-only. This limitation is why we are validating that it is efa-only using device_name and ip addess.

### Tests
* Ran 'test_multiple_nics` integration test

### References
* CLI change: https://github.com/aws/aws-parallelcluster/pull/6930

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/files/amazon-2023/network_interfaces/configure_nw_interface.sh
+++ b/cookbooks/aws-parallelcluster-environment/files/amazon-2023/network_interfaces/configure_nw_interface.sh
@@ -3,16 +3,27 @@
 set -ex
 
 if
-  [ -z "${DEVICE_NAME}" ] ||          # name of the device
   [ -z "${DEVICE_NUMBER}" ] ||       # index of the device
   [ -z "${NETWORK_CARD_INDEX}" ] ||   # index of the network card
-  [ -z "${DEVICE_IP_ADDRESS}" ] ||       # ip of the device
   [ -z "${MAC}" ] ||                 # mac address of the device
   [ -z "${CIDR_BLOCK}" ]                 # CIDR block of the subnet
 then
   echo 'One or more environment variables missing'
   exit 1
 fi
+
+# Check if this is an EFA-only interface (no device name or IP)
+if [ -z "${DEVICE_NAME}" ] && [ -z "${DEVICE_IP_ADDRESS}" ]; then
+  echo "EFA-only interface detected - skipping IP configuration"
+  exit 0
+fi
+
+# If one of these is missing but not both, it is an invalid configuration
+if [ -z "${DEVICE_NAME}" ] || [ -z "${DEVICE_IP_ADDRESS}" ]; then
+    echo "Device name or IP address is missing"
+    exit 1
+fi
+
 echo "Configuring NIC, Device name: ${DEVICE_NAME}, Device number: ${DEVICE_NUMBER}, Network card index:${NETWORK_CARD_INDEX}"
 
 configuration_directory="/etc/systemd/network"
--- a/cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
+++ b/cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
@@ -22,6 +22,12 @@ then
   exit 1
 fi
 
+# Check if this is an EFA-only interface (no device name or IP)
+if [ -z "${DEVICE_NAME}" ] || [ -z "${DEVICE_IP_ADDRESS}" ]; then
+  echo "EFA-only interface detected - skipping IP configuration"
+  exit 0
+fi
+
 SUFFIX=$NETWORK_CARD_INDEX$(printf "%02d" $DEVICE_NUMBER)
 
 ROUTE_TABLE="$(( $SUFFIX + 1000 ))"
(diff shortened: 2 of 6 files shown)


cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
```diff
@@ -8,19 +8,26 @@
 
 set -e
 
+# commented out inputs are not required and don't exist for efa-only interfaces
```
gmarciani: [minor] typo `that are not required`

cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
```diff
@@ -8,19 +8,26 @@
...
 if
-  [ -z "${DEVICE_NAME}" ] ||          # name of the device
+#  [ -z "${DEVICE_NAME}" ] ||          # name of the device
```
gmarciani: Why not removing rather than commenting out?

cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
```diff
@@ -8,19 +8,26 @@
...
+#  [ -z "${DEVICE_IP_ADDRESS}" ] ||    # ip address to assign to the interface
   [ -z "${CIDR_PREFIX_LENGTH}" ] ||   # the prefix length of the device IP cidr block
   [ -z "${NETMASK}" ]                 # netmask to apply to device ip address
 then
   echo 'One or more environment variables missing'
   exit 1
 fi
 
+# Check if this is an EFA-only interface (no device name or IP)
+if [ -z "${DEVICE_NAME}" ] || [ -z "${DEVICE_IP_ADDRESS}" ]; then
```
gmarciani: Where it is documented that efa-only interfaces do not have a device name?

gmarciani:
this change contributes to supporting GB200 instance types. Let' make sure to capture an entry in the changelog

cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
hgreebe: I couldn't find official documentation but when you try to run this script with an efa-only interface type it has the following input:
```
{"DEVICE_NAME"=>"", "DEVICE_NUMBER"=>"1", "NETWORK_CARD_INDEX"=>"1", "GW_IP_ADDRESS"=>"192.168.112.1", "DEVICE_IP_ADDRESS"=>nil, "CIDR_PREFIX_LENGTH"=>"20", "NETMASK"=>"255.255.240.0", "CIDR_BLOCK"=>"192.168.112.0/20", "MAC"=>"02:6e:6f:a7:24:8d"}
```

So it doesn't input a device name or ip address.

gmarciani: This logic is ok for EFA only, but removing the check on `DEVICE_NAME` and `DEVICE_IP_ADDRESS` (lines above) we are accepting invalid values for the other types of interface.

Before your change we required all network interfaces to have both `DEVICE_NAME` and `DEVICE_IP_ADDRESS`. Now we don't check anymore. Removing the check makes sense for EFA only interfaces, but it does not for the other types.

A more robust logic should be: require all variables are set, unless both `DEVICE_NAME` and `DEVICE_IP_ADDRESS` are unset

cookbooks/aws-parallelcluster-environment/files/default/network_interfaces/configure_nw_interface.sh
```diff
@@ -9,18 +9,28 @@
...
 
+# Check if this is an EFA-only interface (no device name or IP)
+if [ -z "${DEVICE_NAME}" ] && [ -z "${DEVICE_IP_ADDRESS}" ]; then
+  echo "EFA-only interface detected - skipping IP configuration"
+  exit 0
+fi
+
+# If one of these is missing but not both, it is an invalid configuration
+if [ -z "${DEVICE_NAME}" ] || [ -z "${DEVICE_IP_ADDRESS}" ]; then
+    echo "One or more environment variables missing"
```
gmarciani: Please specify the name of variables that may be missing.
This script does not have -x, so if we see from the logs the message 'One or more environment variables missing' we do not know if it is the first or second check that is failing.

cookbooks/aws-parallelcluster-environment/files/amazon-2023/network_interfaces/configure_nw_interface.sh
```diff
@@ -3,16 +3,27 @@
 set -ex
```
gmarciani, line 3: thank you for aligning all Os specific scripts.
Since we are doing it, please use the same execution options: `set -ex` everywhere to facilitate troubleshooting in case of issues

gmarciani (approved):

PR #3005, opened 2025-08-12
[IMEX] Remove unnecessary condition to skip creation of main and nodes config file.

gmarciani:
### Description of changes
Remove unnecessary condition to skip creation of main and nodes config file.
We figured out there is no reason to skip the configuration of IMEX config file when queues are updated.
The only relevant condition is to skip it when the files are already created because we do not want to overwrite configurations made by the user (or by an automation) on cluster create/update.

### Tests
* [SUCCEEDED] `test_gb200`
* Spec test (updated)

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
+++ b/cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
@@ -39,8 +39,7 @@ action :configure do
       owner 'root'
       group 'root'
       mode '0755'
-      action :create
-      not_if { file_exists_and_cluster_update?(nvidia_imex_nodes_conf_file) }
+      action :create_if_missing
     end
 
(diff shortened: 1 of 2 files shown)


cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -39,17 +39,15 @@
       owner 'root'
       group 'root'
       mode '0755'
-      action :create
-      not_if { file_exists_and_cluster_update?(nvidia_imex_nodes_conf_file) }
+      action :create_if_missing
```
himani2411, line 42: Can you update the associate Unit tests?

gmarciani: Done

himani2411 (dismissed):

himani2411 (approved):

PR #3020, opened 2025-09-04
Add support for `Feature` specification for ComputeResources

efim-a-efim:
### Description of changes
With this changes, users will be able to specify additional `Feature` values for ComputeResources. Those `Feature`-s will be safely added to the Slurm nodes specification for the partition.

Rationale: In many cases, especially with complex cluster setups, it's crucial to have additional `Feature`-s that can be used to filter nodes for the jobs.

### Tests
Added the following tests:
* `test/unit/slurm/test_slurm_config_generator.py::test_generate_slurm_config_with_custom_features` - tests config generation with custom features specified. Corresponding test data is in `test/unit/slurm/test_generate_slurm_config_with_custom_features`.

**NOTE**: test data for many tests were changed because the new features generation algorithm requires all features in `Feature=...` in config files to be sorted. So many test files were affected, but the only change that were made to them is features ordering.

#### References
This PR has a complimentary PR in `aws-parallelcluster` repo: https://github.com/aws/aws-parallelcluster/pull/6962


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -24,6 +24,8 @@ def __init__(self, queue_name: str, compute_resource_config: Dict, no_gpu, memor
         self.name = compute_resource_config["Name"]
         self.disable_multithreading = compute_resource_config["DisableSimultaneousMultithreading"]
         self.custom_settings = compute_resource_config.get("CustomSlurmSettings", {})
+        # If Features are in CustomSlurmSettings, fetch them and remove from CustomSlurmSettings
+        self.custom_features = self.custom_settings.pop("Feature", "")
```
demartinofra, line 28: Instead of mutating the custom_settings object by removing features I would keep custom_settings aligned with what is injected from config and filter out features when building the `custom += f" {param}={value}"` string as part of the nodename rendering.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -85,24 +87,38 @@ def _definitions(self, dynamic=False):
         return definitions
 
     def _features(self, dynamic=False):
+        features = set(self.custom_features.split(','))
+        # this is a simple workaround for empty Features: split will give [''] and we'll discard it 
+        # as well as all empty strings.
+        features.discard('')
```
demartinofra, line 93: default to None when reading the Feature string and skip the splitting so that you don't need to discard ''?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
```diff
@@ -85,24 +87,38 @@ def _definitions(self, dynamic=False):
...
+
+        # remove "system" features from configured features, so that they will never interfere
+        for feat in ('static', 'dynamic', 'gpu', 'efa') :
+            features.discard(feat)
```
demartinofra, line 97: you could consider leaving these unaltered - if for any reason the user passes them in we could just honor them as overrides

demartinofra (commented):
Thanks for contributing with this change. I left a few minor comments if you could take a look.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
efim-a-efim: Because it's the easiest way to work around human errors like double spaces, space at the end of the string, etc.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
efim-a-efim: OK, I was thinking that they are "system" and shouldn't be managed by users. If it's not a problem - I'll leave them.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/config_renderer.py
efim-a-efim: OK, I'll check how to do it better. Thanks. I'll come back for review :)

PR #3022, opened 2025-09-10
[B200] Add support for P6-B200 instance type

gmarciani:
### Description of changes
With this change we add support for p6-b200 instance type for all supported OSs except Amazon Linux 2, which does not support the required version of the nvidia driver.
To support p6-b200 we need:
1. infiniband packages
2. infiniband kernel module loaded at boot time
3. [NVIDIA NVLink Subnet Manager (NVLSM)](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#getting-started-with-nvlink-subnet-manager), which is a component required by Fabric Manager to work with NVLink in this instance type.

The setup defined here is consistent with what is suggested by EC2 ([doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#public-nvidia-driver)).

This PR requires https://github.com/aws/aws-parallelcluster/pull/6990

### Open Points
To install IB packages I needed to use package manager directly, not through the chef resource. When using the chef resource, the installation fails because no candidate can be found for the packages. When using the package manager directly it succeeds. This is likely a caching issue, but I do not have yet a clear answer. I will investigate the root cause, but this is not a blocker for the PR.

### Tests
* Manually built custom AMIs for all OSs executing the same steps of this PR + https://github.com/aws/aws-parallelcluster-cookbook/pull/3022 and validated that these steps are the ones required to make P6-B200 work (verified with dcgmi diagnosis)
* ONGOING Build image on all OSes
* PENDING integ tests: gpu_health_check


cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/nvidia_nvlsm_ubuntu22+.rb
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and limitations under the License.
+
+provides :nvidia_nvlsm, platform: 'ubuntu' do |node|
+  node['platform_version'].to_i >= 22
+end
+
+use 'partial/_nvidia_nvlsm_common.rb'
+use 'partial/_nvidia_nvlsm_debian.rb'
+
+def platform
```
himani2411, line 22: [Non-blocking] Nitpick: we should stop naming s3 path in a way which forces us to keep changing the URL path

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
```diff
@@ -0,0 +1,109 @@
...
+  # OS dependent
+end
+
+def nvidia_nvlsm_install_dependencies_commands
+  # OS dependent
+end
+
+def nvlsm_installation_enabled?
+  if on_docker? ||
+     node['cluster']['nvidia']['nvlsm']['skip'] == true ||
```
himani2411: Nitpick: why keep `['cluster']['nvidia']['nvlsm']['skip']` instead of `['cluster']['nvidia']['nvlsm']['enable']` which is what we have been using so far for other packages?

cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+            runner = runner(platform: platform, version: version, step_into: ['nvidia_nvlsm']) do |node|
+              node.override['cluster']['artifacts_s3_url'] = cluster_artifacts_s3_url
+              node.override['cluster']['sources_dir'] = source_dir
+              node.automatic['kernel']['machine'] = arch
+            end
+            ConvergeNvidiaNvlsm.install(runner)
+          end
+          cached(:node) { chef_run.node }
+
+          cached(:nvlsm_version) { "2025.03.9-1" }
```
himani2411, line 119: [Non-blocking] Nitpick: Can we keep this as any FAKE Value?

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/nvidia_nvlsm_ubuntu22+.rb
gmarciani: I totally agree with you. For this new dependency we applied the path structure we used so far for other dependencies.I'll keep this outside of the scope of this PR for now, but is we have capacity before the release I will fix it for NVLSM

cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
gmarciani: TL:DR Not in this case

Fake values in unit tests are meant to surface that the behavior does not depend on specific values.
I'm always if favor of fake values when this is the case. This would be the case if, for example, the nvlsm version would be an input from the user and we want to test that our behavior changes accordingly.

However, nvlsm version is not an input value, we want it to be hardwired in the resource, so we want the unit test to verify that specific version.

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_debian.rb
```diff
@@ -0,0 +1,37 @@
...
+def nvidia_nvlsm_checksum
+  if arm_instance?
+    '6bb405a3494d9fcb6dad6e641d02afb71fd4e2f6a2b4ed3d5cf6cbff22964eb3'
+  else
+    '61f280e469624c43eecb0e08305452887e02f73e4763252a41f728d1843f1cc5'
+  end
+end
+
+def nvidia_nvlsm_install_commands
+  "dpkg -i #{nvidia_nvlsm_package_full_name} && apt-mark hold #{nvidia_nvlsm_package}"
```
hanwen-cluster, line 32: Why do we need `apt-mark hold`?

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_rhel.rb
```diff
@@ -0,0 +1,37 @@
...
+def nvidia_nvlsm_checksum
+  if arm_instance?
+    'f21f14843c11ce64136fd1c3fa763b7511e18f160695f54b2a8d763776313539'
+  else
+    '88d5e52183bb5ee763eb864bbd119b591e7f45af32c52bd7ba0aa8f74fc19057'
+  end
+end
+
+def nvidia_nvlsm_install_commands
+  "yum install -y #{nvidia_nvlsm_package_full_name} && yum versionlock #{nvidia_nvlsm_package}"
```
hanwen-cluster, line 32: Why do we need `yum versionlock`?

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
gmarciani: I can align it for sake of consistency. 

However, I think we took a slightly bad practice here.
Attributes are meant to change a default behavior. In this way, if an attribute is missing it will not change the default behavior.
If the default behavior is to execute something, the attribute should express the idea to skip it
Example: for nvidia driver we have the "enabled" attribute because the default behavior is to skip the nvidia driver installation.

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_debian.rb
gmarciani: Because we want the package to be stick with the specified version.
We adopted the same approach also for other dependencies., .e.g nvidia-imex and fabric manager.

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_rhel.rb
gmarciani: Same as above https://github.com/aws/aws-parallelcluster-cookbook/pull/3022/files#r2353448837

hanwen-cluster (approved):

cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
himani2411: Got it!

himani2411 (approved):

PR #3026, opened 2025-09-18
[Gb200] Skip IMEX configuration File creation if the Directory does not exist

himani2411:
### Description of changes - OLDER NOTES NEED TO BE UPDATED

Skip IMEX configuration File creation if the Directory does not exist
* We create the Directory as part of AMI creation
* Skip starting Imex service if the service file does not exist.
Without this change we see below error
```
* template[/opt/parallelcluster/shared/nvidia-imex/nodes_config_q1_cr1.cfg] action create_if_missing[2025-09-18T00:59:55+00:00] INFO: Processing template[/opt/parallelcluster/shared/nvidia-imex/nodes_config_q1_cr1.cfg] action create_if_missing (aws-parallelcluster-platform::nvidia_config line 37)
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]: 
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      * Parent directory /opt/parallelcluster/shared/nvidia-imex does not exist.
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      ================================================================================
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      Error executing action `create_if_missing` on resource 'template[/opt/parallelcluster/shared/nvidia-imex/nodes_config_q1_cr1.cfg]'
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      ================================================================================
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]: 
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      Chef::Exceptions::EnclosingDirectoryDoesNotExist
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      ------------------------------------------------
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      Parent directory /opt/parallelcluster/shared/nvidia-imex does not exist.
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]: 
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      Resource Declaration:
Sep 18 01:00:00 ip-172-31-42-141 user-data[28237]:      ---------------------
```
* in DLAMI the service file is at `/usr/lib/systemd/system/nvidia-imex.service` and they do not configure `/etc/nvidia-imex/nodes_config.cfg` so we should not enable the service as well
 ```
* ○ nvidia-imex.service - NVIDIA IMEX service
     Loaded: loaded (/usr/lib/systemd/system/nvidia-imex.service; enabled; preset: disabled)
     Active: inactive (dead) since Thu 2025-09-18 17:54:33 UTC; 1h 7min ago
   Duration: 2ms
    Process: 2284 ExecStart=/usr/bin/nvidia-imex -c /etc/nvidia-imex/config.cfg (code=exited, status=0/SUCCESS)
   Main PID: 2371 (code=exited, status=0/SUCCESS)
        CPU: 21ms

Sep 18 17:54:31 ip-172-31-69-32.ec2.internal systemd[1]: Starting nvidia-imex.service - NVIDIA IMEX service...
Sep 18 17:54:33 ip-172-31-69-32.ec2.internal systemd[1]: Started nvidia-imex.service - NVIDIA IMEX service.
Sep 18 17:54:33 ip-172-31-69-32.ec2.internal systemd[1]: nvidia-imex.service: Deactivated successfully.
```
### Tests
* [ONGOING] test_gb200 with Custom AMI 

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
+++ b/cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
@@ -19,53 +19,52 @@ action :install do
   return unless nvidia_enabled_or_installed?
   return if on_docker? || imex_installed? || aws_region.start_with?("us-iso")
 
-  # We are using the existence of this directory to verify if Imex was installed by ParallelCluster
-  directory node['cluster']['nvidia']['imex']['shared_dir']
-
   action_install_imex
+
+  action_create_configuration_files
   # Save Imex version in Node Attributes for InSpec Tests
   node.default['cluster']['nvidia']['imex']['version'] = nvidia_imex_full_version
   node.default['cluster']['nvidia']['imex']['package'] = nvidia_imex_package
   node_attributes 'dump node attributes'
 end
 
+action :create_configuration_files do
+  # We create or update IMEX configuration files if ParallelCluster is installing IMEX
+  template nvidia_imex_nodes_conf_file do
+    source 'nvidia-imex/nvidia-imex-nodes.erb'
+    owner 'root'
+    group 'root'
+    mode '0755'
+    action :create
+  end
+
+  template nvidia_imex_main_conf_file do
+    source 'nvidia-imex/nvidia-imex-config.erb'
+    owner 'root'
+    group 'root'
+    mode '0755'
+    action :create
+    variables(imex_nodes_config_file_path: nvidia_imex_nodes_conf_file)
+  end
+
+  # We keep nvidia-imex.service file in this location to give precedence to pcluster configured service file.
+  template "/etc/systemd/system/#{nvidia_imex_service}.service" do
+    source 'nvidia-imex/nvidia-imex.service.erb'
+    owner 'root'
+    group 'root'
+    mode '0644'
+    action :create
+    variables(imex_main_config_file_path: nvidia_imex_main_conf_file)
+  end
+end
+
 action :configure do
   return unless imex_installed? && node['cluster']['node_type'] == "ComputeFleet"
   # Start nvidia-imex on p6e-gb200 and only on ComputeFleet
-  if (is_gb200_node? && pcluster_installed_imex?) || enable_force_configuration?
-    # For each Compute Resource, we generate a unique NVIDIA IMEX configuration file,
-    # if one doesn't already exist in a common, shared location.
-    template nvidia_imex_nodes_conf_file do
-      source 'nvidia-imex/nvidia-imex-nodes.erb'
-      owner 'root'
-      group 'root'
-      mode '0755'
-      action :create_if_missing
-    end
-
-    template nvidia_imex_main_conf_file do
-      source 'nvidia-imex/nvidia-imex-config.erb'
-      owner 'root'
-      group 'root'
-      mode '0755'
-      action :create_if_missing
-      variables(imex_nodes_config_file_path: nvidia_imex_nodes_conf_file)
-    end
-
-    # We keep nvidia-imex.service file in this location to give precedence to pcluster configured service file.
-    template "/etc/systemd/system/#{nvidia_imex_service}.service" do
-      source 'nvidia-imex/nvidia-imex.service.erb'
-      owner 'root'
-      group 'root'
-      mode '0644'
-      action :create
-      variables(imex_main_config_file_path: nvidia_imex_main_conf_file)
-    end
-
+  if is_gb200_node? || enable_force_configuration?
     service nvidia_imex_service do
       action %i(enable start)
       supports status: true
-      only_if { ::File.exist?("/etc/systemd/system/#{nvidia_imex_service}.service") }
     end
   end
 end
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb [hunk 1]
```diff
@@ -58,11 +60,13 @@
       mode '0644'
       action :create
       variables(imex_main_config_file_path: nvidia_imex_main_conf_file)
+      only_if { Dir.exist?(node['cluster']['nvidia']['imex']['shared_dir']) }
```
gmarciani, line 63: It's more correct and safer to skip this resource if the service file already exists. In this way we would prevent overwriting the existing file when custom amis are used.

We can remove this condition and use `:create_if_missing`

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -40,6 +40,7 @@
       group 'root'
       mode '0755'
       action :create_if_missing
+      only_if { Dir.exist?(node['cluster']['nvidia']['imex']['shared_dir']) }
```
gmarciani, line 43: You're using this condition as a way to identify if the AMI executed the install phase. If the configuration phase requires such directory, why not creating it in the configuration phase and avoid every `only_if`?

himani2411: Because the DLAMI already comes with nvidia-imex and we skip the installation during install phase. So if I make this change for creating the directory in config phase, we basically override their installation!

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: The service file of IMEX for DLAMI is on a different location. So I am writing the conditions based on or locations instead of finding if their location exists or not as they could change their location in future

himani2411: Added the details in description

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
gmarciani: [BLOCKING] If you create the directory, put the imex main config file and nodes config file there you're not overriding their installation.
To not override an existing IMEX installation is enough to:
1. Create the files only if they do not exists (create_if_missing)
2. Enable and start IMEX  only if is is defined by our service unit file

At high level my point is: our goal is to not override an existing installation. 
The safest way to prevent it is to 1/ do not create files if they already exists and 2/ do not start a service if it's not the one installed by us. A condition on the existance of a directory is too weak. If the user creates that specific directory (even if it is empty)  IMEX may be broken by new compute nodes executing the config step.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
gmarciani: > The service file of IMEX for DLAMI is on a different location

This is an assumption that we are not in control of.

I suggest to create the file only if it is missing and avoid the only_if

himani2411: Its also an assumption that the service file exist in the location we want it to be.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: So what you are saying is
* Create Directory in Config Phase
* Then create the 2 config and 1 service fle 
* And start the Nvidia-imex
So esentially we will fall in the case where we support nvidia-imex for DLAMi?

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: The location i see if based on the vanilla DLAMI, its an assumption for now that this location will not chnage in future.
But if I add our own unit file at a particular location based on the check that it doesnt exist then I would fall in the weird scenario where i could potentially create 2 imex daemons on DLAMI.

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
himani2411: The creation of directory and the 2 imex conf files are not a problem per see and I can agree on moving those. 

The service file however is a problem as the condition would be difficult to handle

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
gmarciani: Aligned on this offline. Agreement: skip the whole `configure` action if the directory `node['cluster']['nvidia']['imex']['shared_dir']` does not exists AND the service file `"/etc/systemd/system/#{nvidia_imex_service}.service"` does not exist.

Define this condition into a self explanatory variable, e.g. `imex_installed_by_parallelcluster` and use such variable in the `only_if` blocks or , even better, at the beginning of the configure action.

PR #3038, opened 2025-10-23
[Performance] Remove cfn-hup from compute nodes as a temporary mitigation to restore good performance.

gmarciani:
**IMPORTANT NOTE** This change is not meant to be merged to develop or release branches. 
This is a temporary mitigation that will be vended as a custom cookbook to address a performance degradation. 
This change solves a major performance degradation, **but comes with the cost of removing the support for cluster updates.**

### Description of changes
Remove cfn-hup from compute nodes as a temporary mitigation to restore good performance.
Running cfn-hup on compute nodes can cause a major performance degradation on large clusters.
This change will be vended as a custom cookbook to mitigate the performance degradation.

* Describe *what* you're changing and *why* you're doing these changes.

### Tests
* Cluster creation succeeds
* Validated that this change is enough to fix a 20-30% performance degradation on a production weather forecasting workload on large cluster (198 nodes).


himani2411:
Can you mention with this custom cookbook disables the DFSMv2 feature in description

himani2411 (approved):

PR #3040, opened 2025-10-28
Add chef attribute `cluster/in_place_update_on_fleet_enabled` to disable cfn-hup on compute and login nodes and improve performance at scale

gmarciani:
### Description of changes
*This PR mitigates the performance degradation reported in https://github.com/aws/aws-parallelcluster/issues/6449*

Add new chef attribute `cluster/in_place_update_on_fleet_enabled` to disable cfn-hup on compute and login nodes in order to improve performance at scale.
When cfn-hup is disabled on compute and login nodes, the cluster readiness checks executed by the head node are disabled. 
Such attribute takes effect at config time (cluster creation/update, not at build image).

#### Q&A
1. **Why disabling cluster readiness check?**
Cluster readiness checks verify that all running compute/login nodes deployed the expected config version. If cfn-hup is disabled, they cannot apply the config version carried by the cluster update, so the check would always fail. 

1. **Why not calling the attribute in_place_update_on_fleet_disabled?**
We decided to name it `in_place_update_on_fleet_enabled` rather than `in_place_update_on_fleet_disabled` because of consistency (we use positive attributes in the rest of the cookbook) and maintainability (positive attributes are less error prone, e.g. double negations)

1. **Why disabling cfnhup on login nodes if the source of perf degradation are only compute nodes?**
To provide a consistent user experience and implementation. If we keep cfnhup in login nodes we would end up having login nodes supporting in-place updates and compute nodes not supporting it, ultimately leading to potential confusion and complexity.

1. **Why not testing the update of the new attribute?**
Updates to `ExtraChefAttributes` have never been supported as per update policy [here](https://github.com/aws/aws-parallelcluster/blob/14ce3d72dcf1c123630c475c58fff5f4b8610dad/cli/src/pcluster/schemas/common_schema.py#L219-L225)

### User Experience
By default the attribute is true, so cfn-hup is enabled on all cluster nodes.
When set to false, cfn-hup is disabled on both compute and login nodes. When this the case, the cluster readiness checks are disabled because w.o cfnhup compute/login nodes are not able to start an in-place update, so such checks would always fail.

#### [UseCase 1] in-place updates enabled
This is the default behavior, where cfn-hup is enabled on head node, compute nodes and login nodes. Being cfn-hup enabled, compute/login nodes are able to execute in-place updates, so the head node executes the usual cluster readiness check at the end of the update.

#### [UseCase 2] in-place updates disabled
cfn-hup is enabled on head node, but disabled on both compute nodes and login nodes. Being cfn-hup disabled, compute/login nodes are not able to execute in-place updates, so the head node does not execute the cluster readiness check at the end of the update.

### Tests
* Unit tests (Existing and new ones)
* Manually validated all the use cases reported in User Experience.


cookbooks/aws-parallelcluster-platform/templates/supervisord/parallelcluster_supervisord.conf.erb
```diff
@@ -1,8 +1,7 @@
 # Generated by Chef for AWS ParallelCluster <%= node['cluster']['node_type'] -%>
 # Local modifications could be overwritten.
 <%# HeadNode, ComputeFleet, LoginNode -%>
-<% case node['cluster']['node_type'] -%>
-<% when 'HeadNode', 'ComputeFleet', 'LoginNode' -%>
+<% if @cfnhup_enabled -%>
```
himani2411, line 4: This change makes the use of cfn-hup optional even for HeadNode, which is not the case we want.

cookbooks/aws-parallelcluster-shared/attributes/cluster.rb
```diff
@@ -34,3 +34,6 @@
 
 # Default NFS mount options
 default['cluster']['nfs']['hard_mount_options'] = 'hard,_netdev,noatime'
+
+# Cluster Updates
```
himani2411, line 38: Have we tested the use of this DevSettings with Cluster Create where we disable it? 

And what happens if we enable it on an update (so opposite of what the original idea is)

cookbooks/aws-parallelcluster-platform/templates/supervisord/parallelcluster_supervisord.conf.erb
himani2411: ignore, i saw the function next!

himani2411:
I would suggest that when you describe the user experience you refrain from mentioning `[UseCase 1] cfn-hup enabled` or `[UseCase 1] cfn-hup disabled` and explain what the use of cfn-hup is,  just like what cluster_readiness_check is explained

cookbooks/aws-parallelcluster-platform/templates/supervisord/parallelcluster_supervisord.conf.erb
gmarciani: Alright, for the records cfn_hup is always enabled for the HN an could be deisabled on compute and login nodes by setting the new attribute to False.

cookbooks/aws-parallelcluster-shared/attributes/cluster.rb
gmarciani: Updates are not supported for `DevSettings/Cookbook/ExtraChefAttribute` so no reason to test this use case.

See https://github.com/aws/aws-parallelcluster/blob/14ce3d72dcf1c123630c475c58fff5f4b8610dad/cli/src/pcluster/schemas/common_schema.py#L219-L225

Info added to PR description.

gmarciani:
> I would suggest that when you describe the user experience you refrain from mentioning [UseCase 1] cfn-hup enabled or [UseCase 1] cfn-hup disabled and explain what the use of cfn-hup is, just like what cluster_readiness_check is explained

Done, both here and in the PR for the CLI https://github.com/aws/aws-parallelcluster/pull/7071

himani2411 (approved):
LGTM!

PR #3070, opened 2025-12-17
Replace cfn-hup on compute nodes with systemd timers to signal updates

hgreebe:
### Description of changes
* Changes how compute nodes handle in place updates so that we no longer rely on cfn-hup running on compute nodes. 
* Replaces cfn-hup with a systemd timer that periodically checks a file in shared storage and runs an update if that file has been modified with the new cluster config version that has not yet been applied to the compute nodes. This file is updated by the head node when a cluster update occurs in order to signal to the compute nodes to update.
* The check-update.service has a 30 second timeout so that it does not run indefinately if something hangs.
* Revert changes for `in_place_update_on_fleet_enabled` from: https://github.com/aws/aws-parallelcluster-cookbook/commit/6eda378e1bcc060acf40df0f03cfcb89c60d3212#diff-6d6c58cce2dd575c0638ee245d9647b0dfa3cbdef86a136bd816d00538529fb4

### Tests
* Created unit tests to cover the systemd timer and service as well as new update logic
* Ran all the update integ tests

### References
* PR to remove validator test for `in_place_update_true` chef attribute: https://github.com/aws/aws-parallelcluster/pull/7156


cookbooks/aws-parallelcluster-computefleet/files/check_update/check-update.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Check file modification time every minute
```
gmarciani: The description must be more specific: it must be more clear that this is a timer configured by pcluster (add this info in the description and in the file name) and the goal of the timer (checks file modifications, which files, why?)

cookbooks/aws-parallelcluster-computefleet/files/check_update/check-update.timer
```diff
@@ -0,0 +1,11 @@
...
+
+[Timer]
+AccuracySec=1s
```
gmarciani: The accuracy has an impact on CPU wake-ups.
The finer the accuracy, the higher the chances to generate load on the CPU.
Since we are using a 60sec activation, what about using a 20s accuracy?

See https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html

cookbooks/aws-parallelcluster-computefleet/recipes/config/config_check_update_systemd_service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-slurm
+# Recipe:: config_compute
+#
+# Copyright:: 2013-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: Here and in other files: fix copyright year with `Copyright:: 2025 Amazon.com`

cookbooks/aws-parallelcluster-computefleet/recipes/config/config_check_update_systemd_service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-slurm
+# Recipe:: config_compute
```
gmarciani: Fix recipe name

cookbooks/aws-parallelcluster-computefleet/recipes/config/config_check_update_systemd_service.rb
```diff
@@ -0,0 +1,51 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+template '/etc/systemd/system/check-update.service' do
```
gmarciani: Here, above and below: the name of the service and the corresponding timer must be more talkative, expressing that they are pcluster services related to the cluster updates

cookbooks/aws-parallelcluster-platform/recipes/config/config_check_update_systemd_service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-slurm
```
gmarciani: Wrong cookbook. Curreently, the recipe i sin cookbook `aws-parallelcluster-computefleet`.

Also, what about moving this recipe to cookbook `aws-parallelcluster-platform` so that we are already in the direction of applying the same approach to login nodes?

cookbooks/aws-parallelcluster-computefleet/recipes/config.rb
```diff
@@ -12,3 +12,7 @@
 # limitations under the License.
 
 include_recipe 'aws-parallelcluster-computefleet::fleet_status'
+
+if ['ComputeFleet'].include?(node['cluster']['node_type'])
+  include_recipe 'aws-parallelcluster-computefleet::config_check_update_systemd_service'
```
gmarciani: See https://github.com/aws/aws-parallelcluster-cookbook/pull/3070/changes#r2641062088

cookbooks/aws-parallelcluster-computefleet/templates/check_update/check-update.service.erb
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=Check for recent file modifications
```
gmarciani: See https://github.com/aws/aws-parallelcluster-cookbook/pull/3070/changes#r2640842574

cookbooks/aws-parallelcluster-computefleet/templates/check_update/check-update.service.erb
```diff
@@ -0,0 +1,21 @@
...
+
+[Service]
+Type=oneshot
+TimeoutStartSec=30
+ExecStart=/bin/bash -c '\
```
gmarciani: To improve maintainability we  must move this logic into a dedicated script and have the unit file invoke it

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 3.15.0
 ------
 
+**CHANGES**
+- Replace cfn-hup in compute nodes with systemd timers to support in place updates.
```
gmarciani: We must surface the value of this change for the user, which is providing better performance.
Also we should surface that the new mechanism relies on shared storage sync between head node and compute fleet.

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -272,7 +280,7 @@ def update_nodes_in_queue(strategy, queues)
 
 chef_sleep '15'
 
-wait_cluster_ready if cluster_readiness_check_on_update_enabled?
+wait_cluster_ready
```
gmarciani: Why removing this? 
We should keep it as it may be useful to keep a way for the user to skip the readiness check if necessary.

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
```diff
@@ -106,14 +106,3 @@ def wait_sync_file(path)
...
 end
-
-def cfnhup_enabled?
-  # cfn-hup is always enabled on the head node, as it is required to perform cluster updates.
-  # cfn-hup can be disabled on compute nodes and login nodes, limiting the cluster update in the sense that
-  # live updates on compute and login nodes are not possible.
-  node['cluster']['node_type'] == 'HeadNode' || node['cluster']['in_place_update_on_fleet_enabled'] == 'true'
-end
-
-def cluster_readiness_check_on_update_enabled?
```
gmarciani, line 117: Same as https://github.com/aws/aws-parallelcluster-cookbook/pull/3070/changes#r2641117741

I think it is useful to keep a way to skip cluster readiness check through chef attributes. 
The current chef attribute name must be changed as `in_place_update_on_fleet_enabled` will not be  valid anymore. We could expose something like `cluster/update/cluster_readiness_check_enabled`

cookbooks/aws-parallelcluster-shared/attributes/cluster.rb
```diff
@@ -34,6 +38,3 @@
 
 # Default NFS mount options
 default['cluster']['nfs']['hard_mount_options'] = 'hard,_netdev,noatime'
-
-# Cluster Updates
-default['cluster']['in_place_update_on_fleet_enabled'] = 'true'
```
gmarciani, line 39: See https://github.com/aws/aws-parallelcluster-cookbook/pull/3070/changes#r2641119272

cookbooks/aws-parallelcluster-shared/attributes/cluster.rb
```diff
@@ -9,6 +9,10 @@
 default['cluster']['log_base_dir'] = '/var/log/parallelcluster'
 default['cluster']['etc_dir'] = '/etc/parallelcluster'
 
+# Shared file used to manage inplace updates
+default['cluster']['shared_update_path'] = "#{node['cluster']['shared_dir']}/check_update"
+default['cluster']['update_checkpoint'] = "#{node['cluster']['scripts_dir']}/update_checkpoint"
```
gmarciani: I  suggest to use more talkative names, such as

```
default['cluster']['update']['trigger_file'] = "#{node['cluster']['shared_dir']}/update_trigger"
default['cluster']['update']['checkpoint_file'] = "#{node['cluster']['scripts_dir']}/update_checkpoint"
```

With such names for the attributes we better convey that:
1. they are files (when an attribute contains a directory path, it has `dir` as suffix)
2. one is used as a trigger

cookbooks/aws-parallelcluster-computefleet/templates/check_update/check-update.service.erb
```diff
@@ -0,0 +1,21 @@
...
+\
+CURRENT_UPDATE=$(cat "$SHARED_FILE") || exit 0; \
+LAST_APPLIED=$([ -f "$LOCAL_CHECKPOINT" ] && cat "$LOCAL_CHECKPOINT" || echo ""); \
+\
+if [ "$CURRENT_UPDATE" != "$LAST_APPLIED" ]; then \
+echo "$CURRENT_UPDATE" > "$LOCAL_CHECKPOINT" && <%= node['cluster']['scripts_dir'] %>/cfn-hup-update-action.sh; \
+fi'
+
+[Install]
+WantedBy=multi-user.target
```
gmarciani: This service is triggered by a timer. What is the point of having `WantedBy=multi-user.target`?
I think it could be removed as it is actually unused.

cookbooks/aws-parallelcluster-computefleet/templates/check_update/check-update.service.erb
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=Check for recent file modifications
+
+[Service]
```
gmarciani: We are missing logs for the logic executed by this service.
I suggest to set the property [StandardOutput](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#StandardOutput=) to log to a local log file. We must then push such file to cloudwatch.

Also, the logic should log more information:
1. when it starts
2. if the file exists
3. what is the content of the checkpoint file
4. what is the content of the trigger file
5. the decision taken
6. when it completes

All log lines must be timestamped with millis resolution.

cookbooks/aws-parallelcluster-computefleet/templates/check_update/check-update.service.erb
```diff
@@ -0,0 +1,21 @@
...
+Type=oneshot
+TimeoutStartSec=30
```
gmarciani: I suppose this timeout is here to prevent the service being stuck in case of file system unresponsiveness.
This is a good strategy as it is important to put a timeout when dealing with remote resources.
However, I think `TimeoutStartSec` may not be the right parameter to achieve this.

According to the official documentation for [TimeoutStartSec](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#TimeoutStartSec=):

> If a daemon service does not signal start-up completion within the configured time, the service will be considered failed and will be shut down again.

So the timeout covers the whole start logic. The start logic here includes the execution of `cfn-hup-update-action.sh`, which can legitimately last longer than 30 seconds. So with the current approach we have the risk of interrupting legitimate updates.

If you agree with this risk, I suggest to configure the timeout logic differently:
1. define a timeout of 20 seconds to read the shared files (20s is enough to capture file system failures with a simple read operation on a single file)
2. define a timeout for the update action, which must be set through the node_bootstrap_timeout parameter, as it currently is. See https://github.com/gmarciani/aws-parallelcluster-cookbook/blob/wip/mgiacomo/3141/fix-build-ubhu-1218-1/cookbooks/aws-parallelcluster-environment/templates/cfn_hup_configuration/cfn-hook-update.conf.erb#L9-L9

cookbooks/aws-parallelcluster-platform/recipes/config/config_check_update_systemd_service.rb
```diff
@@ -0,0 +1,58 @@
...
+  mode '0644'
+  action :create
+end
+
+file node['cluster']['update']['trigger_file'] do
+  content ''
+  owner 'root'
+  group 'root'
+  mode '0644'
+  action :create_if_missing
```
gmarciani, line 45: Why this file has action `create_if_missing` while the other files have `create`?

cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+
+# This script checks if a cluster update is needed by comparing the shared update file
+# with the local checkpoint. If they differ, it triggers the update action.
+
+set -ex
+
+SHARED_FILE="<%= node['cluster']['update']['trigger_file'] %>"
```
gmarciani: minor: SHARED_FILE is too generic. /Can we use a more talkative variable name, such as SHAED_TRIGGER_FILE

cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
```diff
@@ -0,0 +1,59 @@
...
+    log "Completed pcluster-check-update - no action needed"
+    exit 0
+fi
+log "Shared update file exists: $SHARED_FILE"
+
+# Read current update version from shared file
+CURRENT_UPDATE=$(timeout $FS_TIMEOUT cat "$SHARED_FILE" 2>/dev/null) || {
+    log "Failed to read shared update file: $SHARED_FILE"
+    log "Completed pcluster-check-update - no action needed"
+    exit 0
```
gmarciani: If we fail to read the file, why we returns exit code 0 rather than an error exit code?
What are the reasons where we could fail to read this and how we could defent from systemic issues here?
I'm thinking about corruption of permissions. In that case it could be useful for4 troubleshooting to add a log line with the output of `ls -l $SHARED_FILE`

cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
```diff
@@ -0,0 +1,59 @@
...
+
+# Check if shared file exists
+if ! timeout $FS_TIMEOUT test -f "$SHARED_FILE"; then
+    log "Shared update file does not exist: $SHARED_FILE"
+    log "Completed pcluster-check-update - no action needed"
+    exit 0
+fi
+log "Shared update file exists: $SHARED_FILE"
+
+# Read current update version from shared file
```
gmarciani: by `update version` here I suppose you meant the cluster config version. Can you please specify this in the comments whenever you refer to this version?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 ------
 
 **CHANGES**
+- Replace cfn-hup in compute nodes with systemd timer to support in place updates in order to improve performance.
+  - This new mechanism relies on shared storage to sync updates between the head node and compute nodes.
```
gmarciani: Extra `-` on the second item

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
```diff
@@ -107,15 +107,8 @@ def wait_sync_file(path)
...
 
-def cfnhup_enabled?
-  # cfn-hup is always enabled on the head node, as it is required to perform cluster updates.
-  # cfn-hup can be disabled on compute nodes and login nodes, limiting the cluster update in the sense that
-  # live updates on compute and login nodes are not possible.
-  node['cluster']['node_type'] == 'HeadNode' || node['cluster']['in_place_update_on_fleet_enabled'] == 'true'
-end
-
 def cluster_readiness_check_on_update_enabled?
-  node['cluster']['in_place_update_on_fleet_enabled'] == 'true'
```
gmarciani, line 118: We introduced this devsettings to disable an update mechanism that was causing issues.
We are confident that the systemd based approach will not, but what about keeping this devsettings to disable the mechanism?

If at the end we decide to remove it, then we should mention its deprecation in our changelog. Even if it is a DevSetting, we documented its introduction in the changelog of 3.14.1 so we should do the same if we remove it as some customers may be using it.

cookbooks/aws-parallelcluster-platform/recipes/config/config_check_update_systemd_service.rb
hgreebe: because this is a shared file, so once it is created, it doesn't need to be recreated by other compute nodes

hanwen-cluster:
Great PR!

cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
```diff
@@ -0,0 +1,60 @@
...
+if ! timeout $FS_TIMEOUT test -f "$SHARED_TRIGGER_FILE"; then
+    log "Shared update file does not exist: $SHARED_TRIGGER_FILE"
+    log "Completed pcluster-check-update - no action needed"
+    exit 0
+fi
+log "Shared update file exists: $SHARED_TRIGGER_FILE"
+
+# Read current cluster config version from shared file
+CURRENT_UPDATE=$(timeout $FS_TIMEOUT cat "$SHARED_TRIGGER_FILE" 2>/dev/null) || {
+    log "Failed to read shared update file: $SHARED_TRIGGER_FILE"
```
gmarciani: Can we prepend a [ERROR] to log lines related to failures? This would facilitate the reading of logs when troubleshooting issues

gmarciani (approved):

PR #3088, opened 2026-01-08
[Test] Add retry utility for kitchen tests and use it to mitigate failure on flaky test

gmarciani:
### Description of changes
Add retry utility for kitchen tests and use it to mitigate failure on flaky test which sometimes fails due to  to timeout on yum repolist command.

Notes:
1. Skipping the `bad-url-suffix-check` because it fails on false positive: an URL used for kitchen tests, which is expected to have the commercial URL suffix.

### Tests
* Kitchen test succeeded.


himani2411 (approved):

himani2411:
LGTM

PR #3011, opened 2025-08-16
[Isolated] Update Pypi dependencies and install efs-proxy dependency

hgreebe:
### Description of changes
* Adding PyPi dependencies for latest Python version 3.12 and using appropriate folder names
* Installing Packages with `--no-build-isolation` as with Python 3.12.x so that we check for existing dependecies.
   ** packages are installed with isolation from Py 3.12, where they do not check if the dependencies they want already exist and try to install them again.
* Installing efs-proxy dependecies from s3 in Isolated region as they cannot connect to cargo and get below error
 ```
 Stdout: failed to download from `https://index.crates.io/config.json
Stdout: [6] Couldn't resolve host name (Could not resolve host: index.crates.io)
```

* adding GIT_REF for cookbook util script for uploading it in S3

### Tests
* Build rhel8, AL2, and AL2023 AMIs in ADC
* Ran isolated region integ tests

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3095, opened 2026-01-20
[Python 3.14][Deps] Upgrade Python to 3.14 for all OS except AL2

himani2411:
### Description of changes
* Upgrade Python to 3.14 for all OS except AL2
     * we do not upgrade Python for AL2 as there is a transitive dependency OpenSSL 1.1.1 which is required for latest python versions and its not available in AL2 
* Added extra Chef Attribute `install_python_from_internet` which is used for installing python and other Pypi dependencies from Internet when true; by default the installation is done using s3.
* IMPACT for adding `install_python_from_internet`
    - Quicker code changes and testing when we upgrade since the commits with TEMP  are the changes I have to do manually every time I want to upgrade python and pypi dependencies 
    - In case of Future CVE's a quick mitigation for CX can be provided using this attribute
    - *The `[TEMP] Hardcode` commits will be will be removed from this PR before it is to be merged. These commits are example of the changes that need to be done for initial testing*

* Add unit tests

### Tests
* Unit test 
* BUILD AMI for all OS except AL2
* Integ Tests 
```
test-suites:
  health_checks:
    test_gpu_health_checks.py::test_cluster_with_gpu_health_checks:
      dimensions:
        - regions: [{{ US_WEST_2_GPU_INSTANCE_TYPE_0_CAPACITY_RESERVATION_3_INSTANCES_2_HOURS_NOPG_OS_X86_5 }}]
          instances: [{{ US_WEST_2_GPU_INSTANCE_TYPE_0 }}]
          oss: ["alinux2023"]
          schedulers: ["slurm"]
  basic:
    test_essential_features.py::test_essential_features:
      dimensions:
        - regions: [{{ US_EAST_1_INSTANCE_TYPE_0_xlarge_CAPACITY_RESERVATION_8_INSTANCES_2_HOURS_NOPG_OS_X86_1 }}]
          instances: [{{ US_EAST_1_INSTANCE_TYPE_0 }}.xlarge]
          oss: ["rocky8"]
          schedulers: ["slurm"]
```

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-computefleet/kitchen.computefleet-config.yml
+++ b/cookbooks/aws-parallelcluster-computefleet/kitchen.computefleet-config.yml
@@ -31,7 +31,7 @@ suites:
     attributes:
       cluster:
         custom_node_package: https://github.com/aws/aws-parallelcluster-node/archive/develop.tar.gz
-        python-version: 3.12.11
+        python-version: 3.14.2
         node_virtualenv_path: /opt/parallelcluster/pyenv/versions/node_virtualenv
   - name: fleet_status
     run_list:
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
@@ -19,7 +19,9 @@
 
 # TODO: once the pyenv Chef resource supports installing packages from a path (e.g. `pip install .`), convert the
 # bash block to a recipe that uses the pyenv resource.
-command = "pip install . --no-build-isolation"
+# Use --no-build-isolation when installing from S3 pre-built deps, omit when installing from internet
+pip_install_flags = node['cluster']['install_python_from_internet'] ? "" : "--no-build-isolation"
+command = "pip install . #{pip_install_flags}".strip
 
 dependency_package_name = "pypi-node-dependencies-#{node['cluster']['python-major-minor-version']}-#{node['kernel']['machine']}"
 dependency_folder_name = dependency_package_name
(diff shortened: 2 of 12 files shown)


cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
```diff
@@ -28,24 +30,28 @@
...
-bash 'pip install' do
-  user 'root'
-  group 'root'
-  cwd "#{node['cluster']['base_dir']}"
-  code <<-REQ
-    set -e
-    tar xzf node-dependencies.tgz
-    cd #{dependency_folder_name}
-    #{node_virtualenv_path}/bin/pip install * -f ./ --no-index
+  bash 'pip install node dependencies from S3' do
```
gmarciani, line 44: Where is the pip install command to install python packages from internet?

cookbooks/aws-parallelcluster-computefleet/kitchen.computefleet-config.yml
```diff
@@ -31,7 +31,7 @@ suites:
     attributes:
       cluster:
         custom_node_package: https://github.com/aws/aws-parallelcluster-node/archive/develop.tar.gz
-        python-version: 3.12.11
+        python-version: 3.14.2
```
gmarciani, line 34: We know that the pcluster CLI has some incompatibilities with Python 3.14.
We do not use the CLI in the cookbook, however did you make sure that all the python packages we use in the cookbook are compatible with Python 3.14, e.g. pcluster daemons?

cookbooks/aws-parallelcluster-platform/recipes/install/cookbook_virtualenv.rb
```diff
@@ -33,22 +33,42 @@
...
-  tar xzf cookbook-dependencies.tgz
-  cd #{dependency_package_name}
-  #{virtualenv_path}/bin/pip install * -f ./ --no-index
-  REQ
+  bash 'pip install cookbook dependencies from internet' do
+    user 'root'
+    group 'root'
+    code <<-REQ
+      set -e
+      #{virtualenv_path}/bin/pip install -r #{node['cluster']['base_dir']}/cookbook-requirements.txt
```
gmarciani: [minor] the requirements path `node['cluster']['base_dir']}/cookbook-requirements.txt` is duplicated, can we use a constant instead?

cookbooks/aws-parallelcluster-shared/resources/install_pyenv.rb
```diff
@@ -5,7 +5,7 @@
 
 # Resource:: to create a Python virtual environment for a given user
 
-property :python_version, String
+property :python_version, String, default: '3.14.2'
```
gmarciani: Why not keeping the default from the attribute `['cluster']['python-version']`?

gmarciani:
Can you please surface this upgrade in the changelog?

cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
himani2411: The line 22 `pip_install_flags = node['cluster']['install_python_from_internet'] ? "" : "--no-build-isolation"` is enough for handling the dependencies installation!

`--no-build-isolation` flag is being used when we want to take care of the Pypi dependencies by ourselves and this flag is an indicator to Pip to check if the dependencies we want exist in the environment before looking at the internet. Some ref and the PR where we added it https://github.com/aws/aws-parallelcluster-cookbook/pull/3011
```
** packages are installed with isolation from PyPi, where they do not check if the dependencies they want already exist and try to install them again.
```
By default pip will try to install dependencies (even if they exist) from internet in an isolated environment. Since python's default behavior takes care of thisI just changed the code to remove the installation from S3

cookbooks/aws-parallelcluster-computefleet/kitchen.computefleet-config.yml
himani2411: I checked by running very limited number of  integ tests which passed successfully. I dont see why there should be a gap for pcluster daemons when the node package installation was successful and the cluster creation with essential feature were successful.

cookbooks/aws-parallelcluster-shared/resources/install_pyenv.rb
himani2411: THIS is part of the `[TEMP] Hardcode` commit which is to be remove before we merge.

The use of your suggested approach results in `undefined local variable or method 'node'` 

We can use the `['python-version']` inside the resource as seem in line 16.
*  I would be in favor of removing this property as there is no way of overriding this during runtime
* But in future there could be a possibility that we would differ in the python version for Cookbook or cfn_bootstrap and this resource property can be useful, which is why added the current workaround

himani2411:
> Can you please surface this upgrade in the changelog?

Updated

cookbooks/aws-parallelcluster-computefleet/kitchen.computefleet-config.yml
gmarciani: I was thinking about runtime issues, more than installation issues. 
On the CLI with Py3.14 we were able to install, but some commands failed. 
I suggest to verify the logs in the clusters used for the test just to verify there are no runtime errors.

himani2411: Checked in the logs manually and do not see any error's

gmarciani (approved):

PR #3102, opened 2026-01-26
[Update] Reduce the risk of clustermgtd being stopped on cluster update

gmarciani:
### Description of changes
Reduce the risk of clustermgtd being stopped on cluster update
We keep clustermgtd stopped only during update actions that have high risk of race conditions, that are munge key updates, slurm accounting updates, restart of slurmctld and scontrol reconfigure. Also, in case of update failure, we restart clustermgtd whatever the point of failure faced during the update.

### Q&A
2. **How will we cover this change with integ tests?**
We will introduce/extend integ tests to verify no race conditions are triggered as part of the next sprint. 
For sake of this PR we consider enough to verify that it introduces no regressions.
3. **Why keeping clustermgtd stopped during slurm accounting updates?**
This is not required because of conflicts with clustermgtd, but because the slurm accounting updates includes the restart of slurmdbd, that, in case munge key is updated, must occur after the munge key updates ([docs](https://slurm.schedmd.com/quickstart_admin.html)).

### Tests
* The risks have been deeply assessed during design.
* Unit tests (updated to cover the current changes)
* [PENDING] Integ tests: 
  * `test_update_slurm`
  * `test_custom_munge_key`
  * `test_dynamic_file_systems_update_rollback`
  * `test_update_rollback_failure`
  *  `test_login_nodes_update`
  * `test_queue_parameters_update`
  * `test_dynamic_file_systems_update`

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
@@ -207,17 +207,6 @@ replace_or_add "update node replacement timeout" do
   replace_only true
 end
 
-ruby_block "Update Slurm Accounting" do
-  block do
-    if node['cluster']['config'].dig(:Scheduling, :SlurmSettings, :Database).nil?
-      run_context.include_recipe "aws-parallelcluster-slurm::clear_slurm_accounting"
-    else
-      run_context.include_recipe "aws-parallelcluster-slurm::config_slurm_accounting"
-    end
-  end
-  only_if { ::File.exist?(node['cluster']['previous_cluster_config_path']) && is_slurm_database_updated? }
-end unless on_docker?
-
 # Cover the following two scenarios:
 # - a cluster without login nodes is updated to have login nodes;
 # - a cluster with login nodes is updated to use another pool name.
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -255,6 +248,13 @@ def update_nodes_in_queue(strategy, queues)
   only_if { !(::File.exist?(node['cluster']['previous_cluster_config_path']) && !are_queues_updated?) && !node['cluster']['config'].dig(:Scheduling, :SlurmSettings, :Database).nil? }
 end
 
+execute 'stop clustermgtd' do
+  command "#{cookbook_virtualenv_path}/bin/supervisorctl stop clustermgtd"
+  not_if { ::File.exist?(node['cluster']['previous_cluster_config_path']) && !are_queues_updated? && !are_bulk_custom_slurm_settings_updated? }
+end
+
+update_munge_head_node
```
himani2411, line 245: Can we run slurmdbd integ tests, just to verify that moving the Munge Update after it does not have any effect on the slurmdbd daemon?

https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-slurm/templates/default/slurm/head_node/update_slurm_database_password.sh.erb#L47

himani2411: Reason to test is as per the doc https://slurm.schedmd.com/quickstart_admin.html
>Make sure the MUNGE daemon, munged, is started before you start the Slurm daemons.

gmarciani: Good point, doing

gmarciani: Good point, executed, succeeded.

gmarciani: Discussed offline and agreed that it is safer to execute the slurm accounting updates after the munge key update. The current integ tests is not able to detect potential issues because it is executing the update of munge key without updates of slurm db settings.

gmarciani: Done

PR #3122, opened 2026-02-13
[Workaround Only, Will Revert] Add Topology parameter to NodeName lines for block topology nodes

hehe7318:
### Description of changes
Workaround for Slurm 25.11.2 bug where topology_p_add_rm_node incorrectly removes nodes from topology blocks when they enter power-down state. By setting `Topology=default:<BlockName>` on NodeName lines, the node's `topology_str` is no longer NULL, preventing the buggy code path.

SchedMD has confirmed this is a known issue with a fix in review.

## Changes we are doing

The change in this PR is to read the topology.conf that ParallelCluster creates and stores at `/opt/slurm/etc/` and parse it to get Block Mappings.
So if this file exists we read it to find the BlockName so that we can add `Topology=default:<BlockName>` to the slurm node definition

## Why not create the topology.conf through pcluster_slurm_config_generator.py or merge the configuration in file?
This is a temporary workaround for the bug that we found with slurm and we do not change the generation of separate topology.conf file since in future when we want to expand to use different topology plugin we can easily expand the existing resource we have in our cookbook.

### Tests
* Running

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
+++ b/cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
@@ -40,6 +40,60 @@ class CriticalError(Exception):
     pass
 
 
+def _build_topology_block_mapping(slurm_conf_dir):
+    """
+    Build a mapping of (queue_name, compute_resource_name) -> block_name from topology.conf.
+
+    Returns an empty dict if topology.conf does not exist.
+    """
+    topology_conf_path = path.join(slurm_conf_dir, "topology.conf")
+    mapping = {}
+
+    if not path.exists(topology_conf_path):
+        return mapping
+
+    try:
+        with open(topology_conf_path, encoding="utf-8") as f:
+            for line in f:
+                line = line.strip()
+                if not line or line.startswith("#"):
+                    continue
+                if not line.startswith("BlockName="):
+                    continue
+
+                block_name = None
+                nodes_spec = None
+                for token in line.split():
+                    if token.startswith("BlockName="):
+                        block_name = token.split("=", 1)[1]
+                    elif token.startswith("Nodes="):
+                        nodes_spec = token.split("=", 1)[1]
+
+                if not block_name or not nodes_spec:
+                    continue
+
+                # Extract queue_name and compute_resource_name from node spec
+                # Node spec format: {queue_name}-st-{compute_resource_name}-[1-N]
+                #                or {queue_name}-dy-{compute_resource_name}-[1-N]
+                # Remove the trailing -[1-N] part
+                node_prefix = re.sub(r"-\[.*\]$", "", nodes_spec)
+                # Split by "-st-" or "-dy-" to get queue_name and compute_resource_name
+                parts = re.split(r"-(st|dy)-", node_prefix, maxsplit=1)
+                if len(parts) == 3:
+                    queue_name, _node_type, compute_resource_name = parts
+                    mapping[(queue_name, compute_resource_name)] = block_name
+                    log.info(
+                        "Topology mapping: (%s, %s) -> %s",
+                        queue_name,
+                        compute_resource_name,
+                        block_name,
+                    )
+    except Exception as e:
+        log.warning("Failed to parse topology.conf for block mapping: %s", e)
+
+    return mapping
+
+
 def generate_slurm_config_files(
     output_directory,
     template_directory,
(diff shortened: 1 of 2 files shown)


himani2411:
Added unit tests in https://github.com/aws/aws-parallelcluster-cookbook/pull/3123

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -40,6 +40,60 @@ class CriticalError(Exception):
...
+    if not path.exists(topology_conf_path):
+        return mapping
+
+    try:
+        with open(topology_conf_path, encoding="utf-8") as f:
+            for line in f:
+                line = line.strip()
+                if not line or line.startswith("#"):
+                    continue
+                if not line.startswith("BlockName="):
```
gmarciani, line 61: What about lines starting with `BlockSizes` or `Nodes`?
REf https://slurm.schedmd.com/topology.conf.html

Instead of skipping lines that do not start with `BlockName=`, we should instead consider all the lines containing both `BlockName=` and `Nodes=` because ultimately those are the info we want to parse to build the mapping

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -40,6 +40,60 @@ class CriticalError(Exception):
...
+                    continue
+
+                block_name = None
+                nodes_spec = None
+                for token in line.split():
```
gmarciani, line 66: [non blocking] Why looping+splitting rather than a regex to detect the values?

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -40,6 +40,60 @@ class CriticalError(Exception):
...
+                        nodes_spec = token.split("=", 1)[1]
+
+                if not block_name or not nodes_spec:
+                    continue
+
+                # Extract queue_name and compute_resource_name from node spec
+                # Node spec format: {queue_name}-st-{compute_resource_name}-[1-N]
+                #                or {queue_name}-dy-{compute_resource_name}-[1-N]
+                # Remove the trailing -[1-N] part
+                node_prefix = re.sub(r"-\[.*\]$", "", nodes_spec)
```
gmarciani, line 79: Is this logic able to handle the case of queue name and cr name containing `-st|dy-`?

The extraction of queue name and cr name from a nodename deserves a dedicated function with a dedicated unit test to validate the robustness of the regex in parsing edge cases.

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
himani2411: As per the slurm documentation you referred, the lines should start with BlockName or BlockSizes and they should be on separate lines.

1. IF the line start with Nodes it will fail slurm validation with an error `error: _parse_next_key: Parsing error at unrecognized key: Nodes`
2. And if the line start with BlockSizes with other details in the same line it gives error `error: Parse error in file /opt/slurm/etc/topology.conf line 3: "  BlockName=Block1  Nodes=q1-st-cr1-[1-2]"`

cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
himani2411: Will add the unit test for verifying this use-case, anyway might have to do these changes for code-linters too as the function is too complex

himani2411: However the use of `dy` is an overkill since Block topology we use is only with static nodes/Capacity blocks and only for p6e-gb200 instance.

himani2411: we will handle the workaround + unit tests in https://github.com/aws/aws-parallelcluster-cookbook/pull/3123 instead of using the fragile regex-based topology.conf parsing in pcluster_slurm_config_generator.py which was the old approach in this PR.

himani2411:
Closing as fix from slurm is in v25.11.3 https://github.com/SchedMD/slurm/releases/tag/slurm-25-11-3-1

PR #3123, opened 2026-02-16
[Workaround] Add `Topology=default:<BlockName>` to node definition to avoid nodes being dropped from topology blocks when they enter power-down state. 

himani2411:
# Description of changes

### Context
Slurm 25.11 regression: topology_p_add_rm_node removes nodes from topology blocks on power-down but never re-adds them on power-up. Jobs with --exclusive=topo fail until manual scontrol reconfigure. Confirmed by SchedMD as known bug with fix in review.

Per SchedMD recommendation, set Topology=default:<BlockName> on NodeName lines in the partition config to prevent nodes from being removed. Ref: https://slurm.schedmd.com/slurm.conf.html#OPT_Topology

## Changes we are doing
* We are adding the below changes instead of  https://github.com/aws/aws-parallelcluster-cookbook/pull/3122 which was taking a Regex based approach.
* Extracted build_topology_block_mapping() in pcluster_topology_generator.py as single source of truth for block assignment logic
* Added --block-sizes and --force-topology-configuration CLI args to pcluster_slurm_config_generator.py; passed from Chef recipes (config_head_node.rb, update_head_node.rb)
* Updated unit tests and expected outputs

## TODO
Revert after SchedMD releases the fix.

# FAQ
## Why not create the topology.conf through pcluster_slurm_config_generator.py or merge the configuration in file?
This is a temporary workaround for the bug that we found with slurm and we do not change the generation of separate topology.conf file since in future when we want to expand to use different topology plugin we can easily expand the existing resource we have in our cookbook.

## TESTING
* Unit tests
* [ONGOING] test_gb200 integ test

### References

### Checklist


cookbooks/aws-parallelcluster-slurm/files/default/head_node_slurm/slurm/pcluster_slurm_config_generator.py
```diff
@@ -40,6 +40,35 @@ class CriticalError(Exception):
     pass
 
 
+def _build_topology_block_mapping(cluster_config, block_sizes, force_topology_configuration):
```
hgreebe, line 43: why do we need this separate function if we already have `build_topology_block_mapping`? Don't they both have the same return value?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -71,7 +71,9 @@
             " --compute-node-bootstrap-timeout #{node['cluster']['compute_node_bootstrap_timeout']} #{no_gpu}"\
             " --realmemory-to-ec2memory-ratio #{node['cluster']['realmemory_to_ec2memory_ratio']}"\
             " --slurmdbd-user #{node['cluster']['slurm']['user']}"\
-            " --cluster-name #{node['cluster']['stack_name']}"
+            " --cluster-name #{node['cluster']['stack_name']}"\
+            "#{node['cluster']['p6egb200_block_sizes'].nil? ? '' : " --block-sizes #{node['cluster']['p6egb200_block_sizes']}"}"\
```
hgreebe, line 75: what is the `node['cluster']['p6egb200_block_sizes']` default and where is it coming from? Also for `node['cluster']['slurm']['block_topology']['force_configuration']`?

himani2411:
Closing since the Fix from Slurm is in v25.11.3 https://github.com/SchedMD/slurm/releases/tag/slurm-25-11-3-1

PR #3132, opened 2026-03-03
Split slurm accounting configuration into two parts 

hgreebe:
### Description of changes

* Slurm 25.11+ includes a [fix](https://github.com/SchedMD/slurm/commit/6086f2c3) that allows slurmctld to preserve its ClusterID when registering with slurmdbd. This should resolve cluster ID mismatch issues when enabling accounting on an existing cluster.
* Despite the Slurm fix, preserving the `/var/spool/slurm.state/clustername` file (which contains the ClusterID) still resulted in a cluster ID mismatch error. 
* The bootstrap script used `sacctmgr add cluster`, which always assigns a new ClusterID and cannot accept the existing ID as a parameter. This created a mismatch between:
  - slurmctld's ClusterID (from the state file)
  - slurmdbd's ClusterID (newly generated by sacctmgr)
* To solve this, I removed `sacctmgr add cluster` and rely on slurmctld's automatic registration to preserve the ClusterID.
* This created a new problem: the bootstrap script ran before slurmctld could register the cluster, leaving no cluster in the database to add users to.
* To fix this, I split Slurm accounting configuration into two phases:

  1. **Configure and start slurmdbd** - prepares the accounting database service
  2. **Restart slurmctld** - automatically registers the cluster with preserved ClusterID
  3. **Bootstrap users** - adds accounts and users to the now-registered cluster

This ordering ensures slurmctld registers the cluster (with correct ClusterID) before the bootstrap script attempts to add users.

### Tests
* Ran `test_slurm_accounting`, `test_slurm_accounting_external_dbd`, and `test_slurm`
* I tested that 2 clusters can use the same db
* I tested that a cluster can be updated to use a different db

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
@@ -19,7 +19,7 @@ execute "wait for cluster registration" do
   command "#{node['cluster']['slurm']['install_dir']}/bin/sacctmgr show clusters -Pn cluster=#{node['cluster']['stack_name']} format=cluster | grep -x '#{node['cluster']['stack_name']}'"
   retries 30
   retry_delay 10
-end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
+end
 
 bash "bootstrap slurm database" do
   user 'root'
@@ -44,4 +44,4 @@ bash "bootstrap slurm database" do
 
     exit 0
   BOOTSTRAP
-end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
+end
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-slurm
+# Recipe:: bootstrap_slurm_accounting
+#
+# Copyright:: 2013-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+execute "wait for cluster registration" do
+  command "#{node['cluster']['slurm']['install_dir']}/bin/sacctmgr show clusters -Pn cluster=#{node['cluster']['stack_name']} format=cluster | grep -x '#{node['cluster']['stack_name']}'"
+  retries 30
+  retry_delay 10
+end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
+
+bash "bootstrap slurm database" do
+  user 'root'
+  group 'root'
+  code <<-BOOTSTRAP
+    SACCTMGR_CMD=#{node['cluster']['slurm']['install_dir']}/bin/sacctmgr
+    CLUSTER_NAME=#{node['cluster']['stack_name']}
+    DEF_ACCOUNT=pcdefault
+    SLURM_USER=#{node['cluster']['slurm']['user']}
+    DEF_USER=#{node['cluster']['cluster_user']}
+
+    # Add account-cluster association to database if it is not present yet
+    [[ $($SACCTMGR_CMD list associations -Pn cluster=$CLUSTER_NAME account=$DEF_ACCOUNT format=account | grep $DEF_ACCOUNT) ]] || \
+        $SACCTMGR_CMD -iQ add account $DEF_ACCOUNT Cluster=$CLUSTER_NAME \
+            Description="ParallelCluster default account" Organization="none"
+
+    # Add user-account associations to database if they are not present yet
+    [[ $($SACCTMGR_CMD list associations -Pn cluster=$CLUSTER_NAME account=$DEF_ACCOUNT user=$SLURM_USER format=user | grep $SLURM_USER) ]] || \
+        $SACCTMGR_CMD -iQ add user $SLURM_USER Account=$DEF_ACCOUNT AdminLevel=Admin
+    [[ $($SACCTMGR_CMD list associations -Pn cluster=$CLUSTER_NAME account=$DEF_ACCOUNT user=$DEF_USER format=user | grep $DEF_USER) ]] || \
+        $SACCTMGR_CMD -iQ add user $DEF_USER Account=$DEF_ACCOUNT AdminLevel=Admin
+
+    exit 0
+  BOOTSTRAP
+end unless kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
@@ -4,7 +4,7 @@
 # Cookbook:: aws-parallelcluster-slurm
 # Recipe:: bootstrap_slurm_accounting
 #
-# Copyright:: 2013-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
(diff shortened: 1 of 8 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
```diff
@@ -221,3 +221,11 @@
...
   retry_delay 2
 end unless redhat_on_docker?
+
+ruby_block "Bootstrap Slurm Accounting Users" do
+  block do
+    run_context.include_recipe "aws-parallelcluster-slurm::bootstrap_slurm_accounting"
+  end
+  not_if { node['cluster']['config'].dig(:Scheduling, :SlurmSettings, :Database).nil? }
+  not_if { kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd") }
+end unless on_docker?
```
gmarciani: Why `on_docker?` not specified within the `not_if` block?

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
```diff
@@ -0,0 +1,47 @@
...
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
```
gmarciani, line 17: In recipe `config_slurm_accounting.rb` we used to have an explicit skip on `kitchen_test? OR node['cluster']['node_type'] == "ExternalSlurmDbd"`. Why they were required in the original recipe but not here?

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+#
+# Cookbook:: aws-parallelcluster-slurm
+# Recipe:: bootstrap_slurm_accounting
+#
+# Copyright:: 2013-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gmarciani: [nit] Fix copyright year

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
```diff
@@ -0,0 +1,47 @@
...
+  code <<-BOOTSTRAP
+    SACCTMGR_CMD=#{node['cluster']['slurm']['install_dir']}/bin/sacctmgr
+    CLUSTER_NAME=#{node['cluster']['stack_name']}
+    DEF_ACCOUNT=pcdefault
+    SLURM_USER=#{node['cluster']['slurm']['user']}
+    DEF_USER=#{node['cluster']['cluster_user']}
+
+    # Add account-cluster association to database if it is not present yet
+    [[ $($SACCTMGR_CMD list associations -Pn cluster=$CLUSTER_NAME account=$DEF_ACCOUNT format=account | grep $DEF_ACCOUNT) ]] || \
+        $SACCTMGR_CMD -iQ add account $DEF_ACCOUNT Cluster=$CLUSTER_NAME \
```
gmarciani, line 38: The sacctmgtd commands here assume the cluster was created on SlurmDB. We used to be sure about that because we were creating the cluster in the database before running those commands.
What ensures that the cluster was created in the database? Is it something we are in control of? Is it something that could lead to race conditions because we are not in control of the time the cluster is created?

gmarciani: nvm: we wait for the cluster registration in `wait for cluster registration above`

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
hgreebe: Those i added [here](https://github.com/aws/aws-parallelcluster-cookbook/pull/3132/changes#diff-c14030d9c03bca9bd0926ab68b29b48215b5bc00149ab47b46fbab5aee57f123R230), but i will move them to the top of the recipe.

cookbooks/aws-parallelcluster-slurm/recipes/config/config_head_node.rb
hgreebe: I wanted to copy what was done here: https://github.com/aws/aws-parallelcluster-cookbook/pull/3132/changes#diff-c14030d9c03bca9bd0926ab68b29b48215b5bc00149ab47b46fbab5aee57f123R208

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
```diff
@@ -0,0 +1,47 @@
...
+# License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+execute "wait for cluster registration" do
+  command "#{node['cluster']['slurm']['install_dir']}/bin/sacctmgr show clusters -Pn cluster=#{node['cluster']['stack_name']} format=cluster | grep -x '#{node['cluster']['stack_name']}'"
```
gmarciani: the option `-x` interprets the parameter as a regex. 
So, for example, if stack name is `giacomo.marciani` is can match `giacomoXmarciani`
That said, we are protected by the fact that stack names onyl allows alphanumeric characters, so the stack name cannot include regex symbols (eg: `.`).

However, I think we should be defensive here and enforce the grep paraeter to be interpreted as a fixed string (`-F`. option).

cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
```diff
@@ -76,10 +76,6 @@
   action action
 end unless on_docker?
 
-file "/var/spool/slurm.state/clustername" do
```
gmarciani, line 79: Just for my better understanding: this deletion was the actual workaround. Correct?

cookbooks/aws-parallelcluster-slurm/recipes/update/clear_slurm_accounting.rb
```diff
@@ -23,7 +23,3 @@
   supports restart: false
   action %i(disable stop)
 end
-
-file '/var/spool/slurm.state/clustername' do
```
gmarciani, line 27: Why do we need to remove the file on the cleanup?

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
```diff
@@ -0,0 +1,49 @@
...
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+return if kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
+
+execute "wait for cluster registration" do
+  command "#{node['cluster']['slurm']['install_dir']}/bin/sacctmgr show clusters -Pn cluster=#{node['cluster']['stack_name']} format=cluster | grep -x '#{node['cluster']['stack_name']}'"
+  retries 30
+  retry_delay 10
```
gmarciani, line 23: Is there a scenario that comes to our mind where slurmctld skips or delays too much the registration of the cluster?
If let;s say slurmctld is not registrering the cluster, why not registering ourselves?

cookbooks/aws-parallelcluster-slurm/recipes/config/config_slurm_accounting.rb
hgreebe: yes

cookbooks/aws-parallelcluster-slurm/recipes/update/clear_slurm_accounting.rb
hgreebe: the state file exists before we enable slurm accounting, so its existence is not dependent upon slurm accounting.

cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
hgreebe: registering ourselves would cause the cluster id mismatch error because it would not use the cluster id saved in the state file

gmarciani (approved):

PR #3139, opened 2026-03-11
Clear update checkpoint on update failure

hgreebe:
### Description of changes
* Clear update checkpoint on update failure
* Leads to the following update sequences: 
  * Update fails on compute node
    1. Trigger = v2, checkpoint = v1 → diff detected
    2. Checkpoint written to v2
    3. Update runs → fails
    4. Checkpoint cleared to ''
    5. Next timer tick (60s): trigger = v2, checkpoint = '' → diff detected → retries 

  * Update fails on compute node and head node triggers rollback
    1. Trigger = v2, checkpoint = v1 → diff detected
    2. Checkpoint written to v2
    3. Update runs → fails
    4. Checkpoint cleared to ''
    5. Head node rolls back trigger to v1
    6. Next timer tick: trigger = v1, checkpoint = '' → diff detected → rollback runs 

  * Update succeeds on compute node but head node triggers rollback:
    1. Trigger = v2, checkpoint = v2 (success, no clearing)
    2. Head node rolls back trigger to v1
    3. Next timer tick: trigger = v1, checkpoint = v2 → diff detected → rollback runs 

### Tests
* Ran all update integration tests
* Triggered a transient update failure on compute node to check that it properly retries

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
+++ b/cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
@@ -49,10 +49,15 @@ if [ "$CURRENT_UPDATE" != "$LAST_APPLIED" ]; then
         exit 1
     }
     
-    # Run the update action
+    # Run the update action; on failure clear checkpoint so the next timer tick retries
     log "Executing update script: $UPDATE_SCRIPT"
-    "$UPDATE_SCRIPT"
-    log "Update script completed"
+    if "$UPDATE_SCRIPT"; then
+        log "Update script completed successfully"
+    else
+        log "[ERROR] Update script failed. Clearing checkpoint to allow retry."
+        timeout $FS_TIMEOUT sh -c "echo '' > \"$LOCAL_CHECKPOINT\"" || true
+        exit 1
+    fi
 else
     log "No update needed: trigger and checkpoint match ('$CURRENT_UPDATE')"
 fi


cookbooks/aws-parallelcluster-platform/templates/check_update/pcluster-check-update.sh.erb
```diff
@@ -49,10 +49,15 @@ if [ "$CURRENT_UPDATE" != "$LAST_APPLIED" ]; then
         exit 1
     }
     
-    # Run the update action
+    # Run the update action; on failure clear checkpoint so the next timer tick retries
     log "Executing update script: $UPDATE_SCRIPT"
-    "$UPDATE_SCRIPT"
-    log "Update script completed"
+    if "$UPDATE_SCRIPT"; then
```
gmarciani, line 54: I am not used to this form of checking the success of the command
In the  script we have `set -e` which causes the current script to exit immediately in case of errors.
So I imagine that when `UPDATE_SCRIPT` fails, the current script exists immediately rather than clearing up the checkpoint. Do you know why this is not  the case?

hgreebe: set -e does not cause the script to exit when the failing command is "part of the test in an if statement".
Here is the documentation: https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin

So when we write `if "$UPDATE_SCRIPT"`, the exit code of `$UPDATE_SCRIPT` is being tested by if, so set -e does not apply. If it fails, execution continues into the else branch where we clear the checkpoint.

gmarciani: Great!

gmarciani (approved):

PR #3140, opened 2026-03-12
Do not delete dna json at end of update

hgreebe:
### Description of changes
* Do not delete dna json at end of update
* This prevents the issue where compute nodes get stuck waiting for the DNA files because when the update succeeds, HN deletes DNA file, but there still may be CN that need those to complete their update.

* This change is safe for the following reasons:
  * After the update is complete, the dna.json file are not used. Compute nodes only read them when cfn-hup-update-action.sh runs. They are never re-read if they linger.
  * Stale DNA files from a previous update would not be picked up by a future update because share_compute_fleet_dna.py overwrites them with fresh data at the start of every update (fetch_dna_files 'Share' runs early in the recipe). So stale files get replaced before any compute node reads them.

### Tests
* Ran all the test_update integ tests
* Ran `test_update_race_conditions` and validated that recipes were executed in the correct order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 **CHANGES**
 - Replace cfn-hup in compute nodes with systemd timer to support in place updates in order to improve MPI collective performance at scale.
   This new mechanism relies on shared storage to sync updates between the head node and compute nodes.
+- Stop cleaning up shared DNA files after a successful head node update.
 - Disable dnf-makecache.timer to improve MPI collective performance on RHEL/Rocky at scale.
 - Mitigate the risk of transient build-image failures in RHEL and Rocky caused by out-of-sync repo mirrors.
 - Upgrade Slurm to version 25.11.3 (from 24.11.7).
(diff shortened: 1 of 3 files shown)


cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -299,7 +299,3 @@ def update_nodes_in_queue(strategy, queues)
   cookbook 'aws-parallelcluster-environment'
   mode '0644'
 end
-
-fetch_dna_files 'Cleanup' do
```
gmarciani, line 303: [Doc] Better to capture this change in the changelog of cookbook. It's a low level implementation details, but with important consequences, so better to have an evidence. We will not surface in the cli changelog.

[Test] Let's capture this change in the corresponding unit test

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 **CHANGES**
 - Replace cfn-hup in compute nodes with systemd timer to support in place updates in order to improve MPI collective performance at scale.
   This new mechanism relies on shared storage to sync updates between the head node and compute nodes.
+- Stop cleaning up shared DNA files after a successful head node update.
```
gmarciani: [nit] We know that we are referring to the update recipe, but I suggest to specify update/rollback, to avoid doubts for the users.

gmarciani (approved):

PR #3144, opened 2026-03-13
Do not delete dna json at end of rollback failure

hgreebe:
### Description of changes
* Do not delete dna json at end of rollback failure
* This is so bootstrapping nodes that finish bootstrapping after the rollback fails do not get stuck waiting for dna json
* Add a marker that is used to determine if we are in an update failed state or a rollback failed.
* Here is the flow: 
  * Update fails, rollback fails:
    1. Update fails > no marker > delete dna, write marker
    2. Rollback fails > marker exists > keep dna, delete marker

  * Update fails, rollback succeeds:
    1. Update fails > no marker > delete dna, write marker
    2. Rollback succeeds > success path deletes marker

  * Update succeeds:
    1. Update succeeds > success path deletes marker if it exists (It should not exist)
### Tests
* Ran all test_update integration tests
* Manually tested state of marker and dna json

### References
* Changes to test: https://github.com/aws/aws-parallelcluster/pull/7288

### DIFF
--- a/cookbooks/aws-parallelcluster-entrypoints/libraries/update_failure_handler.rb
+++ b/cookbooks/aws-parallelcluster-entrypoints/libraries/update_failure_handler.rb
@@ -60,21 +60,29 @@ module ErrorHandlers
 
     def run_recovery
       Chef::Log.info("#{log_prefix} Running recovery commands")
-      cleanup_dna_files if @cleanup_dna_files
       start_clustermgtd if @start_clustermgtd
+      cleanup_dna_files if @cleanup_dna_files
     end
 
     def cleanup_dna_files
       marker = "#{cluster_attributes['shared_dir']}/update_failed_marker"
-      if ::File.exist?(marker)
-        # Marker exists from previous update failure — this is a rollback failure, keep DNA files
-        Chef::Log.info("#{log_prefix} Rollback failure detected, keeping DNA files for bootstrapping nodes")
-        ::File.delete(marker)
-      else
-        # No marker — this is an update failure, clean up DNA files and write marker
+      begin
+        if ::File.exist?(marker)
+          # Marker exists from previous update failure — this is a rollback failure, keep DNA files
+          Chef::Log.info("#{log_prefix} Rollback failure detected (marker found at #{marker}), keeping DNA files")
+          ::File.delete(marker)
+        else
+          # No marker — this is an update failure, clean up DNA files and write marker
+          Chef::Log.info("#{log_prefix} Update failure detected (no marker at #{marker}), cleaning up DNA files")
+          command = "#{cookbook_virtualenv_path}/bin/python #{cluster_attributes['scripts_dir']}/share_compute_fleet_dna.py --region #{cluster_attributes['region']} --cleanup"
+          command_runner.run_with_retries(command, description: "cleanup DNA files")
+          ::File.write(marker, '')
+        end
+      rescue => e
+        # If marker I/O fails, fall back to deleting DNA files
+        Chef::Log.warn("#{log_prefix} Error during marker check (#{e.message}), falling back to cleaning up DNA files")
         command = "#{cookbook_virtualenv_path}/bin/python #{cluster_attributes['scripts_dir']}/share_compute_fleet_dna.py --region #{cluster_attributes['region']} --cleanup"
         command_runner.run_with_retries(command, description: "cleanup DNA files")
-        ::File.write(marker, '')
       end
     end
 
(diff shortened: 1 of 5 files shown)


cookbooks/aws-parallelcluster-entrypoints/libraries/update_failure_handler.rb
```diff
@@ -65,8 +65,17 @@ def run_recovery
     end
 
     def cleanup_dna_files
-      command = "#{cookbook_virtualenv_path}/bin/python #{cluster_attributes['scripts_dir']}/share_compute_fleet_dna.py --region #{cluster_attributes['region']} --cleanup"
-      command_runner.run_with_retries(command, description: "cleanup DNA files")
+      marker = "#{cluster_attributes['shared_dir']}/update_failed_marker"
+      if ::File.exist?(marker)
+        # Marker exists from previous update failure — this is a rollback failure, keep DNA files
+        Chef::Log.info("#{log_prefix} Rollback failure detected, keeping DNA files for bootstrapping nodes")
```
gmarciani: [nit] We retain DNA files not only for bootstrapping nodes, but also for nodes that are going to apply the update. I would simply say "keeping DNA files"

The log line should include the path of the marker file to facilitate troubleshooting, because that is the decision maker.

cookbooks/aws-parallelcluster-entrypoints/libraries/update_failure_handler.rb
```diff
@@ -65,8 +65,17 @@ def run_recovery
...
+        ::File.delete(marker)
+      else
```
gmarciani: Include a logline here, similar to the one above to explain how we took the decision to cleanup

cookbooks/aws-parallelcluster-entrypoints/recipes/update.rb
```diff
@@ -26,6 +26,12 @@
 
 include_recipe 'aws-parallelcluster-slurm::update' if node['cluster']['scheduler'] == 'slurm'
 
+# Clean up update failure marker on success
+file "#{node['cluster']['shared_dir']}/update_failed_marker" do
+  action :delete
+  only_if { ::File.exist?("#{node['cluster']['shared_dir']}/update_failed_marker") }
```
gmarciani: why only_if is required?
Does the delete action fail if the file is missing?
If the only_if statement is indeed required, then avoid the duplication of the path and use a constant instead

himani2411: IF this action is supposed to be done on success, then why is it not at the end of this recipe?

Even though DevSettings are not officially supported,
* keeping this cleanup after the custom node update should be supported since some customers could be using this paramter
* And this custom node path runs on the HeadNode which is the center point of identifying if the update succeeds or not.
* In future if we provide a Wiki with a workaround of using this DevSettings we should not face issues during updates.

cookbooks/aws-parallelcluster-entrypoints/libraries/update_failure_handler.rb
```diff
@@ -65,8 +65,17 @@ def run_recovery
     end
 
     def cleanup_dna_files
-      command = "#{cookbook_virtualenv_path}/bin/python #{cluster_attributes['scripts_dir']}/share_compute_fleet_dna.py --region #{cluster_attributes['region']} --cleanup"
-      command_runner.run_with_retries(command, description: "cleanup DNA files")
+      marker = "#{cluster_attributes['shared_dir']}/update_failed_marker"
+      if ::File.exist?(marker)
```
gmarciani: I am a bit concerned by executing this check this way.
We are checking a shared file, so transient networking issues could make this function fail.
If this function fails, the handler exists with two main consequences:
1. DNA files are not deleted (bad, but not too bad)
2. clustermgtd is not restarted (really bad)

My suggestion is:
1. in the update failure handler, within the function run_recovery, restart clustermgtd before cleaning up dna files. In this waym, if there are failures in the shared storage, clustermgtd gets retsrated anyway.
2. wrap the logic around marker I/O with exception handling. In case of errors, we are basically in a situation where we do not know if we are in update or rollback. If we are in doubt we should always delete DNA files because we prefere race condition on rollback failure than race condition on rollback succeess
    * if we do not delete (but we should have)  -> race conditions on rollback success
    * if we delete (but we should have not) -> race condition on rollback failure

cookbooks/aws-parallelcluster-entrypoints/recipes/update.rb
gmarciani: This is a good point. The removal of the marker should be the very last step of the update

gmarciani:
Can we capture this change in the changelog?
It will not be reflected to the main product changelog, but good to have a reference here.

gmarciani (approved):

PR #3156, opened 2026-03-31
[DCV] Fix Xdcv crashes caused by DCV misconfiguration that attempts GL initialization when GPU acceleration is not supported.

gmarciani:
### Description of changes
Fix Xdcv crashes caused by DCV misconfiguration that attempts GL initialization when GPU acceleration is not supported.

**1. Disable dcv-gl on instances where GPU acceleration is not supported**
DCV-GL should always be disabled on instance types that do not support GPU acceleration, such as g5g or instance types w/o GPU.
`enable-gl-in-virtual-sessions` was left at its default value (`default-on`) even on instance types that don't support GPU acceleration. On RHEL9 and Rocky9 this caused Xdcv to segfault during GLX initialization. On other OSes the misconfiguration was handled more gracefully w/o crashes, but we see that DCV is trying to setup DVC-GL, which is in any case not correct for those instance types. 
We now explicitly set `enable-gl-in-virtual-sessions = "always-off"` in `dcv.conf` when GPU acceleration is not supported.

**2. Restrict display encoders to software-only on non-GPU instances**
Instance types that do not support GPU acceleration should not use hardware encoding, but software encoding.
`display-encoders`, which is the dcv.conf parameter controlling this aspect was controlled by the `cluster.dcv.is_graphic_instance` attribute, which was never actually set (it was accidentally removed in https://github.com/aws/aws-parallelcluster-cookbook/commit/04893c21). As a result, the software-only encoder list was never applied. We now correctly set the value for this setting.

**3. Disable NVIDIA EGL platform config files on non-GPU instances**
DCV should never try to initialize NVIDIA libraries on instance types that do not support GPU acceleration.
However, pcluster bakes into the AMI the NVIDIA libraries, hence on non-GPU instances these libraries are present on disk but have no hardware to back them. When Xdcv initializes the virtual session, the EGL loader discovers the NVIDIA EGL config files and attempts to load NVIDIA EGL libraries, causing the crash because there is no GPU. We now mask those files, so the EGL loader falls back cleanly to `swrast` software rendering.
Apparently this defects affects only the first virtual session. It seems that after the first crash, DCV is smart enough to directly fall back to the correct rendering. However, we want to fix this setup to prevent crashes on the first session.

### Tests
* PR checks (include tests covering this change)
* ONGOING `test_dcv_configuration` executed on all OSs

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/templates/dcv/dcv.conf.erb
+++ b/cookbooks/aws-parallelcluster-platform/templates/dcv/dcv.conf.erb
@@ -41,7 +41,13 @@
 # 'dcv-gl' feature (a specific license will be required).
 # Allowed values: 'always-on', 'always-off', 'default-on', 'default-off'.
 # If not specified, the default value is 'default-on'.
+<% if @gpu_accel_supported %>
 #enable-gl-in-virtual-sessions = "default-on"
+<% else %>
+# Explicitly disable dcv-gl on instances where GPU acceleration is not supported
+# (e.g. g5g) to prevent Xdcv segfaults during GLX initialization.
+enable-gl-in-virtual-sessions = "always-off"
+<% end %>
 
 # Property "virtual-session-source-profile" specifies whether the shell that
 # runs the session starter script should source the user profile.
@@ -99,7 +105,10 @@ virtual-session-source-profile = true
 # second that are sent to the client. A higher value consumes more bandwidth
 # and resources. By default it is set to 25. Set to 0 for no limit
 #target-fps = 30
-<% unless node['cluster']['dcv']['is_graphic_instance'] %>
+<% unless @gpu_accel_supported %>
+# On instances that do not support GPU acceleration, DCV must use
+# software-only encoders because there is no GPU that can be used
+# for hardware encoding.
 display-encoders=['turbojpeg', 'lz4', 'ffmpeg']
 <% end %>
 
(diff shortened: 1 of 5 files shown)


cookbooks/aws-parallelcluster-platform/templates/dcv/dcv.conf.erb
```diff
@@ -99,7 +105,7 @@ virtual-session-source-profile = true
 # second that are sent to the client. A higher value consumes more bandwidth
 # and resources. By default it is set to 25. Set to 0 for no limit
 #target-fps = 30
-<% unless node['cluster']['dcv']['is_graphic_instance'] %>
```
himani2411, line 102: IS this an attribute we set, i cant find it being mentioned anywhere so we add these display-encodes for all  instances regardless it is GPU or not? Or is this an ExtraChefAttribute?

cookbooks/aws-parallelcluster-platform/test/controls/dcv_spec.rb
```diff
@@ -238,6 +238,10 @@
     it { should be_owned_by 'root' }
     it { should be_grouped_into 'root' }
     it { should be_mode 0755 }
+
+    unless instance.graphic? && instance.nvidia_installed? && instance.dcv_gpu_accel_supported?
+      its('content') { should match /enable-gl-in-virtual-sessions\s*=\s*"always-off"/ }
```
himani2411, line 243: Can you also add the check for `display-encoders`?

gmarciani: Good catch! Addressed

cookbooks/aws-parallelcluster-platform/templates/dcv/dcv.conf.erb
gmarciani: This attribute was supposed to be set by the cookbook config phase (not injected through dev settings) and it used to be. However,  we mistakenly removed it in https://github.com/aws/aws-parallelcluster-cookbook/commit/04893c21. Added info in the Pr description.

himani2411 (dismissed):

himani2411 (approved):

PR #3159, opened 2026-04-09
Slurm accounting should not be case sensitive

Timothy-Wang-Arctos:
### Description of changes
Prior to 3.15.0 the slurm accounting bootstrap was not case sensitive. The split from https://github.com/aws/aws-parallelcluster-cookbook/pull/3132 made it case sensitive causing the bootstrap to fail when the stack has capital letters in it. 

Bootstrap would fail with 

"Mixlib::ShellOut::ShellCommandFailed: execute[wait for cluster registration] (aws-parallelcluster-slurm::bootstrap_slurm_accounting line 20) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '1'"

with the change to the grep command the return is the expected [0]

/opt/slurm/bin/sacctmgr show clusters -Pn cluster=AL2023-Arctos format=cluster | grep -Fx AL2023-Arctos echo $?
1

/opt/slurm/bin/sacctmgr show clusters -Pn cluster=AL2023-Arctos format=cluster | grep -Fxi AL2023-Arctos al2023-arctos
echo $?
0

### Tests
attempting to deploy with a stackname with uppercase letters caused fail. Rerunning with all lowercase stackname was successful 

### DIFF
--- a/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
+++ b/cookbooks/aws-parallelcluster-slurm/recipes/config/bootstrap_slurm_accounting.rb
@@ -18,7 +18,7 @@
 return if kitchen_test? || (node['cluster']['node_type'] == "ExternalSlurmDbd")
 
 execute "wait for cluster registration" do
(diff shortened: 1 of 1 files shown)


gmarciani:
Hello @Timothy-Wang-Arctos ,

thanks for reporting this issue. This fix is included in https://github.com/aws/aws-parallelcluster-cookbook/pull/3164, which includes also other fixes related to the registration of the cluster with SLURM accounting.

Best

PR #3175, opened 2026-05-06
Fix DynamoDB retry behavior and error message on compute node bootstrap

hehe7318:
### Description of changes

When a compute node cannot reach DynamoDB during bootstrap (for example, a private subnet is missing the DynamoDB VPC gateway endpoint), Chef's `ruby_block[retrieve compute node info]` retries silently with no useful diagnostic information. The default configuration (30 retries × ~5 min per attempt) also exceeds Slurm's default ResumeTimeout (2100s), so the instance is terminated before Chef can emit a final error.

This change:

- **Tightens timeouts and retries** so Chef fails fast within ResumeTimeout:
  - `retries`: 30 → 5
  - `aws_connection_timeout_seconds`: 30 → 10
  - `aws_read_timeout_seconds`: 60 → 30
  - `shell_timeout_seconds`: 300 → 60
  - Worst-case total time: ~5–6 minutes (under the 35 minutes `ResumeTimeout` default).
- **Improves the failure message** to include the CLI exit code and stderr, plus a hint about the most common cause (missing DynamoDB VPC endpoint on a private subnet).
- Fixes a pre-existing typo: `shell_timout_seconds` → `shell_timeout_seconds`.

Related bug: https://github.com/aws/aws-parallelcluster-node/pull/702

### Tests
- E2E on a cluster whose compute subnet has no DynamoDB VPC endpoint and no NAT:
  - Chef fails within 6 minutes instead of looping silently for >2 hours.
  - `/var/log/chef-client.log` on the compute node contains the new error message with CLI stderr and the VPC-endpoint hint.
```
2026-05-06T20:24:56.000Z [2026-05-06T20:24:56+00:00] INFO: Retrying execution of ruby_block[retrieve compute node info], 1 attempt left
2026-05-06T20:26:10.000Z [2026-05-06T20:26:10+00:00] INFO: Retrying execution of ruby_block[retrieve compute node info], 0 attempt left
2026-05-06T20:26:10.000Z     
    Error executing action `run` on resource 'ruby_block[retrieve compute node info]'
    
    RuntimeError
    ------------
    Failed to query DynamoDB for compute node info: the aws cli call did not return in time and was terminated. This usually means the compute node cannot reach DynamoDB. If the compute subnet has no internet egress (NAT/IGW), ensure a DynamoDB VPC gateway endpoint is configured and attached to the subnet's route table.
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ via custom Slurm settings or the cluster name contains upper-case letters.
 - Remove deprecated parameter `AccountingStorageUser` from Slurm configuration that was causing harmless error messages.
 - Fix DCV prerequisite installation potentially blocking on interactive prompts by exporting `DEBIAN_FRONTEND=noninteractive` to child processes.
 - Fix slurmrestd failing to start on AL2023 because the http-parser library was not discoverable by the dynamic linker.
-- Fix compute node bootstrap failing without a clear error when the compute subnet cannot reach DynamoDB (e.g. when a DynamoDB VPC gateway endpoint is missing).
+- Fix compute node bootstrap hanging without a clear error when the compute subnet cannot reach DynamoDB.
 
 3.15.0
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ via custom Slurm settings or the cluster name contains upper-case letters.
 - Remove deprecated parameter `AccountingStorageUser` from Slurm configuration that was causing harmless error messages.
 - Fix DCV prerequisite installation potentially blocking on interactive prompts by exporting `DEBIAN_FRONTEND=noninteractive` to child processes.
 - Fix slurmrestd failing to start on AL2023 because the http-parser library was not discoverable by the dynamic linker.
+- Fail fast and log a descriptive error when a compute node cannot reach DynamoDB during bootstrap, so the root cause
+  (e.g. a missing DynamoDB VPC endpoint) is visible in the Chef logs before Slurm's ResumeTimeout terminates the instance.
 
 3.15.0
 ------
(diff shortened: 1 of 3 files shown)


himani2411:
Why reduce all off these timeouts?why not just reduce the retries? 
```
retries: 30 → 5
aws_connection_timeout_seconds: 30 → 10
aws_read_timeout_seconds: 60 → 30
shell_timeout_seconds: 300 → 60
```


Also, Are there scenarios other than a missing VPC endpoint where a compute node cannot reach DynamoDB? If so (e.g., security group misconfiguration, NACLs, route table issues, DNS resolution failures, or DynamoDB service throttling), these aggressive timeouts could lead to a high instance turnover rate — nodes being marked unhealthy and replaced before transient issues resolve.

hehe7318:
> Why reduce all off these timeouts?why not just reduce the retries?

`shell_timeout` is the real timeout to kill the attempt in this case. It was 5 mins before, too long. If we only reduced retries to 5 and kept shell_timeout=300, a broken node still burns `5 × 300s = 25 min`.

> Also, Are there scenarios other than a missing VPC endpoint where a compute node cannot reach DynamoDB? If so (e.g., security group misconfiguration, NACLs, route table issues, DNS resolution failures, or DynamoDB service throttling), these aggressive timeouts could lead to a high instance turnover rate — nodes being marked unhealthy and replaced before transient issues resolve.

The scenario you mentioned, most of them are config errors (SG, NACL, route table, DNS, and also missing endpoint in this PR). In these cases, longer timeout doesn't help, it will just delay the error. We want fast fail + protected mode trigger.

For the transient error like DDB throttling or other brief DDB errors → The boto3 has its own retry mechanism, also in our logic we tolerate it for 5~6 minutes. Do we think it's not enough? The only real risk is a region wide DDB outage > 6 min, I think it's rare. Even if we encounter that issue, `slurm_resume` will try to wake up new nodes.

himani2411 (approved):

PR #3183, opened 2026-05-19
[GPU] In GPU Health Check, skip DCGM diagnostics when NVIDIA MIG is enabled because dcgmi diag does not support MIG.

gmarciani:
### Description of changes
In GPU Health Check, skip DCGM diagnostics when NVIDIA MIG is enabled because dcgmi diag does not support MIG.

### Tests
* SUCCESS Manually verified that the health check succeeds both on p4d.24xlarge with MIG enabled and on g4dn.2xlarge with MIG disabled
* SUCCESS `test_cluster_with_gpu_health_checks` on g4dn.2xlarge AL2023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 - Improve resilience of EBS volume attachment during cluster creation by retrying on transient IMDS connectivity failures.
 - Further reduce transient build-image failures on RHEL and Rocky caused by out-of-sync repo mirrors by resetting metadata upon retry.
 
+**CHANGES**
+- In GPU Health Check, skip DCGM diagnostics when NVIDIA MIG is enabled because dcgmi diag does not support MIG.
+
 **BUG FIXES**
 - Fix cluster creation failure caused by Slurm accounting bootstrap failing when ClusterName is overridden 
 via custom Slurm settings or the cluster name contains upper-case letters.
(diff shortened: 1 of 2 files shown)


hanwen-cluster:
Minor: there is `is_dcgm_supported` function to skip all checks when DCGM is not supported. With the function we are skipping other `nvidia-smi` commands too. Can you enlighten me on why those preliminary tests are still useful? If they are useful, should we move down the  `is_dcgm_supported`?

gmarciani:
> Can you enlighten me on why those preliminary tests are still useful? If they are useful

DCGM Diagnostics is the most important check we do in GPU Health Check. 
However, even without executing it, our GPU health check executes other valuable checks:
1.  nvidia-smi queries, such as `nvidia-smi -q -i $gpu_id`, give us the confidence that GPUs are accessible and the driver is responsive. Also, the outoput of nvidia-smi commands it's useful to conduct analysis on broken environment.
2. starting fabric-manager gives us confidence that fabric manager would work.
3. starting dcgm service gives us the confidence that the service is functional and is able to interact with the driver. Notice that the dcgm service is not only useful for the diagnostics. The end user could use dcgm to 

That said your comment made me notice a defect in my code that I will address:
1. the GPU persistence flipping is there only because it is a requirement of DCGM Diangostics, so in the case of MIG, there is no real value in doing it and we can skip it.

> should we move down the is_dcgm_supported?

Yes, good point

hanwen-cluster:
Thanks for the explanation. 

pcluster cookbook already enables GPU persistence here https://github.com/aws/aws-parallelcluster-cookbook/blob/f1bfb7a5217434aab62015286654791ddeb1c1dc/cookbooks/aws-parallelcluster-platform/templates/nvidia/parallelcluster_nvidia_service.erb#L4

Maybe we can remove the persistence enablement in GPU health check script completely

gmarciani:
> Maybe we can remove the persistence enablement in GPU health check script completely

We cannot because GPU health checks could run on a custom AMI where that portion of the cookbook may not be executed

gmarciani:
@hanwen-cluster, I figured out that at it is better to exit when MIG is enabled the same way we exit when DCGM is not supported. 

There is value in executing the FM start, as I mentioned [here](https://github.com/aws/aws-parallelcluster-cookbook/pull/3183#issuecomment-4489611780), but the current approach of our health check is to ignore the failure if FM fails to start, so it is not a check at all. I think we should change the approach in the health check and be more strict, failing if FM fails to start, but in any case it's a bigger shift out of scope for this PR, so better to keep the current PR focused and simple.

hanwen-cluster (approved):

PR #3060, opened 2025-12-09
[Build] Mitigate the impact of out-of-sync repo mirrors in Rocky and RHEL by refreshing local cache + add retries to flaky kitchen test

gmarciani:
### Description of changes
Mitigate the impact of out-of-sync repo mirrors in Rocky and RHEL by refreshing local cache at every failure.
The strategy to mitigate consists in two steps:
1. Execute dns install with --refresh option to force DNF to re-download metadata from the mirror it's currently using. 
1. Cleanup DNF metadata at every retry to force DNF to re-evaluate available mirrors and select another one the next attempt.

To mitigate this issue I introduced an utility function to wrap code with retries and custom callback between retries, which could be helpful also in other flaky scenarios.

### Tests
1. Build image on all supported OS, which includes the execution of the flaky kitchen test.
1. Injected fault to trigger the retry logic: example of output
```
Stdout:       * ruby_block[install_packages_with_metadata_refresh] action run[2026-01-07T18:10:40+00:00] INFO: Processing ruby_block[install_packages_with_metadata_refresh] action run (aws-parallelcluster-platform::install line 27)
Stdout: [2026-01-07T18:10:40+00:00] INFO: Installing packages: vim, ...
Stdout: [2026-01-07T18:10:40+00:00] WARN: TESTING: Injecting artificial failure on attempt 1
Stdout: [2026-01-07T18:10:40+00:00] ERROR: Attempt 1/10 failed: Artificial failure for testing retry mechanism (attempt 1)
Stdout: [2026-01-07T18:10:40+00:00] INFO: Executing custom on_retry callback...
Stdout: [2026-01-07T18:10:40+00:00] INFO: Executing: dnf clean metadata --quiet
Stdout: [2026-01-07T18:10:40+00:00] INFO: Exit status: 0
Stdout: [2026-01-07T18:10:40+00:00] INFO: Sleeping 5 seconds before next attempt...
Stdout: [2026-01-07T18:10:45+00:00] WARN: TESTING: Injecting artificial failure on attempt 2
Stdout: [2026-01-07T18:10:45+00:00] ERROR: Attempt 2/10 failed: Artificial failure for testing retry mechanism (attempt 2)
Stdout: [2026-01-07T18:10:45+00:00] INFO: Executing custom on_retry callback...
Stdout: [2026-01-07T18:10:45+00:00] INFO: Executing: dnf clean metadata --quiet
Stdout: [2026-01-07T18:10:46+00:00] INFO: Exit status: 0
[...]

PR #3187, opened 2026-05-21
[Build] Further Mitigate build image failures in RHEL/Rocky caused by out of sync repositories

hanwen-cluster:
### Description of changes
Generalizes the approach created by https://github.com/aws/aws-parallelcluster-cookbook/pull/3060

Introduces a `robust_package` resource that wraps the Chef `package` resource with a metadata-refresh retry strategy on RHEL/Rocky to mitigate transient build failures from out-of-sync RHUI mirrors. On Amazon Linux and Debian-based platforms it falls back to the stock `package` resource with retries, so behavior is unchanged.

In addition, this commit does some refactor to reduce code duplication.

### Tests
* Build image is consistently succeeding

### Checklist


cookbooks/aws-parallelcluster-shared/resources/robust_package.rb
```diff
@@ -0,0 +1,58 @@
...
+#   robust_package 'install enroot prerequisites' do
+#     packages prerequisites
+#   end
+#
+
+provides :robust_package
+unified_mode true
+
+property :packages, [String, Array], required: true, name_property: false
+property :max_retries, Integer, default: 10
```
gmarciani, line 36: [Non blocking] Retries are required when the repo is undergoing maintenance/synchronization.
I do not known how much it could take the repo to recover but 50 seconds seems a too short interval. What about doubling its chances with a max_retries =20?

cookbooks/aws-parallelcluster-shared/resources/robust_package.rb
```diff
@@ -0,0 +1,58 @@
...
+  retry_delay = new_resource.retry_delay
+
+  if platform?('redhat', 'rocky')
+    ruby_block "robust_package install #{new_resource.name}" do
+      block do
+        dnf_install_with_refresh(packages, max_retries: max_retries, retry_delay: retry_delay)
+      end
+    end
+  else
+    package packages do
```
gmarciani, line 53: [Not blocking] Why not using the new robust way also for the other OSs?

gmarciani (approved):

cookbooks/aws-parallelcluster-shared/resources/robust_package.rb
hanwen-cluster: It is not a one way door. If we observe failures in other OSes, we can use it

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
```diff
@@ -194,3 +194,42 @@ def run_cmd(cmd, timeout: nil, raise_on_error: false)
...
+#      available mirrors and potentially pick a different one next attempt.
+#
+# @param packages [Array<String>, String] Package names (optionally version-pinned)
+# @param max_retries [Integer] Maximum number of retry attempts (default: 10)
+# @param retry_delay [Integer] Seconds to wait between retries (default: 5)
+#
+# @example
+#   dnf_install_with_refresh(%w(dkms rpm-build check check-devel subunit))
+#
+def dnf_install_with_refresh(packages, max_retries: 10, retry_delay: 5)
```
gmarciani, line 216: [Testing] Can we cover the new functions with unit tests?

cookbooks/aws-parallelcluster-shared/resources/robust_package.rb
```diff
@@ -0,0 +1,58 @@
...
+provides :robust_package
+unified_mode true
+
+property :packages, [String, Array], required: true, name_property: false
+property :max_retries, Integer, default: 10
+property :retry_delay, Integer, default: 5
+
+default_action :install
+
+action :install do
```
gmarciani, line 41: [Testing] This new resource is not covered by unit test

cookbooks/aws-parallelcluster-shared/resources/robust_package.rb
hanwen-cluster: The 10 retries with 5 retry delay is picked because the code before this PR had such setup. For now, because the build is passing consistently, how about leaving as is?

Clearing the metadata could also leads to changing of the repo mirrors. So if one mirror is under maintenance, another mirror might be functional.

PR #2875, opened 2025-02-10
[Scaling] Remove usage of cfn-init in Compute Fleet

himani2411:
### Description of changes
Changing the Create and Update Path of Compute Nodes in a cluster as we need to remove usage of cfn-init due to CFN API throttling issues and improve fleet scaling time.

For `Create Path` we revert to a approach we used in ParallelCluster 3.8.0 
* In CLI will rely on using cloud native approach of using write_files directives as part of user_data.sh ( Changes in https://github.com/aws/aws-parallelcluster/pull/6655)
* In Cookbook, we add changes which we use/excute during Updation of the node
      * create shared sub-directory `/opt/parallelcluster/shared/dna` which is used for storing dna.json
      * create script `/opt/parallelcluster/scripts/share_compute_fleet_dna.py` which is executed only by root user on HeadNode
      * Create `/opt/parallelcluster/scripts/cfn-hup-update-action.sh` which is executed only by root user on Compute node  . This script monitors the shared /dna directory and runs cookbook Update recipes on the node.
      
For `Update Path` we will rely on HeadNode to share dna.json and extra.json for each node as per their Launch Templates
using EC2 DescribeLaunchTemplateVersions API
* In CLI, the CFN Launch templates are updated.
* In Cookbook,
          * HeadNode run a new script to get latest dna.json and store it in shared directory ( as part of  fetch_dna_files resource )
          * cfn-hup invokes an update hook action script which monitors the shared directory for checking latest dna.json files and runs cookbook update recipes.

Dependent on CLI https://github.com/aws/aws-parallelcluster/pull/6655

### Tests
Same as https://github.com/aws/aws-parallelcluster/pull/6655

### References
* Link to impacted open issues.
* Link to related PRs in other packages (i.e. cookbook, node).
* Link to documentation useful to understand the changes.

### Checklist
- Make sure you are pointing to **the right branch**.
- If you're creating a patch for a branch other than `develop` add the branch name as prefix in the PR title (e\.g\. `[release-3.6]`).
- Check all commits' messages are clear, describing what and why vs how.
- Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- Check if documentation is impacted by this change.

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3188, opened 2026-05-27
[UpdateWorkflow] Make login node use head-node-driven orchestration for cluster updates

gmarciani:
**This PR depends on https://github.com/aws/aws-parallelcluster/pull/7409**

### Description of changes
Make login node use head-node-driven orchestration for the update workflow.

With this change, login nodes do not depend on cfn-hup/cfn-init anymore and mirror the same update mechanism already adopted for compute nodes. With this change we expect the update workflow to be more resilient.

The changes in this PR mirrors the changes made for compute fleet in the following PRs:
1. https://github.com/aws/aws-parallelcluster-cookbook/pull/3070: this is the core change that removes cfn-hup
1. https://github.com/aws/aws-parallelcluster-cookbook/pull/2875: change in cookbook to remove cfn-init, required by change #1
1. https://github.com/aws/aws-parallelcluster/pull/6655: change in cli to remove cfn-init, required by change #1
1. https://github.com/aws/aws-parallelcluster-cookbook/pull/3093: minor change related to logs

In particular:
1. cfn-hup daemon is removed from login nodes
2. cfn-init commands are removed frmo login nodes
3. the system service to check for cluster updates is added to login nodes
4. when the update starts on head node, the head node waits for login nodes launch template to be updated and stores the launch template in the shared folder to be consumed by the login node update workflow. We need a custom logic to implement this waiting because we cannot define such a fine grained dependency in the CFN because the head node launch template and login nodes launch template live in different stacks. 

#### Notes
1. This change required to fix a bug: when internal shared storage is EFS we used to not mount `/opt/parallelcluster/shared`, but it was expected (we mount it when internal storage is EBS). We fixed as part of this PR because we need to reference directly a file in that shared directory.
1. the chef attribute `in_place_update_on_fleet_enabled` is removed as it is not useful anymore. It was used to disable cfn-hup on login nodes and consequentely skip the readiness checks. Now that neither compute nor login nodes run cfn-hup, it makes no sense to keep it. The readiness check is anyways gated by its own only_if logic.
1. `skip-security-exclusions-check` is safe: bandit `B108` is only complaining about the hardcoded path `/tmp/dna.json`. This is not introduced by this change, we have always accepted to store dna.json in /tmp.
1. `skip-recursive-deletion-check` is safe: the check detects a false positive; we are moving existing lines of code that contains `recursive: true` for directory creation, not for deletions.
1. The scripts `cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py` and `cookbooks/aws-parallelcluster-environment/templates/cfn_hup_configuration/ComputeFleet/cfn-hup-update-action.sh.erb` should be renamed and moved to reflect the fact that it is now used to share also login fleet's dna. I am not doing it as part of this PR to facilitate the review. Will rename/move in follow up PR.

### Tests
* SUCCESS
```
test-suites:
  update:
    test_update.py::test_update_slurm:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
    test_update.py::test_dynamic_file_systems_update:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update.py::test_dynamic_file_systems_update_rollback:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update.py::test_login_nodes_update:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update_rollback_failure.py::test_update_rollback_failure:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
    test_update_race_conditions.py::test_update_race_conditions:
      dimensions:
        - instances:
            - c5.xlarge
          oss:
            - alinux2023
          regions:
            - us-east-1
          schedulers:
            - slurm
```


cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
```diff
@@ -24,11 +24,11 @@
 from botocore.config import Config
 from retrying import retry
 
-COMPUTE_FLEET_SHARED_LOCATION = "/opt/parallelcluster/shared/"
+SHARED_LOCATION = "/opt/parallelcluster/shared/"
```
himani2411, line 27: From what i understand the `/opt/parallelcluster/shared` is the shared location with HN and compute nodes only https://docs.aws.amazon.com/parallelcluster/latest/ug/directories-v3.html

And `/opt/parallelcluster/shared_login_nodes` is the same for HN and LN; are we changing this dependency?

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
```diff
@@ -200,10 +254,68 @@ def cleanup(directory_loc):
...
+    """
+    lt_config = get_compute_launch_template_ids(LAUNCH_TEMPLATE_CONFIG)
+    if not lt_config:
+        return
+
+    login_pools = lt_config.get("LoginPools") or {}
+    if not login_pools:
+        return
+
+    for pool_name, pool in login_pools.items():
```
himani2411, line 321: Some of these validations for Login Nodes are also done in `get_latest_dna_data_for_login_nodes` can we make them separate functions to reduce code-duplication?

Why are we validating `lt_version` when we know its hardcoded to be `$Latest` always? Why not do `LogicalID`?

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
```diff
@@ -200,10 +254,68 @@ def cleanup(directory_loc):
...
+            raise RuntimeError(f"Could not fetch UserData for launch template {lt_name} of login pool {pool_name}")
+
+        write_directives = get_write_directives_section(user_data)
+        if not write_directives:
+            raise RuntimeError(
+                f"Could not extract write_files from UserData of launch template {lt_name} of login pool {pool_name}"
+            )
+
+        actual_version = _extract_cluster_config_version(write_directives)
+        if actual_version != expected_config_version:
```
himani2411, line 326: I understand we should raise an exception if cluster-config_version do not match; but if it does match why waste an iteration and not just write the file? 

So we do the same validations as per https://github.com/aws/aws-parallelcluster-cookbook/pull/3188/changes/60fb4a86a4d1a30857fb019faf13033972439d6e#r3344504716 we could also extent this function to write the file in an else condition. what are your thought?

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_efs.rb
```diff
@@ -87,7 +87,10 @@
 node['cluster']['internal_shared_dirs'].each do |dir|
   # Don't mount the login nodes shared dir to compute nodes
```
himani2411, line 88: Unclear on what the use-case is and I am missing the historical reasoning of keeping separate  ` /opt/parallelcluster/shared_login_nodes` and `/opt/parallelcluster/shared`. Lets align on this offline.

cookbooks/aws-parallelcluster-environment/resources/cfn_hup_configuration.rb
himani2411, line 19: [Non-Blocking] Some of the configuration we do in `configure` is only needed for HeadNode now; can we separate this call for HeadNode where we call entire `configure` action and only `extra_configuration` for ComputeFleet and LoginNodes so that we dont create un-wanted files like `/etc/cfn/hooks.d/pcluster-update.conf` and `/etc/cfn/cfn-hup.conf` for all the Nodes in cluster. This can lead to confusion for CX on whether we actually use cfn-hup or not.

himani2411: [Non-Blocking]Lets remove `update_hook_script_dir` and other `ComputeFleet` specific variables which are no longer needed.

cookbooks/aws-parallelcluster-environment/templates/cfn_hup_configuration/cfn-hook-update.conf.erb
```diff
@@ -1,8 +1,7 @@
 [parallelcluster-update]
 triggers=post.update
-<% case node['cluster']['node_type'] -%>
-<% when 'HeadNode', 'LoginNode' -%>
+<% if node['cluster']['node_type'] == 'HeadNode' -%>
```
himani2411: [Non-Blocking] Another cleanup as per https://github.com/aws/aws-parallelcluster-cookbook/pull/3188/changes#r3344604499

We dont really need this if condition once we stop calling the configure action of cfn_hup for all nodes in cluster.

cookbooks/aws-parallelcluster-platform/recipes/config.rb
```diff
@@ -26,6 +26,6 @@
 include_recipe 'aws-parallelcluster-platform::supervisord_config'
 fetch_config 'Fetch and load cluster configs'
 include_recipe 'aws-parallelcluster-platform::config_login' if node['cluster']['node_type'] == 'LoginNode'
-if ['ComputeFleet'].include?(node['cluster']['node_type'])
+if %w(ComputeFleet LoginNode).include?(node['cluster']['node_type'])
```
himani2411, line 29: [Nitpick] Can we convert this into a function?

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
```diff
@@ -287,7 +289,7 @@ def update_nodes_in_queue(strategy, queues)
 
 chef_sleep '15'
 
-wait_cluster_ready if cluster_readiness_check_on_update_enabled?
+wait_cluster_ready
```
himani2411, line 292: what about using `cluster_readiness_check_enabled?` Are we removing use of `cluster_readiness_check_enabled` attributes?

The function we removed checked for `cluster_readiness_check_enabled` as well as `in_place_update_on_fleet_enabled`  
```
def cluster_readiness_check_on_update_enabled?
  cluster_readiness_check_enabled? && node['cluster']['in_place_update_on_fleet_enabled'] == 'true'
end
```

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
```diff
@@ -107,15 +107,17 @@ def wait_sync_file(path)
   end
 end
 
-def cfnhup_enabled?
-  # cfn-hup is always enabled on the head node, as it is required to perform cluster updates.
-  # cfn-hup can be disabled on compute nodes and login nodes, limiting the cluster update in the sense that
-  # live updates on compute and login nodes are not possible.
-  node['cluster']['node_type'] == 'HeadNode' || node['cluster']['in_place_update_on_fleet_enabled'] == 'true'
-end
+def login_nodes_enabled?
```
himani2411, line 110: why are we looking at `launch-templates-config.json` instead of looking at cluster-config.yaml to identify if LoginPools exist in cluster similar to how we look for slurmdbd is being used in cluster?


why rely on a file which might not exist and raise an exception on it when you can check using cluster-config.yaml?

CHANGELOG.md
```diff
@@ -25,6 +28,7 @@ via custom Slurm settings or the cluster name contains upper-case letters.
 - Fix Xdcv segfault caused by DCV attempting GL initialization when GPU acceleration is not supported.
 - Fix slurmrestd failing to start on AL2023 because the http-parser library was not discoverable by the dynamic linker.
 - Fix compute node bootstrap hanging without a clear error when the compute subnet cannot reach DynamoDB.
+- Fix login nodes not mounting `/opt/parallelcluster/shared` when EFS is used as the internal shared storage type.
```
himani2411, line 31: Lets align off-line for this usecase

cookbooks/aws-parallelcluster-slurm/recipes/update/update_head_node.rb
himani2411: Ignore the comment I see cluster_readiness_check_enabled as an only_if guard used inside wait_cluster_ready

cookbooks/aws-parallelcluster-environment/recipes/init/mount_internal_use_efs.rb
gmarciani: `/opt/parallelcluster/shared` contains files that are expected to be shared cluster-wide.
`/opt/parallelcluster/shared_login_nodes` contains files that are expected to be shared only with login nodes.
We introduced `/opt/parallelcluster/shared_login_nodes` when we introduced login nodes because at that point we needed to share files that were only useful to login nodes, such as ssh keys.

As of today, when EBS is used as internal file system, `/opt/parallelcluster/shared` is shared both with compute nodes and login nodes; whereas when EFS is used, the directory is only shared with compute nodes.

cookbooks/aws-parallelcluster-platform/recipes/config.rb
gmarciani: I'm generally in favor of wrapping repetitive logic into dedicated functions. 
However, in this case i do not see a real advantage because the logic to be wrapped is a self-explicative one-liner. There are two options:
1. have a function `should_deploy_check_up[date_systemd_service`: this would reduce clarity because the reader must go into the function to understand when the recipe is included. Considering the condition is really simple, I see more loss than gains.
2. have a function `is_compute_or_login_node`: we are replacing a self-explicative one-liner with another self-explicative one-liner + function definition. So, the net result is that we are increasing the lines of code rather than reducing it.

what do you think?

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
gmarciani: > why are we looking at launch-templates-config.json instead of looking at cluster-config.yaml

That's a great question, I hated to do that :-) 

We can do it and we should, but it would require some refactoring that I preferred to keep out of scope to reduce the moving parts of this change. 
The reason is that when the function `login_nodes_enabled?` is called, the config.yaml is not updated yet. 
The config is updated by the resource `fetch_config` which is executed after the resource `fetch_dna_files`.
See https://github.com/aws/aws-parallelcluster-cookbook/blob/v3.15.1/cookbooks/aws-parallelcluster-platform/recipes/update.rb#L18-L21

The plan is to verify if we can switch the two resources without side effects and if we can, then use the config-yaml file to determine whether login nodes are used or not.

> why rely on a file which might not exist and raise an exception on it when you can check using cluster-config.yaml?

The file `launch-template-config.yaml` always exists because it is written by cfn-init write directives of the head node that are executed before the cookbook. It is actually the file `config.yaml` that does not exist yet when the function `login_nodes_)enabled` is executed (See answer above).
See https://github.com/aws/aws-parallelcluster/blob/1ff9fc7bff02f14ec51510954597658ef8ee7577/cli/src/pcluster/templates/cluster_stack.py#L1554-L1559

What do you think?

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
gmarciani: **Current State**
Code and documentation are misaligned.
As of today, when EBS is used as internal shared storage, the directory `/opt/parallelcluster/shared` is shared with both compute and login nodes; whereas when EFS is used, that one is shared only with compute nodes.

**Decision to take**
1. Add sharing of /opt/parallelcluster/shared with login nodes when EFS is used (align EFS case to EBS case).
2. Remove the sharing of /opt/parallelcluster/shared when EBS is used (align EBS case to EFS case)


**My opinion**
I think the EBS case is the correct one. Having a directory shared cluster-wide allows to avoid files duplication.

**Hisdtorical background**
When we introduced login nodes and the directory `/opt/parallelclutser/login_nodes`, we evaluated the possibility to have 3 folders: shared (globally shared), shared_compute_nodes (shared with compute only) and shared_login_nodes (shared with login nodes only). We decided not to do that for sake of simplicity because the majority of the files in /opt/parallelcluster/shared are needed by both compute and login nodes (eg: config.yaml). 

**Why this Pr make more important to share /shared with both CN and LN?**
The current PR makes it even more important to have a folder shared with both compute and login nodes because the files used to orchestrate the update workflow (eg: update_trigger) are meant to be shared with both of them. Without a common folder, the update orchestrator should duplicate those files in two different locations.

**Next Steps**
Let's sync on this offline to take a decision. I am more in favor of aligning the EFS case to the EBS one, than the other way around because it is useufl, for example in this case, to have a folder globally shared rather than duplicating files in two separate folders.

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
gmarciani: > Some of these validations for Login Nodes are also done in get_latest_dna_data_for_login_nodes can we make them separate functions to reduce code-duplication?

done

> Why are we validating lt_version when we know its hardcoded to be $Latest always? 

Because I prefer to have a single decision maker: launch-template-config.json, the same way we do for compute nodes.
The version id is hard wired for login nodes in that file, but since that file is the decision maker, the downstream logic must not do any assumption. Doing this way is also more open to evolution. Let's say we find a way to remove the dependency of login nodes on head node. Inn that case, we would simply need to change the value of version id in the json file without the need for any change in the downstream logic.

Think about the alternatives:
1. hardwire $Latest both in launch-template-config.json and in this logic: we have two decision makers -> bad
2. hardwire $Latest in this function and not in launch-template.json: single decision maker, but asymmetric w.r.t the compute fleet case -> bad
3. remove $Latest in launch-template-config.json and hardwire it in this function: single decision maker, but asymmetric w.r.t the compute fleet case -> bad


> Why not do LogicalID?

We need the launch template name to describe the launch template and understand whether is has been updated or not. The launch template logical id is included in the launch-template-config.json only because we use that id in the dna file for the launch template, so it is not useful in this function.

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
gmarciani: > if it does match why waste an iteration and not just write the file?

This function has the only goal to check whether the login nodes launch templates have been updated with the expected configuration. Failing with an exception when this is not the case ti required so that the function in the cookbook that calls it will retry the check. If we do not fail, the upstream cookbook function will succeed, and the cookbook will think that the launch templates have been updated when this is not the case.

Are you aligned?

> we could also extent this function to write the file in an else condition. 

Why?

cookbooks/aws-parallelcluster-environment/resources/cfn_hup_configuration.rb
gmarciani: Fixed

cookbooks/aws-parallelcluster-environment/templates/cfn_hup_configuration/cfn-hook-update.conf.erb
gmarciani: Fixed.

cookbooks/aws-parallelcluster-environment/resources/cfn_hup_configuration.rb
gmarciani: Very good catch! Fixed.

Note: The file `cfn-hup-update-action.sh` and the corresponding template should now be renamed to reflect that it is not invoked by cfnhup anymore. There are other files in the cookbook that should be renamed/moved to reflect this change. As mentioned in the Pr description, I am intentionally postponing such refactoring to keep this PR focused.

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
himani2411: I am aligned on raising an exception when there is a mismatch.

why waste an iteration when the cluster configurations match; why not write the file immediately?

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
himani2411: Aligned with the choice of sharing` /opt/parallelcluster/shared` when we use EFS similar to EBS usecase

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
himani2411: I am aligned on keeping the change as is for now as it relies on cfn-init on HN; which guarantees that `launch-templates-config.json` will exist.

cookbooks/aws-parallelcluster-platform/recipes/config.rb
himani2411: Agreed lets keep as is.

cookbooks/aws-parallelcluster-environment/resources/cfn_hup_configuration.rb
himani2411: Agreed, we could have handled this as part of the re-naming too! Anyway thanks for the change.

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
himani2411: Aligned offline that this is a safe guard to any manual changes that can be done on the file so its better to assert if all the values we write exist in  launch-template-config.json

cookbooks/aws-parallelcluster-environment/files/cfn_hup_configuration/share_compute_fleet_dna.py
himani2411: Aligned offline that we should keep the writing separate to maintain a single goal and in future if we manage to remove circular dependency we just need to remove this check rather than re-factor the entire file.

himani2411 (approved):

PR #3189, opened 2026-05-27
[Deps] Support 3rd party dependency upgrade using ExtraChefAttributes

himani2411:
### Description of changes

Make ParallelCluster's third-party dependency downloads override-aware (through ExtraChefAttributes) so that we can upgrade dependencies by redirecting to an alternative source (typically the upstream public repo or GitHub release) without affecting the other dependencies.

## Why

Today, dependency download URLs are hardcoded against `node['cluster']['artifacts_s3_url']`. We need to test new dependency versions before mirroring them to the ParallelCluster S3 bucket, but overriding `artifacts_s3_url` is too coarse — it would redirect every dependency, not just the one being tested.

This change introduces per-dependency `base_url` (and where applicable, `version` and `sha`) Chef attributes that can be overridden via ExtraChefAttributes. Each download URL is built from the dependency-specific attribute, so we can override exactly one dependency at a time.

### Phase 1 — standalone dependencies

These dependencies are downloaded from a single URL pattern and don't need any helper logic to switch between S3 and a public repo.

- **EFS-utils** — add `efs_utils.base_url` attribute
- **Pyxis** — add `pyxis.base_url` attribute
- **HTTP parser** — add `http_parser.base_url` and `http_parser.version` attributes
- **pyenv** — add unit test for ExtraChefAttributes override scenario since they already have Chef attributes
- **libjwt** — add unit test for ExtraChefAttributes override scenario since they already have Chef attributes
- **PMIx** — add unit test for ExtraChefAttributes override scenario since they already have Chef attributes
- **CUDA** — add `nvidia.cuda.base_url`, `nvidia.cuda.samples_base_url`, `nvidia.cuda.version`, and `nvidia.cuda.driver_version_suffix` attributes; remove the hardcoded version-suffix lookup table
- **NVIDIA driver** — add `nvidia.driver_base_url` attribute (URL pattern is identical between S3 and the public Tesla URL, so no helper needed)

### Phase 2 — NVIDIA library helpers + repo-aware dependencies

The NVIDIA public repo uses a different URL layout than the ParallelCluster S3 buckets — packages live under `{platform}/{arch}/{filename}` instead of `{platform}/{filename}`. To keep the overhead in one place, three helpers are added to `cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb`:

- `default_artifacts_url?(base_url)` — detects whether a given `base_url` still points at the ParallelCluster S3 mirror.
- `nvidia_repo_arch` — returns `'sbsa'` for ARM and `'x86_64'` otherwise (NVIDIA repo arch convention).
- `nvidia_package_url(base_url, platform, filename)` — builds `{base_url}/{platform}/{filename}` for the S3 mirror and `{base_url}/{platform}/{arch}/{filename}` for the public NVIDIA repo.

Each NVIDIA dependency is updated to switch URL construction to `nvidia_package_url`, paired with a per-dependency `base_url` attribute:

- **DCGM** — `nvidia.dcgm_base_url`
- **Enroot caps** — `enroot.caps_base_url`. Also detects the package-name swap (`enroot-caps` on S3 vs `enroot+caps` on the public repo) via `default_artifacts_url?`.
- **Fabric Manager** — `nvidia.fabricmanager.base_url`
- **NVIDIA IMEX** — `nvidia.imex.base_url`
- **NVLSM** — `nvidia.nvlsm.base_url` and `nvidia.nvlsm.version` (previously hardcoded). Checksum verification is skipped when the `base_url` is overridden, since checksums are pre-computed only for the S3-hosted artifacts.

## Tests

Each updated resource has unit tests added that cover, for every supported OS:
- both x86_64 and aarch64 architectures
- both default S3 and overridden public `base_url` scenarios
- where applicable: filename swaps (enroot caps), checksum-skip behavior (NVLSM), and resource integration (IMEX, NVLSM)

NVIDIA library helpers themselves are exercised through the resource-driven tests; the `nvidia_enabled?` parameterized values are covered separately in `spec/unit/libraries/nvidia_spec.rb`.

* Build image to test default behavour successful

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-environment/test/controls/efs_spec.rb
+++ b/cookbooks/aws-parallelcluster-environment/test/controls/efs_spec.rb
@@ -4,9 +4,9 @@ control 'tag:install_efs_utils_installed' do
 
   only_if { !os_properties.redhat_on_docker? }
 
-  describe file("#{node['cluster']['sources_dir']}/efs-utils-2.4.0.tar.gz") do
+  describe file("#{node['cluster']['sources_dir']}/efs-utils-#{node['cluster']['efs']['version']}.tar.gz") do
     it { should exist }
-    its('sha256sum') { should eq '9b60c039c162388091d6fab6e9c6cfc5832f34b26b6d05b0a68b333147d78a25' }
+    its('sha256sum') { should eq node['cluster']['efs']['sha256'] }
     its('owner') { should eq 'root' }
     its('group') { should eq 'root' }
     its('mode') { should cmp '0644' }
--- a/cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
+++ b/cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
@@ -14,6 +14,32 @@ def nvidia_enabled?
   ['yes', true, 'true'].include?(node['cluster']['nvidia']['enabled'])
 end
 
+# Whether a base_url attribute still points to the default ParallelCluster artifacts location.
+# Used by: fabric manager, dcgm, nvlsm, enroot (caps).
+# Limitation: this check assumes attribute precedence resolves at attribute-load time
+# (the ExtraChefAttributes path). If artifacts_s3_url is overridden after cookbook
+# attributes load — for example via node.override from inside a recipe — base_url
+# will still embed the original value and this check may return false incorrectly.
+def default_artifacts_url?(base_url)
+  base_url.include?(node['cluster']['artifacts_s3_url'])
+end
+
+# NVIDIA public repo uses 'x86_64' or 'sbsa' as the arch directory
+def nvidia_repo_arch
+  arm_instance? ? 'sbsa' : 'x86_64'
+end
+
+# Build download URL for NVIDIA-hosted packages (dcgm, fabric manager, nvlsm).
+# Default: downloads from ParallelCluster S3 mirror — URL pattern is {base_url}/{platform}/{filename}
+# Overridden: downloads from NVIDIA public repo — URL pattern is {base_url}/{platform}/{arch}/{filename}
+def nvidia_package_url(base_url, platform, filename)
+  if default_artifacts_url?(base_url)
+    "#{base_url}/#{platform}/#{filename}"
+  else
+    "#{base_url}/#{platform}/#{nvidia_repo_arch}/#{filename}"
+  end
+end
+
 #
 # Check if the instance has a GPU
 #
--- a/cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
+++ b/cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
@@ -44,9 +44,15 @@ action :install_nvlsm_dependencies do
 end
 
 action :install_nvlsm do
+  base_url = node['cluster']['nvidia']['nvlsm']['base_url']
   remote_file "#{node['cluster']['sources_dir']}/#{nvidia_nvlsm_package_full_name}" do
     source nvidia_nvlsm_url
-    checksum nvidia_nvlsm_checksum
+    # The NVLSM checksum is specific to each distribution and architecture combination,
+    # and a single overridden value cannot satisfy all OS/architecture variants when
+    # build-image runs in parallel across them. Skip the checksum when base_url is
+    # overridden; the cookbook still validates the checksum on the default S3 path
+    # where the per-OS/per-arch values are pinned.
+    checksum nvidia_nvlsm_checksum if default_artifacts_url?(base_url)
     mode '0644'
     retries 3
     retry_delay 5
@@ -70,11 +76,11 @@ def nvidia_nvlsm_package
 end
 
 def nvidia_nvlsm_version
-  "2025.03.9-1"
+  node['cluster']['nvidia']['nvlsm']['version']
 end
 
 def nvidia_nvlsm_url
-  "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_nvlsm/#{platform}/#{nvidia_nvlsm_package_full_name}"
+  nvidia_package_url(node['cluster']['nvidia']['nvlsm']['base_url'], platform, nvidia_nvlsm_package_full_name)
 end
 
 def nvidia_nvlsm_package_full_name
--- a/cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
+++ b/cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
@@ -199,3 +199,76 @@ describe 'nvidia_nvlsm:install' do
     end
   end
 end
+
+# Tests for the NVIDIA library helpers nvidia_package_url and default_artifacts_url?
+# as exercised through nvidia_nvlsm URL construction and checksum verification.
+#
+# Default S3 path:   {base_url}/{platform}/{filename} (with checksum verification)
+# Public repo path:  {base_url}/{platform}/{arch}/{filename} (checksum skipped because
+# the test harness pre-computes checksums only for the S3-hosted artifacts)
+describe 'nvidia_nvlsm install URL and checksum override behavior' do
+  s3_nvlsm_base_url = "#{cluster_artifacts_s3_url}/dependencies/nvidia_nvlsm"
+  public_nvlsm_base_url = 'https://fake-public.example.DOMAIN/compute/cuda/repos'
+
+  for_all_oses do |platform, version|
+    debian = (platform == 'ubuntu')
+    ext = debian ? 'deb' : 'rpm'
+    package_join = debian ? '_' : '-'
+    arch_join = debian ? '_' : '.'
+    expected_platform = if platform == 'amazon'
+                          "amzn#{version}"
+                        elsif %w(redhat rocky).include?(platform)
+                          "rhel#{version}"
+                        else
+                          "#{platform}#{version.delete('.')}"
+                        end
+
+    %w(x86_64 aarch64).each do |arch|
+      arch_suffix = debian ? arch_suffix_debian[arch] : arch_suffix_rhel[arch]
+      package_filename = "nvlsm#{package_join}2025.03.9-1#{arch_join}#{arch_suffix}.#{ext}"
+
+      [
+        ['default S3 base_url',
+         s3_nvlsm_base_url,
+         "#{s3_nvlsm_base_url}/#{expected_platform}/#{package_filename}",
+         true],
+        ['overridden public base_url',
+         public_nvlsm_base_url,
+         "#{public_nvlsm_base_url}/#{expected_platform}/#{arch == 'aarch64' ? 'sbsa' : 'x86_64'}/#{package_filename}",
+         false],
+      ].each do |scenario, base_url, expected_source, checksum_expected|
+        context "on #{platform}#{version} #{arch} with #{scenario}" do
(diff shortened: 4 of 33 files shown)


cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
```diff
@@ -44,9 +44,10 @@
 end
 
 action :install_nvlsm do
+  base_url = node['cluster']['nvidia']['nvlsm']['base_url']
   remote_file "#{node['cluster']['sources_dir']}/#{nvidia_nvlsm_package_full_name}" do
     source nvidia_nvlsm_url
-    checksum nvidia_nvlsm_checksum
+    checksum nvidia_nvlsm_checksum if default_artifacts_url?(base_url)
```
gmarciani, line 55: We should always check the checksum. Why here we do it conditionally?

cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
```diff
@@ -199,3 +199,76 @@ def self.install(chef_run)
...
+  public_nvlsm_base_url = 'https://fake-public.example.DOMAIN/compute/cuda/repos'
+
+  for_all_oses do |platform, version|
+    debian = (platform == 'ubuntu')
+    ext = debian ? 'deb' : 'rpm'
+    package_join = debian ? '_' : '-'
+    arch_join = debian ? '_' : '.'
+    expected_platform = if platform == 'amazon'
+                          "amzn#{version}"
+                        elsif %(redhat rocky).include?(platform)
```
gmarciani: May be a minor thing, but why `%` (string) instead of `%w` (array of strings)?
I guess it works either way in this case because platform never has truncated platform names, but worth double checking.

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
```diff
@@ -14,6 +14,31 @@ def nvidia_enabled?
   ['yes', true, 'true'].include?(node['cluster']['nvidia']['enabled'])
 end
 
+# Whether a base_url attribute still points to the default ParallelCluster artifacts location.
+# Used by: fabric manager, dcgm, nvlsm, enroot (caps).
+# Note: if artifacts_s3_url is overridden via ExtraChefAttributes without also overriding
```
gmarciani: Why do I need to specify both?
Example: I want to use dependencies from the official bucket but just one from a custom bucket. What should I do?

cookbooks/aws-parallelcluster-environment/test/controls/efs_spec.rb
```diff
@@ -4,9 +4,9 @@
 
   only_if { !os_properties.redhat_on_docker? }
 
-  describe file("#{node['cluster']['sources_dir']}/efs-utils-2.4.0.tar.gz") do
+  describe file("#{node['cluster']['sources_dir']}/efs-utils-#{node['cluster']['efs']['version']}.tar.gz") do
     it { should exist }
-    its('sha256sum') { should eq '9b60c039c162388091d6fab6e9c6cfc5832f34b26b6d05b0a68b333147d78a25' }
+    its('sha256sum') { should eq node['cluster']['efs']['sha256'] }
```
gmarciani, line 9: The change to the test is correct: now the version and checksum are not "hard wired", but depend on the value of attributes. However, I am concerned that we are losing coverage. Before this change, the test used to verify that the version was the expected one. After this change we do not have this guarantee anymore. We should have a test that verifies the default values

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
himani2411: The NVLSM checksum is specific to each distribution and architecture combination. As a result, parallel image builds across all OS/architecture variants are not possible without a code change that supports providing checksums for all permutations.

So a conditional change is the simplest solution

cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
himani2411: The %w would be the correct solution as it does the accurate member matching rather than substring matching which % will do.

cookbooks/aws-parallelcluster-environment/test/controls/efs_spec.rb
himani2411: The test is still verifying the default value — it's just doing so through the attribute layer instead of inlining a duplicate copy of that value.

If no override is provided (the production path), the cookbook's defaults take effect: node['cluster']['efs']['version'] = '2.4.0' and the corresponding SHA. The recipe downloads efs-utils-2.4.0.tar.gz, and the InSpec test resolves the same attribute and asserts efs-utils-2.4.0.tar.gz exists with the matching SHA. The default value is verified end-to-end, just not by hardcoding 2.4.0 a third time in the test file.

Hardcoding the old version doesn't add a check — it makes the test incompatible with the overriding scenario we want to support. When the dependency-update build-image runs, it overrides efs.version to a newer version, the cookbook downloads that newer file, and InSpec asserts efs-utils-2.5.0.tar.gz. With the hardcoded 2.4.0, the test fails not because the cookbook misbehaved but because the test was written against a stale literal.

Switching to the attribute-driven assertion is what lets the test stay correct under both paths:

No override → defaults used → test asserts default version is on disk → passes.
Override applied → override used → test asserts overridden version is on disk → passes.
Either way, the assertion is still "the cookbook installed the version it was supposed to install."

cookbooks/aws-parallelcluster-platform/spec/unit/resources/nvidia_nvlsm_spec.rb
himani2411: Updated

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
himani2411: I will update the comment to make the scenario clear.

You don't need to specify both — that's the whole point of the per-dependency base_url attributes.

The default for every dependency is computed from node['cluster']['artifacts_s3_url']:

```
default['cluster']['nvidia']['nvlsm']['base_url'] = "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_nvlsm"
```
So out of the box, all dependencies pull from the same S3 bucket.

To override just one dependency to a custom bucket, the user sets only that dependency's base_url via ExtraChefAttributes, e.g.:
```

{
  "cluster": {
    "nvidia": {
      "nvlsm": {
        "base_url": "https://my-custom-bucket.example.com/nvlsm"
      }
    }
  }
}

```
Every other dependency continues to use the default S3 bucket. This is exactly the use case we are supporting — override one base_url at a time to validate a new version against the upstream public source while everything else stays on the production S3 mirror.

The `default_artifacts_url?(base_url)` helper checks whether the per-dependency base_url has been overridden — not whether `artifacts_s3_url` was overridden. So a user who overrides only `nvlsm.base_url `triggers the override-aware path for NVLSM only; everything else keeps using the S3 path.

himani2411: ## The two override scenarios

### Scenario A: Override `artifacts_s3_url` via `--json-attributes` (the only realistic pattern)

```
--json-attributes='{"cluster": {"artifacts_s3_url": "https://my-mirror.example.com"}}'
```

Chef's load order:
1. JSON attributes are merged into the node tree at **normal** precedence (level 3).
2. Cookbook attribute files are loaded, setting **default** precedence (level 1).
3. When `platform.rb` evaluates:
   ```ruby
   default['cluster']['nvidia']['nvlsm']['base_url'] = "#{node['cluster']['artifacts_s3_url']}/dependencies/nvidia_nvlsm"
   ```
   the `node['cluster']['artifacts_s3_url']` interpolation reads the merged value — which is **"https://my-mirror.example.com/"** (normal wins over default).
4. So `nvlsm.base_url` becomes **`"https://my-mirror.example.com/dependencies/nvidia_nvlsm"`**.

Now in the action:
```ruby
default_artifacts_url?(base_url)
  → base_url.include?(node['cluster']['artifacts_s3_url'])
  → "https://my-mirror.example.com/dependencies/nvidia_nvlsm".include?("https://my-mirror.example.com")
  → true ✓
```

**Result: helper returns true correctly.** Download URL is `<mirror>/<platform>/<filename>` (the simple S3 layout). Download works **as long as your mirror has the same layout as the PCluster S3 bucket**. And this is the usual scenario we use with ExtraChefAttributes during build image `--json-attributes /etc/parallelcluster/image_dna.json`

### Scenario B: Override `artifacts_s3_url` AFTER cookbook attributes load [Uncommon but possible]

If something like a recipe does `node.override['cluster']['artifacts_s3_url'] = "https://my-mirror.example.com"` AFTER `platform.rb` already evaluated:

- `nvlsm.base_url` = `"https://us-east-1-aws-parallelcluster.../dependencies/nvidia_nvlsm"` (frozen at load time with the *original* artifacts_s3_url)
- Now `node['cluster']['artifacts_s3_url']` = `"https://my-mirror.example.com"` (overridden post-hoc)

```ruby
default_artifacts_url?(base_url)
  → "https://us-east-1-aws-parallelcluster...".include?("https://my-mirror.example.com")
  → false ✗
```

**Result: helper returns false incorrectly.** Here i am saying `incorrectly`  since the function goal is to answer `"Are we still pointing at the default PCluster artifacts location?"` And the goal of user is to override this artifact URL with the mirror URL.

 Code path goes "treat as public NVIDIA repo, insert arch directory". Download URL becomes:
```
https://us-east-1-aws-parallelcluster.../dependencies/nvidia_nvlsm/<platform>/<arch>/<file>
```
That URL doesn't exist — the PCluster S3 bucket doesn't have the `<arch>` subdirectory. **Download fails with HTTP 403/404.**

## So which scenario does the comment describe?

The comment in `nvidia.rb` was written about Scenario B, but Scenario B isn't how the the use-case overrides attributes. ExtraChefAttributes are passed via `--json-attributes` which is processed **before** cookbook attribute files load. So Scenario A is the only realistic case.


Even this is the Uncommon scenario I would like to keep this comment so that later if we make a mistake of using `node.override` in our recipe we will be quickly able to pinpoint the issue when it fails.
 I will update the comment to make the message clear.

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
gmarciani: Got it, could you please add a comment about this limitation above the condition?

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
gmarciani: Thanks, I am aligned with the approach just confused by the comment. Thanks for addressing

cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
himani2411: updated

himani2411: Updated

gmarciani (approved):

PR #3193, opened 2026-06-02
[Deps] Support ArmPL, stunnel, dcv dependency upgrade using ExtraChefAttributes

hehe7318:
### Description of changes
Follow-up to #3189, completing the remaining dependencies (ArmPL, stunnel, dcv) for the third-party dependency upgrade automation.

This moves the ArmPL, stunnel, and dcv version / sha256 / `base_url` values out of hardcoded constants and into `platform.rb` attributes so they can be overridden via `ExtraChefAttributes`, matching the pattern already merged for the NVIDIA dependencies.

* `armpl.base_url`, `armpl.version`, `armpl.gcc.patch_version`, `gcc.base_url`
* `stunnel.base_url`, `stunnel.version`, `stunnel.sha256`
* `dcv.base_url`

ArmPL additionally skips the per-platform directory segment (`armpl/RHEL-9/...`) when `base_url` is overridden to a non-S3 source, because Arm's public CDN does not partition tarballs by platform. gcc is downloaded from the new `gcc.base_url` attribute.

There is no behavior change when the defaults are kept.

### Tests
* Added unit tests, passed.
* `build_image` integ test passed.

### Checklist
- [x] Make sure you are pointing to **the right branch** (`develop`).
- [x] Make sure **to have added unit tests** to cover the new/modified code.
- [x] Check if documentation is impacted by this change. (No customer-facing change; internal refactor enabling the dependency-upgrade pipeline.)


cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
```diff
@@ -24,11 +24,12 @@
 # https://developer.arm.com/downloads/-/arm-compiler-for-linux
 #
 # Usually we upgrade gcc version as well (see below).
-# We upload ArmPL to a ParallelCluster bucket (account for it in scope of the upgrade) and download it from there
-# to install ArmPL on the AMI.
-# We download gcc directly from gnu.org repository to install correct gcc version on the AMI.
-property :armpl_major_minor_version, String, default: '24.10'
-property :gcc_patch_version, String, default: '0'
+# By default ArmPL and gcc are downloaded from a ParallelCluster S3 bucket
+# (managed via dependency_downloader_and_uploader). The armpl.base_url and
```
himani2411: Please do not use pipeline in the comments, keep the comments generalized with how we override using ExtraChefAttributes

cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
```diff
@@ -100,7 +104,7 @@
   end
 
   gcc_version = "#{gcc_major_minor_version}.#{new_resource.gcc_patch_version}"
```
himani2411, line 95: Why do we override the `gcc_patch_version` using ExtraChefAttributes when we cannot override the `gcc_major_minor_version` which is platform specific?

cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
```diff
@@ -24,11 +24,12 @@
...
-# We upload ArmPL to a ParallelCluster bucket (account for it in scope of the upgrade) and download it from there
-# to install ArmPL on the AMI.
-# We download gcc directly from gnu.org repository to install correct gcc version on the AMI.
-property :armpl_major_minor_version, String, default: '24.10'
-property :gcc_patch_version, String, default: '0'
+# By default ArmPL and gcc are downloaded from a ParallelCluster S3 bucket
+# (managed via dependency_downloader_and_uploader). The armpl.base_url and
+# gcc.base_url attributes can be overridden (e.g. via ExtraChefAttributes) to
+# point at the public sources directly.
+property :armpl_major_minor_version, String, default: lazy { node['cluster']['armpl']['version'] }
```
himani2411: Why do we do lazy load on node attributes? Any impact on attribute precedence (https://docs.chef.io/client/19/cookbooks/attributes/attribute_precedence/) ?

himani2411: Please test the default behavior by image Build

cookbooks/aws-parallelcluster-platform/spec/unit/resources/arm_pl_spec.rb
```diff
@@ -189,3 +189,90 @@ def self.setup(chef_run)
...
+# Default S3 mirror partitions ArmPL tarballs by platform directory
+# (armpl/RHEL-9/...). A non-S3 source (Arm's public CDN) is not partitioned, so
+# the platform segment is skipped. gcc is always downloaded from gcc.base_url.
+describe 'arm_pl:setup base_url override behavior' do
+  for_all_oses do |platform, version|
+    context "on #{platform}#{version}" do
+      cached(:artifacts_s3_url) { 'https://REGION-aws-parallelcluster.s3.REGION.AWS_DOMAIN/archives' }
+      cached(:sources_dir) { 'sources_test_dir' }
+      cached(:package_manager) { platform == 'ubuntu' ? 'deb' : 'rpm' }
+      cached(:armpl_version) { '24.10' }
```
himani2411: USe any fake value so that when we want to test we dont have to override/update these values in tests.

cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
hehe7318: Sorry for that. It's my fault -> created by AI, I didn't carefully check the comment. I have removed the comment. The code can explain itself.

cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
hehe7318: It was a implementation mistake. Switched to the same pattern as the NVIDIA driver resource: `property :armpl_major_minor_version, String` with a `_armpl_major_minor_version` helper that falls back to `node['cluster']['armpl']['version']`. No more attribute precedence concern.

cookbooks/aws-parallelcluster-platform/spec/unit/resources/arm_pl_spec.rb
hehe7318: Done.

cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
hehe7318: Yeah, `gcc_major_minor_version` can not be overridden. 

Do you mean the patch version override is not worth it? I'm happy to drop it if so.

himani2411: Yes lets drop it for now; the better way to handle this is to use GCC in the parent AMI which is already tracked in our backlog; except for some distro most of the OS already have a GCC version higher than what we are pinning.

hehe7318: Done!

cookbooks/aws-parallelcluster-platform/resources/arm_pl/partial/_arm_pl_common.rb
hehe7318: build_image integ test passed

cookbooks/aws-parallelcluster-platform/attributes/platform.rb
```diff
@@ -8,6 +8,14 @@
 
 # ArmPL
 default['conditions']['arm_pl_supported'] = arm_instance?
+default['cluster']['armpl']['version'] = '24.10'
+default['cluster']['armpl']['base_url'] = "#{node['cluster']['artifacts_s3_url']}/armpl"
+default['cluster']['gcc']['base_url'] = "#{node['cluster']['artifacts_s3_url']}/dependencies/gcc"
```
himani2411: Nitpick: lets remove this if we dont use or override it

hehe7318: Done!

himani2411 (approved):

PR #3207, opened 2026-06-19
[Build] Install the NVIDIA driver and CUDA toolkit from the distribution package manager using NVIDIA local repo packages instead of the run file installers.

gmarciani:
**THIS PR CANNOT BE MERGED UNTIL WE PUBLISH THE REFERENCED ARTIFACTS TO OUR PUBLIC BUCKETS**

### Description of changes
Migrate the NVIDIA driver and CUDA toolkit installation from the .run runfile installers to NVIDIA local-repo packages (.rpm/.deb) installed through the distribution package manager.

This is the approach suggested both by NVIDIA and AWS. Compared to the original run file approach, it is also more robust against kernel bumps.

As part of this change we also:
1. converted the installation of cuda toolkit from recipe to resource.
2. remove the nvidia local repos to save space in the AMI. This is safe because if the user must change the nvidia software versions they should download different local repos in any case.
3. install fabric manager only if it is not already installed (this fixes a pre-existing bug in our code, where we used to install fabric manager even if it was already installed)

### Tests
1. Unit test (added/updated) included in this PR check
4. Build image on all supported OSs (with DevSettings pointing to NVIDIA local repos from Nvidia website because artifacts are not published to our bucket yet)
5. Successful integ tests:
```
test-suites:
  basic:
    test_essential_features.py::test_essential_features:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["g4dn.xlarge"]
          oss: ["alinux2023","ubuntu2204"]
          schedulers: ["slurm"]
  health_checks:
    test_gpu_health_checks.py::test_cluster_with_gpu_health_checks:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["g4dn.xlarge"]
          oss: ["alinux2023","ubuntu2204"]
          schedulers: ["slurm"]

``` 


cookbooks/aws-parallelcluster-platform/resources/nvidia_repo/partial/_nvidia_repo_common.rb
```diff
@@ -1,41 +1,114 @@
...
+end
+
+# Local filesystem path of the downloaded driver local-repo installer.
+def driver_repo_package_path
+  "#{node['cluster']['sources_dir']}/#{driver_repo_package_file}"
+end
+
+# Download URL of the driver local-repo installer.
+def driver_repo_source_url
+  "#{node['cluster']['nvidia']['driver_base_url']}/#{driver_repo_package_file}"
```
himani2411, line 109: This URL will not work when we want to Override using ExtarChefattributes. Please verify if the devSettings will work with the new changes.

gmarciani: checked and it works

cookbooks/aws-parallelcluster-platform/libraries/nvidia.rb
```diff
@@ -14,6 +14,15 @@ def nvidia_enabled?
   ['yes', true, 'true'].include?(node['cluster']['nvidia']['enabled'])
 end
 
+# Convert a full CUDA version to its 'major-minor' dashed form used in CUDA repo
+# and package names, e.g. '13-0' for '13.0.2'. The patch level (the '.2') is
+# intentionally dropped because CUDA repos and packages are keyed on major-minor
+# only (e.g. 'cuda-repo-rhel9-13-0-local', 'cuda-toolkit-13-0').
+def cuda_version_dash(cuda_version)
+  major, minor = cuda_version.split('.')
```
hanwen-cluster, line 22: (Interesting Ruby syntax :) )

cookbooks/aws-parallelcluster-platform/resources/nvidia_cuda/partial/_nvidia_cuda_install_debian.rb
```diff
@@ -0,0 +1,20 @@
...
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+action :install_cuda do
+  apt_package cuda_toolkit_package do
```
hanwen-cluster: How about using [robust_package](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/aws-parallelcluster-shared/resources/robust_package.rb)? Therefore, both debian and rhel logic can live in [_nvidia_cuda_common.rb](https://github.com/aws/aws-parallelcluster-cookbook/pull/3207/changes#diff-9a4b22bd1097a221908377f31f86932e1bd77497da72e266de3bcd055a32504f)

hanwen-cluster: There are more occurrences in nvidia driver too

cookbooks/aws-parallelcluster-platform/resources/nvidia_cuda/partial/_nvidia_cuda_common.rb
```diff
@@ -0,0 +1,121 @@
...
+property :cuda_version, String, default: node['cluster']['nvidia']['cuda']['version']
+
+action :setup do
+  return unless nvidia_enabled?
+  return if on_docker?
+
+  # Share CUDA versions with InSpec tests. We expose the major.minor under a
+  # dedicated attribute and leave the canonical 'version' attribute as the full
+  # version, so other resources (e.g. nvidia_repo) can rely on it regardless of order.
+  node.default['cluster']['nvidia']['cuda']['major_minor_version'] = cuda_major_minor
```
hanwen-cluster, line 32: Does it mean a `pcluster build-image` of a specific pcluster version will no longer pin the patch version of CUDA?

cookbooks/aws-parallelcluster-platform/resources/nvidia_cuda/partial/_nvidia_cuda_install_debian.rb
gmarciani: The advantage that i see using robust_package is the robustness on RHEL/Rocky when dealing with RHUI repository, which is not what we do in this case. Do you see any advantage other than allowing us to remove the RHEL partial for cuda?

cookbooks/aws-parallelcluster-platform/resources/nvidia_cuda/partial/_nvidia_cuda_common.rb
gmarciani: No, the build-image still installs the CUDA version that is specified in the cookbook attribute `['cluster']['nvidia']['cuda']['version']`. The difference now is that such pinning is implicit in the installed repository.

That attribute is used by the resource `nvidia_repo` to download and install the cuda repository for that patch version. We then use the major-minor version to install the cuda-toolkit package from that repo because this is the way nvidia ships the package in the repo.

Do you agree we are ok with the current approach?

cookbooks/aws-parallelcluster-platform/resources/nvidia_cuda/partial/_nvidia_cuda_install_debian.rb
gmarciani: synced offline: better to use the resource `package` rather than robust_package and remove the platform specific partials. I'll fix

gmarciani: Done

hanwen-cluster (dismissed):

cookbooks/aws-parallelcluster-platform/test/controls/nvidia_spec.rb
```diff
@@ -15,9 +15,12 @@
       (node['cluster']['nvidia']['enabled'] == 'yes' || node['cluster']['nvidia']['enabled'] == true)
   end
 
+  # driver_version holds either the version pcluster installed or, on images that
```
himani2411, line 18: Can we add spec test to check the nvidia repo is removed?

gmarciani: Added to the nvidia_repo kitchen

himani2411 (approved):

PR #3211, opened 2026-06-24
[Node] Install aws-parallelcluster-node from S3 in all regions

himani2411:
### Description of changes

The node package was installed from S3 only in us-iso regions; everywhere else it install from PyPI ('pip install aws-parallelcluster-node'), breaking air-gapped/proxied build-image in commercial and govcloud.

Set custom_node_package to the official S3 node URL in all regions by default, routing through the existing custom_parallelcluster_node S3 install path. Reuse the existing install_python_from_internet attribute as the opt-out: set it to install from PyPI instead. A user-provided custom_node_package still takes precedence.

node_spec: default path now asserts S3 install; added a context covering the install_python_from_internet PyPI fallback.

Note: requires the node package to be present at
<s3_url>/parallelcluster/<version>/node/ in all partitions

### Tests
* Build Image with DevSettings

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
@@ -50,7 +50,7 @@ unless node['cluster']['install_python_from_internet']
   end
 end
 
-bash "install custom aws-parallelcluster-node" do
+bash "install aws-parallelcluster-node" do
   cwd Chef::Config[:file_cache_path]
   code <<-NODE
     set -e
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
@@ -50,7 +50,7 @@ unless node['cluster']['install_python_from_internet']
   end
 end
 
-bash "install aws-parallelcluster-node" do
+bash "install aws-parallelcluster-node from #{node['cluster']['custom_node_package']}" do
   cwd Chef::Config[:file_cache_path]
   code <<-NODE
     set -e
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/install/parallelcluster_node.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/install/parallelcluster_node.rb
@@ -33,18 +33,16 @@ activate_virtual_env node_virtualenv_name do
   not_if { ::File.exist?("#{virtualenv_path}/bin/activate") }
 end
 
-if aws_region.start_with?("us-iso") && !is_custom_node?
-  node_package = "aws-parallelcluster-node-#{node['cluster']['parallelcluster-node-version']}.tgz"
-
-  node.override['cluster']['custom_node_package'] = "#{node['cluster']['s3_url']}/parallelcluster/#{node['cluster']['parallelcluster-node-version']}/node/#{node_package}"
-end
-
-if is_custom_node?
-  include_recipe 'aws-parallelcluster-computefleet::custom_parallelcluster_node'
-else
-  execute "install official aws-parallelcluster-node" do
+# Install the official node package from S3 in all regions (air-gapped/proxied
+# support), unless the user opts into installing from PyPI via
+# install_node_from_internet. A user-provided custom_node_package always wins.
+if node['cluster']['install_node_from_internet']
+  # install_node_from_internet was requested: install from PyPI.
+  execute "install aws-parallelcluster-node from Pypi" do
     command "#{virtualenv_path}/bin/pip install aws-parallelcluster-node==#{node['cluster']['parallelcluster-node-version']}"
     retries 3
     retry_delay 5
   end
+else
+  include_recipe 'aws-parallelcluster-computefleet::custom_parallelcluster_node'
 end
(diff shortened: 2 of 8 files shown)


cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
```diff
@@ -50,7 +50,7 @@
   end
 end
 
-bash "install custom aws-parallelcluster-node" do
+bash "install aws-parallelcluster-node" do
```
gmarciani: Why removing "custom" since the resource it is actually installing the custom node?

cookbooks/aws-parallelcluster-computefleet/recipes/install/parallelcluster_node.rb
```diff
@@ -33,7 +33,10 @@
   not_if { ::File.exist?("#{virtualenv_path}/bin/activate") }
 end
 
-if aws_region.start_with?("us-iso") && !is_custom_node?
+# Install the official node package from S3 in all regions (air-gapped/proxied
+# support), unless the user opts into installing from PyPI via
+# install_python_from_internet. A user-provided custom_node_package always wins.
+if !is_custom_node? && !node['cluster']['install_python_from_internet']
```
gmarciani: When do we expect to use `install_python_from_internet`?

cookbooks/aws-parallelcluster-shared/attributes/versions.rb
```diff
@@ -6,6 +6,13 @@
 # Set to true when testing new Python versions before artifacts are available in S3
 default['cluster']['install_python_from_internet'] = false
 
+# Opt-out: install the aws-parallelcluster-node package from PyPI instead of the
+# official S3 bucket.
+default['cluster']['install_node_from_internet'] = false
+
+# Official ParallelCluster Node Package
+default['cluster']['custom_node_package'] = "#{node['cluster']['s3_url']}/parallelcluster/#{node['cluster']['parallelcluster-node-version']}/node/aws-parallelcluster-node-#{node['cluster']['parallelcluster-node-version']}.tgz"
```
himani2411, line 14: NOTE FOR REVIEWER: Addition of a default attribute will always make `is_custom_node?` to be always true; which leads to node package being installed in cluster finalize recipe https://github.com/aws/aws-parallelcluster-cookbook/blob/4019f023496fe57829acc560cc74d3580ac0b4e9/cookbooks/aws-parallelcluster-entrypoints/recipes/finalize.rb#L21-L23
Setting a default value for an attribute is now making the user-intent unclear.

So every pcluster create and update command will lead to node package being installed on eevry node; the only solution is to add a new attribute or go to the original node.override if-else condition we have set; do you agree?

gmarciani: If the cost of simplifying code and making the user intent clear is to add a new chef attribute, then we should do that

himani2411: As discussed will change how we identify the node is custom (the URL should differ from the default we set) rather than adding a new attribute.

cookbooks/aws-parallelcluster-computefleet/recipes/install/parallelcluster_node.rb
himani2411: The `['cluster']['install_python_from_internet']` is a generalized attribute which was to install latest python and any python related dependencies from internet.

cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
himani2411: Modified to add the source from which we are installing node package

cookbooks/aws-parallelcluster-shared/attributes/versions.rb
gmarciani, line 14: [Minor] Not necessary to handle in this PR, but it would be good to rename the attribute in a follow up PR. 

The attribute `custom_node_package` used to make sense because we were using it only to specify a custom package. Now we are instead using it also to specify the official package, so it should be renamed to `node_package`.

cookbooks/aws-parallelcluster-shared/libraries/helpers.rb
```diff
@@ -75,11 +75,17 @@ def load_cluster_config(config_path)
...
-# Check if custom node is specified in the config
+# Check if a custom node package is specified in the config using DevSettings.
+# We ship a default custom_node_package (the official S3 package), so we
+# compare the effective value against the full default-precedence value:
+# only a value the customer changed counts as custom.
 #
 def is_custom_node?
   custom_node_package = node['cluster']['custom_node_package']
-  !custom_node_package.nil? && !custom_node_package.empty?
+  return false if custom_node_package.nil? || custom_node_package.empty?
```
gmarciani, line 85: what is the scenario where it is nil?/empty? My understanding is that it must always have a value (either the official package or the custom package). If nil/empty the cookbook would break even before this check.

I think the logic should be: return True unless `node['cluster']['custom_node_package']` is equal to the default value.

himani2411: Apart from anyone manually changing the default in the cookbook or dna.json i dont expect it to be nil or empty. i would rather keep the check as its more of a defensive check rather than an expected behavior; also the cookbook is not going to fail even after this check; since `is_custom_node?` will only be run as part of finalize node recipe not part of Build AMI.

cookbooks/aws-parallelcluster-shared/attributes/versions.rb
himani2411: Will create a follow-up PR to handle this

gmarciani (approved):

PR #3212, opened 2026-06-25
Enforce NFSv4-only on the head-node NFS server

hanwen-cluster:
### Description of changes
Enforce NFSv4-only on the ParallelCluster-managed NFS server to fix the sporadic `nfs-server.service` start failure during cluster creation (lockd `makesock failed / EADDRINUSE` on port 32768).

Root cause: a v3-capable nfsd brings up the in-kernel lockd, which binds on port 32768. On head-node creation lockd sometimes lost the race for 32768 to an ephemeral allocation, failing the nfs-server start. A v4-only nfsd never calls lockd_up(), so 32768 is never requested.

Changes (aws-parallelcluster-environment):
- conf.d-capable OSes (AL2023, RHEL/Rocky 9, Ubuntu 22+): drop in /etc/nfs.conf.d/parallelcluster-nfs.conf, leaving the distro's /etc/nfs.conf intact.
- RHEL/Rocky 8 (nfs-utils 2.3.3, no conf.d auto-include): render /etc/nfs.conf directly.
- The drop-in also pins the ancillary v3-client ports ([lockd]/[mountd]/[statd]) rather than relying on nfs::server4. The upstream nfs.erb has no 'amazon' platform_family branch, so on AL2023 it leaves those ports unpinned; pinning them in the drop-in keeps the NFSv3 client able to mount external v3 servers through a restricted firewall on every OS.

Deliberately unchanged:
- The full NFS stack (incl. the NFSv3 client: rpcbind, rpc-statd, the lockd module + pinned ports) is still installed via nfs::server4, so cluster nodes can still mount external NFSv3 servers through a restricted firewall (aws-parallelcluster#6949 left as-is). rpcbind is NOT masked, because that would break v3 client mounts. We need to keep v3 client supported because there are customers using v3: https://github.com/aws/aws-parallelcluster/issues/6949
- Customers can re-enable NFSv3 on the head-node server by overriding node['nfs']['v3'] = 'yes'; per-port knobs (e.g. set a port to 0 for dynamic) keep working since templates read from node['nfs'][...].

### Tests
- Verified on live head nodes (Ubuntu 24, AL2023, RHEL 8): nfs-server starts cleanly, /proc/fs/nfsd/versions and rpcinfo show NFSv4 only, no nlockmgr registered, nothing bound to 32768; statd/mountd remain on pinned ports. Confirmed the RHEL 8 baseline (pre-change) bound lockd on 32768 and served v3, which this change removes.
- integration test PR https://github.com/aws/aws-parallelcluster/pull/7458

### Checklist


cookbooks/aws-parallelcluster-environment/spec/unit/resources/nfs_spec.rb
```diff
@@ -33,41 +33,37 @@ def self.configure(chef_run)
...
           chef_run
         end
-
-      else
-        pending "to be implemented"
       end
 
-      it 'disables service at boot' do
-        is_expected.to disable_service(disabled_service)
+      it 'does not start the nfs server at boot' do
```
gmarciani, line 48: Why do we expect the server service to be disabled at boot?

cookbooks/aws-parallelcluster-environment/templates/nfs/nfs.conf.erb
```diff
@@ -0,0 +1,33 @@
+# Generated by Chef for <%= node['fqdn'] %>
```
gmarciani, line 1: If this is mirroring what the nfs cookbook does, why not referencing directly the template provided by that cookbook rather than making our own copy? I am asking not only to avoid unnecessary code owned by us, but also to be better future proof. Example: what if the nfs cookbook introduces changes that must be reflected on this config file? With the current solution we have the responsibility to maintain the configuration. By referencing the original template instead we do not.

cookbooks/aws-parallelcluster-environment/templates/nfs/parallelcluster-nfs.conf.erb
```diff
@@ -0,0 +1,31 @@
...
+# ParallelCluster NFS drop-in.
+# Read after /etc/nfs.conf, so it overrides only the keys below and leaves the rest of the
+# distro's /etc/nfs.conf (e.g. [general] pipefs-directory, [gssd]) untouched.
+#
+# We render the pinned ancillary ports here rather than relying on nfs::server4 to write them to
+# /etc/nfs.conf, because the upstream nfs.erb has no 'amazon' platform_family branch and so leaves
+# the ports unpinned on Amazon Linux 2023. Pinning them here keeps the NFSv3 client able to mount
+# external v3 servers through a restricted firewall on every conf.d OS. All values come from the
+# node['nfs'][...] attributes, so the upstream knobs still work (e.g. set a port to 0 for dynamic).
+[lockd]
```
gmarciani, line 13: Are we pinning to ports in the ephemeral range?
Also, why are we injecting configurations for NFSv3 if our goal is to move to NFSv4 only? I assume that we are doing because we see advantages of NFSv3 over NFSv4. However, those advantages are not enough to prefer NFSv3 over NFSv4, so why taking responsibility on those settings?

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
   
 
 **CHANGES**
+- Enforce NFSv4-only on the ParallelCluster-managed NFS server (head node). NFSv3 can be re-enabled on the server by
```
gmarciani: We are referencing the DevSetting attribute. Our policy is to never reference DevSettings attributes unless they are introduced to mitigate an impact where the pcluster default does not work (eg: cfn-hup causing update failures or perf regression). Why do you think we should surface this in the changelog? Also, if we think it is necessary we must specify that the attribute is a DevSetting.

cookbooks/aws-parallelcluster-environment/spec/unit/resources/nfs_spec.rb
hanwen-cluster: This test is on `pcluster build-image` phase, where the packages are installed, but services are not enabled.

The nfs server is enabled during cluster creation on head node only https://github.com/aws/aws-parallelcluster-cookbook/pull/3212/changes#diff-d0939a94ec0c666d09491554cff84731276abe64f1d30953a70e59a9b19456b5R21-R26

cookbooks/aws-parallelcluster-environment/templates/nfs/nfs.conf.erb
hanwen-cluster: Yes. This is based on https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/cookbooks/nfs/templates/default/nfs.conf.erb

However, in the nfs cookbook, version of the NFS server (`nfsd`) is not configured. In other words, it is missing the following section:
```
[nfsd]
threads=<%= node['nfs']['threads'] %>
<% unless node['nfs']['v3'].nil? -%>
vers3=<%= node['nfs']['v3'] %>
<% end -%>
<% unless node['nfs']['v4'].nil? -%>
vers4=<%= node['nfs']['v4'] %>
<% end -%>
```

Considering all the incompleteness (not supporting AL2023, not configuring server versions) of the nfs cookbook, this PR moves to our homegrown configuration

cookbooks/aws-parallelcluster-environment/templates/nfs/parallelcluster-nfs.conf.erb
hanwen-cluster: `nfs.conf` is used by both client and server. We are using v4-only for server, therefore, the ports configurations don't really matter. However, on the client side, without specifying the ports, v3 mount will use dynamic ports, which is a behavior change from 3.15.0

> Are we pinning to ports in the ephemeral range?
Only `lockd` is in the ephemeral range. I will look into the possibility to move it to non-ephemeral range

CHANGELOG.md
hanwen-cluster: I don't have strong opinion here
How about:
> Enforce NFSv4-only on the ParallelCluster-managed NFS server (head node). The NFSv3 client stack (rpcbind, rpc-statd, lockd) are unchanged, so cluster nodes can still mount external NFSv3 servers.

cookbooks/aws-parallelcluster-environment/spec/unit/resources/nfs_spec.rb
gmarciani: agreed

cookbooks/aws-parallelcluster-environment/templates/nfs/nfs.conf.erb
gmarciani: if the config managed by the external cookbook is only missing the nfsd configuration, then what concerns do you have in simply adding the missing config rather than owinf the full config?

cookbooks/aws-parallelcluster-environment/templates/nfs/parallelcluster-nfs.conf.erb
gmarciani: Discussed offline: we are aligned on ignoring whatever nfs v3 because in 3.16.0 we are moving to nfsv4 only. if the customer wants to use nfsv3 they must configure it by themselves

cookbooks/aws-parallelcluster-environment/templates/nfs/nfs.conf.erb
hanwen-cluster: We were owning the config anyway in [default-nfs-kernel-server.conf.erb](https://github.com/aws/aws-parallelcluster-cookbook/pull/3212/changes#diff-5050e1e8ee166559d6c50e824ef2b82d916e67c2ceee2f0d043dd4db69d55b41) and [‎cookbooks/aws-parallelcluster-environment/templates/nfs/nfs-ubuntu22+.conf.erb](https://github.com/aws/aws-parallelcluster-cookbook/pull/3212/changes#diff-99e8ceed1c589461ab6427cd038d96031407390aafcb23ce1884eefbe2c87285)

Now the main concern is that AL2023 is not supported by nfs cookbook So the ports have to be pinned in our homegrown logic

cookbooks/aws-parallelcluster-environment/templates/nfs/parallelcluster-nfs.conf.erb
hanwen-cluster: We discussed offline again. So we will keep the behavior of client side support of v4/v3 unchanged comparing to previous pcluster versions

CHANGELOG.md
hanwen-cluster: Done

gmarciani (approved):

PR #3223, opened 2026-07-10
[Cookbook]Add permission diagnostic checks to pcluster-diag

hehe7318:
### Description of changes
Implement the permission checks planned for the diagnostics tool: reserved user/group verification, critical-path ownership verification, and instance-profile consistency verification.

Checks added:
* **ReservedUsersAndGroups**: reserved users (pcluster-admin, slurm, munge) and groups (pcluster-admin, slurm, munge, pcluster-slurm-share) exist and do not share their uid/gid with another account. A duplicate uid makes the kernel resolve it to the first `/etc/passwd` entry, so a daemon relying on a reserved user's NOPASSWD sudoers rules can run as the wrong user and lose its privileges.
* **CriticalPathsHaveExpectedPermissions**: critical files have their expected owner, group, and mode. Data-driven so paths can be added incrementally. Checks `computefleet-status.json` (clusterstatusmgtd write target; losing write access stalls compute-fleet transitions), `/etc/munge/munge.key` (munged refuses to start unless the key is private to the munge user, breaking Slurm auth cluster-wide), and `/var/spool/slurm.state` (slurmctld exits fatal on wrong ownership/perms of its StateSaveLocation).
* **ImdsRoleMatchesCfnHupConfig** (head node): the IAM role IMDS reports matches the `role=` recorded in `/etc/cfn/cfn-hup.conf`, so cfn-hup credential retrieval does not fail after an out-of-band instance-role change.

Each check reports failures as structured `CheckError` codes owned by the check, consistent with the diagnostics Result API.

### Tests
Unit tests passed.

Verified end-to-end on a running ubuntu2404 cluster:
* On a healthy head node, all checks pass.
* Injecting a duplicate uid 400 (`useradd -o -u 400 pc-admin-test`) makes `ReservedUsersAndGroups` report FAILURE with the shared-id code `E2`; a shared reserved gid (`groupadd -o -g 405 dup-share-test`) is reported the same way.
* Loosening the permissions of any `CriticalPathsHaveExpectedPermissions` target (`computefleet-status.json`, `/etc/munge/munge.key`, or `/var/spool/slurm.state`) makes the check report FAILURE with code `E3` naming the exact path and mode.
* Pointing the cfn-hup config at a role IMDS does not report (editing `role=` in `/etc/cfn/cfn-hup.conf`) makes `ImdsRoleMatchesCfnHupConfig` report FAILURE.
* Removing the faults returns all checks to PASSED.


cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
```diff
@@ -0,0 +1,124 @@
...
+from pcluster_diag.core.constants import CLUSTER_ADMIN_GROUP, CLUSTER_ADMIN_USER, COMPUTEFLEET_STATUS_PATH
+from pcluster_diag.models.check import Check
+from pcluster_diag.models.check_error import CheckError
+from pcluster_diag.models.context import Context, NodeType
+from pcluster_diag.models.result import Result
+from pcluster_diag.util import filesystem
+
+
+@dataclass(frozen=True)
+class CriticalPath:
```
gmarciani, line 41: This is an expectation on a wathever path, I would not call it "Critical"

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
```diff
@@ -0,0 +1,124 @@
...
+        self._critical_paths = CRITICAL_PATHS if critical_paths is None else critical_paths
+
+    @property
+    def description(self) -> str:
+        """Return the human-readable description of this Check."""
+        return "Verify that ParallelCluster critical files have their expected owner, group, and permissions."
+
+    def should_run(self, context: Context) -> bool:
+        """Run only when at least one critical path applies to the current node type."""
+        return bool(self._applicable(context))
```
gmarciani: The check should always run because it is always applicable to any node type and config that we must check critical files. If there is no critical file to check, then the check can succeed

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/instance_profile.py
```diff
@@ -0,0 +1,118 @@
...
+                ],
+            )
+
+        if imds_role != configured_role:
+            return Result.failure(
+                self,
+                errors=[
+                    CheckError(
+                        self.ROLE_MISMATCH,
+                        "The IAM role returned by IMDS ('{}') does not match the role configured in {} "
```
gmarciani: Let's simplify the error message. We decided that the error message should only describe the observation against the expected result. Here we can simplify saying that the role do not match and they must make sure they do.

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/reserved_users.py
```diff
@@ -0,0 +1,167 @@
...
+    @property
+    def description(self) -> str:
+        """Return the human-readable description of this Check."""
+        return "Verify that each reserved ParallelCluster user has a unique uid (not shared with another user)."
+
+    def run(self, context: Context) -> Result:
+        """Pass when every present reserved user owns its uid alone; fail listing each shared uid."""
+        errors = []
+        for name in RESERVED_USER_IDS:
+            uid = users.get_user_uid(name)
```
gmarciani: what about making this check a bit more general and return two errors: 
1. the expect user does not exist
1. the user is sharing the uid


Similarily to what we did for the critical files check.

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/reserved_users.py
```diff
@@ -0,0 +1,167 @@
...
+                        ),
+                    )
+                )
+
+        if errors:
+            return Result.failure(self, errors=errors)
+        return Result.passed(self)
+
+
+class ReservedGroupsHaveUniqueIds(Check):
```
gmarciani: This check is almost equal to the user id check. Any chance to reduce code duplicastion?

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/core/registry.py
```diff
@@ -123,4 +130,13 @@ def get(self, identifier: str) -> Optional[Check]:
...
-DEFAULT_REGISTRY = Registry().register(CfnHupRunsOnlyOnHeadNode()).register(SampleCheck())
+DEFAULT_REGISTRY = (
+    Registry()
+    .register(CfnHupRunsOnlyOnHeadNode())
+    .register(ReservedUsersExist())
+    .register(ReservedUsersHaveUniqueIds())
+    .register(ReservedGroupsHaveUniqueIds())
+    .register(CriticalPathsHaveExpectedPermissions())
+    .register(ImdsRoleMatchesCfnHupConfig())
+    .register(SampleCheck())
```
gmarciani: Let;s take the chance to remove the SampleCheck

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
hehe7318: "Critical" is the selection criterion that I put a path in this check. A path is on the list because a permission drift on it breaks the cluster (fleet down, Slurm auth down, slurmctld won't start). But you're right the dataclass itself is just a path+expected permission record. I am happy to rename it to `ExpectedPathPermissions` while keeping the critical semantics on the list and the check. If you want me to do it, let me know.

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
hehe7318: Thanks for catching it. Done!

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/instance_profile.py
hehe7318: Done!

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/reserved_users.py
hehe7318: Done!

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/reserved_users.py
hehe7318: Yes, done!

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/core/registry.py
hehe7318: Done!

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
gmarciani: It's a minor, but if you agree let's do it

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/util/filesystem.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    info = Path(path).stat()
+    return PathStat(
+        owner=_owner_name(info.st_uid),
+        group=_group_name(info.st_gid),
+        mode=_octal_mode(info.st_mode),
+    )
+
+
+def _owner_name(uid: int) -> str:
```
gmarciani, line 51: Seems like a better fit for `_group_name` and `_group_name` would be to place them in `util/users.py` the same way we already do for `get_user_uid` and `get_user_gid` and other related to this topic.

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
```diff
@@ -0,0 +1,145 @@
...
+    SLURM_USER,
+)
+from pcluster_diag.models.check import Check
+from pcluster_diag.models.check_error import CheckError
+from pcluster_diag.models.context import Context, NodeType
+from pcluster_diag.models.result import Result
+from pcluster_diag.util import filesystem
+
+
+@dataclass(frozen=True)
```
gmarciani, line 40: data classes should be in the models folder

gmarciani (approved):
Approved with non blocking comments. Please address in follow up PR.

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
hehe7318: Addressed in https://github.com/aws/aws-parallelcluster-cookbook/pull/3229

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/util/filesystem.py
hehe7318: Addressed in https://github.com/aws/aws-parallelcluster-cookbook/pull/3229

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/critical_paths.py
hehe7318: Addressed in https://github.com/aws/aws-parallelcluster-cookbook/pull/3229

PR #3232, opened 2026-07-20
Install fabric managed, nvlsm, and imex from Nvidia repo

hgreebe:
### Description of changes
* Install fabric manages, nvlsm, and imex from Nvidia repo
* Continue same package lock we did when we installed them from rpm files
* Remove versions for fabric manager, nvlsm, and imex because we install the version included in the repo
* Remove references to download url because we are retrieving from the repo
* Change unit tests to test they get installed from repo
* Keep the exisiting behavior of locking the packages

### Tests
1. Unit test (added/updated) included in this PR check
4. Build image on all supported OSs (with DevSettings pointing to NVIDIA local repos from Nvidia website because artifacts are not published to our bucket yet)
5. Successful integ tests:
```
test-suites:
  basic:
    test_essential_features.py::test_essential_features:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["g4dn.xlarge"]
          oss: ["rhel8"]
          schedulers: ["slurm"]
  health_checks:
    test_gpu_health_checks.py::test_cluster_with_gpu_health_checks:
      dimensions:
        - regions: ["us-east-1"]
          instances: ["g4dn.xlarge"]
          oss: ["rhel8"]
          schedulers: ["slurm"]

``` 

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3237, opened 2026-07-22
Create resource to lock nvidia packages

hgreebe:
### Description of changes
* Create resource to lock nvidia packages
* Removed unused `platform` variable
* Use `if platform_family?('debian')` in common so we don't need platform specific files.

### Tests
* Ran build image on all OSs

### References
* Follow up to: https://github.com/aws/aws-parallelcluster-cookbook/pull/3232


cookbooks/aws-parallelcluster-platform/resources/nvidia_nvlsm/partial/_nvidia_nvlsm_common.rb
```diff
@@ -54,7 +54,11 @@ def nvidia_nvlsm_package
 end
 
 def nvidia_nvlsm_install_dependencies_commands
-  # OS dependent
+  if platform_family?('debian')
```
gmarciani: [Best Practice] This OS-dependent logic should be captured in the OS-specific resource.

cookbooks/aws-parallelcluster-platform/resources/package_lock.rb
```diff
@@ -12,13 +12,25 @@
...
+unified_mode true
+
+property :package_name, String, name_property: true
 
-def nvidia_nvlsm_install_dependencies_commands
-  "apt install -y infiniband-diags ibutils"
+default_action :lock
+
+action :lock do
+  if platform_family?('debian')
```
gmarciani: [Best Practice] The OS-specific logic should be in the OS-specific package_lock resource.
The counter-argument maybe "We introduced this resource to avoid Os-specific logic in the consumers; why should we reintroduce it for the package_lock now? It seems we are just trading once complexity for another!"

Point is that we should always use OS-specific resource pattern, but we should reduce also code duplication. If we define the Os-specific resource for package_lock we do it just once and do not need to redefine it whenever we need a package lock

gmarciani (approved):

PR #3239, opened 2026-07-22
Fix default node package install shadowed by empty custom_node_package from CLI

hanwen-cluster:
### Description of changes

In #3211, the change relied on a default custom_node_package attribute pointing at the official S3 package, but the CLI always injects custom_node_package into dna.json (empty when there is no DevSettings/NodePackage). So the effective value was "", and build-image failed on a cluster without DevSettings/NodePackage.

- Revert the default on custom_node_package. It now means only "customer-supplied override", and revert change in is_custom_node? to a simple non-empty check.
- Add `official_node_package` attribute in `versions.rb`.
- parallelcluster_node recipe selects the source at converge time: PyPI when install_node_from_internet is set, the customer's package when provided, otherwise the official S3 package. Deciding at converge time avoids the precedence trap, so the empty custom_node_package from dna.json is harmless.

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-shared/spec/unit/libraries/helpers_spec.rb
+++ b/cookbooks/aws-parallelcluster-shared/spec/unit/libraries/helpers_spec.rb
@@ -46,17 +46,11 @@ end
 
 describe 'is_custom_node?' do
   let(:node) { Chef::Node.new }
-  let(:default_base) { 'DEFAULT_BASE_URL' }
-  let(:default_package) { "#{default_base}/node/DEFAULT_NODE_PACKAGE" }
 
-  before { node.default['cluster']['custom_node_package'] = default_package }
+  # Ensure the cluster attribute namespace exists.
+  before { node.override['cluster'] = {} }
 
-  it 'returns false when the package equals the shipped default' do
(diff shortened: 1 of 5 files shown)


cookbooks/aws-parallelcluster-shared/spec/unit/libraries/helpers_spec.rb
```diff
@@ -46,17 +46,11 @@
 
 describe 'is_custom_node?' do
   let(:node) { Chef::Node.new }
-  let(:default_base) { 'DEFAULT_BASE_URL' }
-  let(:default_package) { "#{default_base}/node/DEFAULT_NODE_PACKAGE" }
 
-  before { node.default['cluster']['custom_node_package'] = default_package }
+  # Ensure the cluster attribute namespace exists (it always does on a real node).
+  before { node.override['cluster']['region'] = 'us-east-1' }
```
himani2411: Can you use a fake value?

hanwen-cluster: Done

himani2411 (approved):

PR #3241, opened 2026-07-24
[Nvidia] Install Xorg driver on RHEL family for DCV GPU acceleration

gmarciani:
### Description of changes
We made the following changes to make the DCV configuration work on RHEL/Rocky after changing the way we install nvidia software (https://github.com/aws/aws-parallelcluster-cookbook/pull/3207)
* Install Xorg driver. Apparently, this used to be installed by the runfile and it must be now installed explicitly form the repo.
* Disable Wayland so that GDM always start Xorg. This used to be necessary only in Ubuntu, but with the new installation method it became necessayr on RHEL as well.

### Tests
SUCCESS build+kitchen on all OSS, barring build/kitchen failure on RHEL9 which is broken by an unrelated issue.

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/resources/dcv/partial/_dcv_common.rb
+++ b/cookbooks/aws-parallelcluster-platform/resources/dcv/partial/_dcv_common.rb
@@ -75,6 +75,9 @@ action_class do
     # DO NOT install dcv-gl on non-GPU instances, or will run into a black screen issue
     install_dcv_gl
 
+    # Ensure GDM starts an Xorg server (not Wayland) before launching X
+    disable_wayland
+
     # Configure the X server to start automatically when the Linux server boots and start the X server in background
     bash 'Launch X' do
       user 'root'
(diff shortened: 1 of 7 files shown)


cookbooks/aws-parallelcluster-platform/resources/dcv/partial/_dcv_common.rb
```diff
@@ -94,6 +97,33 @@ def allow_gpu_acceleration
...
+            sed -i '/\\[daemon\\]/a WaylandEnable=false' #{conf}
+          fi
+        fi
+      DISABLEWAYLAND
+    end
+
+    # Restart GDM so it re-reads the setting and comes up on Xorg.
+    service 'gdm' do
+      action :restart
+    end
```
hanwen-cluster: Do we need to restart gdm? Asking because we didn't restart gdm in the old ubuntu code

gmarciani: It is not strictly required, but a safe guard that I think makes sense. We are not in control of the time when gdm reads the updated config. Restarting it ensures that it reloads the updated config.

hanwen-cluster: NTH: we should move the disablement of Wayland in build-image phase instead of configuration phase

hanwen-cluster (approved):

cookbooks/aws-parallelcluster-platform/resources/dcv/partial/_dcv_common.rb
gmarciani: I agree, we can do as NTH for the release in a follow up PR

PR #3261, opened 2026-08-13
[BuildImage] Fix nvidia installation gate + epel and custom node package on proxied envs

gmarciani:
### Description of changes
1. Fix NVIDIA installation gate: now the entire nvidia installation recipe is skipped if nvidia is disabled or the nvidia driver is already installed. Without this change, nvidia-imex installation is triggered even if the parent image has nvidia disabled. As a result, for exmaple on DLAMI, the installation will attempt to install the latest nvidia-imex from the remote repository, so wrong version from wrong source. Does not need a changelog entry because the fix became mandatory only after changing the nvidia installation, which is something we introduced in 3.16.0.
2. Presign custom node package URL against the regional S3 endpoint. Without this change, the custom aws-parallelcluster-node package download fail with 403 when the artifacts bucket lives outside us-east-1. No need for changelog entry because this is a fix to a DevSettings. We need this fix because otherwise we cannot validate the build-image in isolated networks + in the past our users needed to adopt this devsettings to mitigate issues, so it is important to fix it.

### FAQ
1. **if the problem was only the gate on nvidia-imex, why adding the general gate to the entire nvidia recipe?**
Because this is what we should have done since the beginning: a simple gate for the entire nvidia software stack. Instead we had multiple and sometimes divergent gates that may create issues, like in this case.
1. **If the general gate is the solution, then why not removing the specific gates?**
We will do it post-release as a cleanup. For now our priority is to merge the minimal change that fixes the problem. 

### Tests
SUCCESS

```
test-suites:
  createami:
    test_createami.py::test_build_image_no_internet:
      dimensions:
        # RHEL family (rhel8, rhel9) -> Red Hat Enterprise Linux reservation platform
        - regions: [{{ g4dn_2xlarge_CAPACITY_RESERVATION_2_INSTANCES_3_HOURS_NOPG_rhel9 }}]
          instances: ["g4dn.2xlarge"]
          oss: ["rhel8", "rhel9"]
        # Non-RHEL family (alinux2023, ubuntu2204, ubuntu2404, rocky8, rocky9) -> Linux/UNIX reservation platform
        - regions: [{{ g4dn_2xlarge_CAPACITY_RESERVATION_5_INSTANCES_3_HOURS_NOPG_ubuntu2404 }}]
          instances: ["g4dn.2xlarge"]
          oss: ["alinux2023", "ubuntu2204", "ubuntu2404", "rocky8", "rocky9"]
```

### DIFF
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/install/custom_parallelcluster_node.rb
@@ -59,7 +59,7 @@ bash "install aws-parallelcluster-node from #{node['cluster']['custom_node_packa
     source #{node_virtualenv_path}/bin/activate
     pip uninstall --yes aws-parallelcluster-node
     if [[ "#{node['cluster']['custom_node_package']}" =~ ^s3:// ]]; then
-      custom_package_url=$(#{cookbook_virtualenv_path}/bin/aws s3 presign #{node['cluster']['custom_node_package']} --region #{node['cluster']['region']})
+      custom_package_url=$(#{cookbook_virtualenv_path}/bin/aws s3 presign #{node['cluster']['custom_node_package']} --region #{aws_region} --endpoint-url https://s3.#{aws_region}.#{aws_domain})
     else
       custom_package_url=#{node['cluster']['custom_node_package']}
     fi
--- a/cookbooks/aws-parallelcluster-computefleet/spec/unit/recipes/custom_parallelcluster_node_spec.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/spec/unit/recipes/custom_parallelcluster_node_spec.rb
@@ -33,7 +33,7 @@ describe 'aws-parallelcluster-computefleet::custom_parallelcluster_node' do
   source #{virtualenv_path}/bin/activate
   pip uninstall --yes aws-parallelcluster-node
   if [[ "#{custom_node_s3_url}" =~ ^s3:// ]]; then
-    custom_package_url=$(#{cookbook_virtualenv_path}/bin/aws s3 presign #{custom_node_s3_url} --region #{region})
+    custom_package_url=$(#{cookbook_virtualenv_path}/bin/aws s3 presign #{custom_node_s3_url} --region test_region --endpoint-url https://s3.test_region.test_aws_domain)
   else
     custom_package_url=#{custom_node_s3_url}
   fi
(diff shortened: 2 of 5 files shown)


cookbooks/aws-parallelcluster-shared/resources/package_repos/partial/_package_repos_rpm.rb
```diff
@@ -42,3 +42,17 @@
...
+#
+# The yum-epel default mirrorlist (mirrors.fedoraproject.org) returns a set of arbitrary
+# third-party mirror hostnames. In isolated / proxy-filtered networks (e.g. `pcluster
+# build-image` in a private subnet with a domain allowlist) those mirror hosts are not
+# reachable, so `yum makecache epel` fails with "All mirrors were tried". dl.fedoraproject.org
+# serves the repository metadata directly (no redirect to random mirrors) and can be reached
+# wherever *.fedoraproject.org is allowlisted, so it works both online and in isolated networks.
+def configure_epel_repo
+  node.default['yum']['epel']['baseurl'] =
+    "https://dl.fedoraproject.org/pub/epel/#{node['platform_version'].to_i}/Everything/#{node['kernel']['machine']}/"
```
hanwen-cluster: This link might be slow in China

gmarciani: Discussed offline, descoped from the last minute fix. 
We will handle this the way a user would do, which is to manually retrieve all the mirrors and allowlist them in the proxy.
Done in https://github.com/aws/aws-parallelcluster/pull/7561

cookbooks/aws-parallelcluster-platform/resources/nvidia_imex/partial/_nvidia_imex_common.rb
```diff
@@ -91,6 +92,10 @@ def nvidia_imex_service
   'nvidia-imex'
 end
 
+def nvidia_imex_full_version
+  "#{node['cluster']['nvidia']['driver_version']}-1"
```
hanwen-cluster: Are you sure imex version always ends with `-1`?

gmarciani: yes, but not future proof, so I removed.
Also, figutred out we do not need to pin, it would be a redundant safew guard so no need to add this complexity anyways

hanwen-cluster (approved):

PR #3267, opened 2026-08-19
[Pcluster-diag] Modify FSx checks to remove unwanted checks

himani2411:
### Description of changes
**EFA data-path ping**
  - Ping all discovered `@efa` peers, not just the first, so one bad NID no longer fails a healthy fabric.
  - Only ping peers LNet actually discovered (`@tcp` primary with an `@efa` rail). Leftover entries from a failed manual ping, and the node's own nids, are skipped — a self-ping was previously enough to make a fully broken path look healthy.
  - New W4 `SOME_EFA_PEERS_UNREACHABLE` when only some peers fail; E11 now means *all* failed, i.e. local breakage (usually the missing self-referencing SG rule).
  - Messages no longer claim traffic "falls back to TCP" — a failed ping can't tell us that.

  **EFA device binding**
  - Instance types not in the FSx doc table now expect all devices bound, so a partial bind reports E17 instead of passing silently.
  - I5 no longer claims the count matched an expectation that was never computed: new W3 `PARTIAL_BIND_UNVERIFIABLE` when it's undecidable, I10 when every device is bound.

  **Removed checks that fired on healthy filesystems or added no coverage**
  - Import-state (E4), failover-pin (I1), TCP-fallback (W3), and I8 binding reference, plus the now-dead `lctl ...import` parser.

  **Parsing fix**
  - `lfs check: error: ...` was read as a target named "lfs", producing a phantom E3. Now a row must name an MDT/OST.

  Note: W3 is reused mid-series (`TCP_FALLBACK` is deleted, then the number is reassigned to `PARTIAL_BIND_UNVERIFIABLE`), so reviewing commit-by-commit is clearer than the squashed diff.

### Tests
* Unit tests
* Integ test as part of https://github.com/aws/aws-parallelcluster/pull/7572
* Added a unknown peer ` lnetctl peer add --prim_nid 192.168.5.161@tcp --nid 5.162.0.0@efa` (so polluted a peer table)
```
 {
      "check_id": "LustreFilesystem",
      "check_description": "Verify that FsxLustre filesystems are installed, mounted, reachable, and riding EFA.",
      "status": "WARNING",
      "warnings": [
        {
          "code": "W4",
          "message": "EFA ping from 66.248.0.248@efa failed for 1 of the 2 discovered @efa peers (5.162.0.0@efa), while the others answered -- the local EFA path works, so whatever is wrong is specific to those peers. Confirm in `lnetctl peer show` that those nids belong to peers this node should be reaching: an @efa nid left behind by an earlier manual ping, or a rail of a peer that is no longer part of the filesystem, fails exactly the same way."
        }
      ],
      "infos": [
        {
          "code": "I1",
          "message": "Lustre client version: 2.15.6."
        },
        {
          "code": "I2",
          "message": "Active LNet transports: tcp, efa."
        },
        {
          "code": "I6",
          "message": "EFA driver version: 3.1.0g."
        },
        {
          "code": "I7",
          "message": "kefalnd (EFA LND) version: 1.2.2."
        },
        {
          "code": "I4",
          "message": "The EFA-Lustre configuration service configure-efa-fsx-lustre-client.service is installed and not failed."
        },
        {
          "code": "I5",
          "message": "1 of 1 EFA devices are bound to LNet, matching the expected count for this instance type."
        }
      ]
    },
    {
      "check_id": "FsxTargetsAreReachable",
      "check_description": "Deep-probe each FsxLustre OST/MDT for reachability (lfs check servers).",
      "status": "PASSED"
    }
  ]
}
```
* Passing checks 
 ```
 {
      "check_id": "LustreFilesystem",
      "check_description": "Verify that FsxLustre filesystems are installed, mounted, reachable, and riding EFA.",
      "status": "PASSED",
      "infos": [
        {
          "code": "I1",
          "message": "Lustre client version: 2.15.6."
        },
        {
          "code": "I2",
          "message": "Active LNet transports: tcp, efa."
        },
        {
          "code": "I6",
          "message": "EFA driver version: 3.1.0g."
        },
        {
          "code": "I7",
          "message": "kefalnd (EFA LND) version: 1.2.2."
        },
        {
          "code": "I4",
          "message": "The EFA-Lustre configuration service configure-efa-fsx-lustre-client.service is installed and not failed."
        },
        {
          "code": "I5",
          "message": "1 of 1 EFA devices are bound to LNet, matching the expected count for this instance type."
        }
      ]
    },
    {
      "check_id": "FsxTargetsAreReachable",
      "check_description": "Deep-probe each FsxLustre OST/MDT for reachability (lfs check servers).",
      "status": "PASSED"
    }
  ]
}
 ```

### References

### Checklist

### DIFF
--- a/cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/fsx_connectivity.py
+++ b/cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/fsx_connectivity.py
@@ -40,9 +40,9 @@ executes each probe in isolation and aggregates their findings. The probes are:
   that (re)configures LNet on every boot, then detects common root causes: no EFA device bound to LNet at
   all (so Lustre falls back to TCP), fewer devices bound than the instance type is expected to bind (the
   expected count comes from a per-instance-family table, NOT the raw device count -- several families bind
-  only a subset by design, and instance types with no known/static expectation are not flagged beyond the
-  "none bound" case), and a non-working EFA data path (typically a missing self-referencing security-group
-  rule).
+  only a subset by design; an instance type absent from that table is expected to bind all present devices,
+  and only a genuinely unknown instance type is left unflagged beyond the "none bound" case), and a
+  non-working EFA data path (typically a missing self-referencing security-group rule).
   See https://docs.aws.amazon.com/fsx/latest/LustreGuide/configure-efa-clients.html
 
 :class:`FsxTargetsAreReachable` is kept separate because it is a heavier, opt-in
@@ -180,13 +180,15 @@ class LustreFilesystem(Check):
     )
     NO_DEVICES_BOUND = CheckError(
         10,
-        "{} EFA devices are exposed but none are bound to LNet -- Lustre will fall back to TCP. "
-        "Re-run the EFA-Lustre client configuration.",
+        "{} EFA devices are exposed but none are bound to LNet -- Lustre will fall back to TCP. Check the "
+        "EFA-Lustre client configuration output (`journalctl -u {}`).",
     )
     UNDERBOUND_DEVICES = CheckError(
         17,
         "Only {} of {} expected EFA devices are bound to LNet on {} -- Lustre will not use the full EFA "
-        "fabric. Re-run the EFA-Lustre client configuration.",
+        "fabric. Check the EFA-Lustre client configuration output (`journalctl -u {}`). If this instance "
+        "type binds fewer devices by design, confirm the expected count against official FSX doc "
+        "https://docs.aws.amazon.com/fsx/latest/LustreGuide/configure-efa-clients.html",
     )
     EFA_PING_FAILED = CheckError(
         11,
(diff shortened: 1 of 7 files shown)


hanwen-cluster:
General comment: We should avoid changing meaning of existing error/warning/info code, and use new codes

himani2411:
> General comment: We should avoid changing meaning of existing error/warning/info code, and use new codes

Agreed to not re-shuffling the codes, but do not agree on changing the message of any code. However the message should not convey a different or opposite meaning of what was conveyed before. So a message change where we convey a different remediation or ask to check a different log is acceptable.


Also some of the codes were not being invoked ever because the checks were never being invoked. So I am changing their error codes. For example TCP_FALLBACK was never invoked because the 2 checks using that code was never called https://github.com/aws/aws-parallelcluster-cookbook/pull/3262/changes

cookbooks/aws-parallelcluster-platform/files/pcluster-diag/pcluster_diag/checks/fsx_connectivity.py
```diff
@@ -222,9 +224,26 @@ class LustreFilesystem(Check):
         "idle or freshly-booted node; if the node has been driving filesystem I/O it may indicate a silent "
         "TCP fallback.",
     )
-    TCP_FALLBACK = CheckWarning(3, "target {} is connected over @tcp despite EFA being configured (TCP fallback).")
```
himani2411, line 225: NOTE TO REVIEWER: TCP_FALLBACK was never invoked because the 2 checks using that code was never called https://github.com/aws/aws-parallelcluster-cookbook/pull/3262/changes. So I am changing the meaning of this code.

hanwen-cluster (approved):

PR #3271, opened 2026-08-27
[UpdateWorkflow] Remove the fuser pre-check before shared storage unmount

hehe7318:
### Description of changes
`fuser -mv <mount point>` ran before every EBS, EFS and FSx unmount and only logged its output, but it blocks in an uninterruptible kernel wait when the storage server is unreachable, so the Chef run never reaches the unmount. `timeout 10` cannot help: Chef enforces it while waiting on the child, and such a process cannot be reaped.

The `umount -fl`(force+lazy) needs no knowledge of who holds the mount.

### Tests
* All the spec test passed

### Checklist


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This file is used to list changes made in each version of the AWS ParallelCluste
 3.17.0
 ------
 
+**BUG FIXES**
+- Fix cluster update hanging and eventually failing when a node has an unresponsive shared storage mount by no longer running `fuser` on the mount point before unmounting it.
```
gmarciani: This changelog entry seems overstating. 
I agree with the removal of checking active processes before unmount. However, the fix is not enough to prevent the hanging. If the storage is unresponsive, we know that the cluster update may hang on the gates.

hehe7318: Refined to:
Fix cluster update hanging on the `fuser` probe that ran before every shared storage unmount when the storage was unresponsive.

Does it look good to you?

gmarciani (approved):

PR #3278, opened 2026-09-09
Ship only AWS CLI v2 in ParallelCluster AMIs

hehe7318:
### Description of changes
ParallelCluster AMIs shipped two AWS CLIs: v2 at `/usr/local/bin/aws`, installed system-wide by the `awscli` recipe, and v1 inside the cookbook virtualenv. The v1 copy came from the `awscli` entry in `cookbook_virtualenv/requirements.txt`: AWS CLI v2 is not published on PyPI, so `pip install awscli` always resolves to v1.

This PR makes the cookbook use only the system-wide AWS CLI v2:

* Add a `cluster/aws_cli_bin` attribute (`/usr/local/bin/aws`) and repoint every `#{cookbook_virtualenv_path}/bin/aws` call to it.
* Remove `awscli` from `cookbook_virtualenv/requirements.txt`.
* Install the cookbook virtualenv dependencies from S3 with `pip install -r cookbook-requirements.txt -f ./ --no-index` instead of `pip install *`. The pre-built bundle is used as an offline package index, so `requirements.txt` is the single source of truth for both install paths and the virtualenv stops picking up `awscli` (and its exclusive dependencies `colorama`, `docutils`, `rsa`, `pyasn1`) without rebuilding the S3 bundle.
* Invert the `awscli` InSpec control: `/usr/local/bin/aws` must be v2 and the cookbook virtualenv must not contain `awscli`.

No S3 artifact needs to be uploaded or removed. With this change the extra wheel is simply not installed.

Side effect: in air-gapped environments, the node package presign no longer needs the extra SigV4 configuration that AWS CLI v1 required.

### Tests
* ChefSpec for `aws-parallelcluster-platform`, `aws-parallelcluster-computefleet` and `aws-parallelcluster-slurm` (0 failures); cookstyle clean.
* pip dry-run of `requirements.txt` against the live `pypi-cookbook-dependencies-3.14-x86_64.tgz` bundle resolves 22 packages with the same versions as today, `awscli` excluded.
* `pcluster build-image` (alinux2023, x86_64) with this cookbook through the production install path, i.e. the S3 bundle that still contains `awscli-1.44.18`: build succeeded, the cookbook virtualenv contains the 22 packages from `requirements.txt` and no `awscli`/`bin/aws`, `/usr/local/bin/aws` is `aws-cli/2.36.40`, and the node package presign through CLI v2 worked.
* Cluster created from the resulting AMI. chef-client.log shows every call going through `/usr/local/bin/aws`.

### DIFF
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
@@ -53,7 +53,7 @@ when 'HeadNode'
   unless on_docker? || kitchen_test? && !node['interact_with_ddb']
     execute 'initialize compute fleet status in DynamoDB' do
       # Initialize the status of the compute fleet in the DynamoDB table. Set it to RUNNING.
-      command "#{node['cluster']['aws_cli_bin']} dynamodb put-item --table-name #{node['cluster']['ddb_table']}"\
+      command "/usr/local/bin/aws dynamodb put-item --table-name #{node['cluster']['ddb_table']}"\
               " --item '{\"Id\": {\"S\": \"COMPUTE_FLEET\"}, \"Data\": {\"M\": {\"status\": {\"S\": \"RUNNING\"}, \"lastStatusUpdatedTime\": {\"S\": \"#{Time.now.utc}\"}}}}'" \
               " --region #{node['cluster']['region']}"
       retries 3
--- a/cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
+++ b/cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
@@ -53,7 +53,7 @@ when 'HeadNode'
   unless on_docker? || kitchen_test? && !node['interact_with_ddb']
     execute 'initialize compute fleet status in DynamoDB' do
       # Initialize the status of the compute fleet in the DynamoDB table. Set it to RUNNING.
-      command "#{cookbook_virtualenv_path}/bin/aws dynamodb put-item --table-name #{node['cluster']['ddb_table']}"\
+      command "#{node['cluster']['aws_cli_bin']} dynamodb put-item --table-name #{node['cluster']['ddb_table']}"\
               " --item '{\"Id\": {\"S\": \"COMPUTE_FLEET\"}, \"Data\": {\"M\": {\"status\": {\"S\": \"RUNNING\"}, \"lastStatusUpdatedTime\": {\"S\": \"#{Time.now.utc}\"}}}}'" \
               " --region #{node['cluster']['region']}"
       retries 3
--- a/cookbooks/aws-parallelcluster-platform/files/cookbook_virtualenv/requirements.txt
+++ b/cookbooks/aws-parallelcluster-platform/files/cookbook_virtualenv/requirements.txt
@@ -3,7 +3,6 @@ supervisor
 requests
 jsonschema
 jinja2
-requests
 PyYAML
 retrying
 click
--- a/cookbooks/aws-parallelcluster-platform/files/cookbook_virtualenv/requirements.txt
+++ b/cookbooks/aws-parallelcluster-platform/files/cookbook_virtualenv/requirements.txt
@@ -1,4 +1,3 @@
-awscli
 boto3
 supervisor
 requests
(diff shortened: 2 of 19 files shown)


cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
```diff
@@ -53,7 +53,7 @@
   unless on_docker? || kitchen_test? && !node['interact_with_ddb']
     execute 'initialize compute fleet status in DynamoDB' do
       # Initialize the status of the compute fleet in the DynamoDB table. Set it to RUNNING.
-      command "#{cookbook_virtualenv_path}/bin/aws dynamodb put-item --table-name #{node['cluster']['ddb_table']}"\
+      command "#{node['cluster']['aws_cli_bin']} dynamodb put-item --table-name #{node['cluster']['ddb_table']}"\
```
gmarciani: why do we need an attribute to control the installation path of aws cli?

If there is a valid reason for that path to be an attributed, then we must use the same attribute when we install the awscli at `install awscli`.

[minor] To be consistent with other attributes related to dependencies, this attribute should be rather be named `node['cluster']['aws_cli']['install_dir']`, similarly to `node['cluster']['slurm']['install_dir']`

cookbooks/aws-parallelcluster-platform/files/cookbook_virtualenv/requirements.txt
gmarciani: Since we are touching this file, can we remove the duplicate `requests` in a dedicated commit

cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
hehe7318: No real reason: the awscli recipe installs the CLI at a fixed location, so the attribute only half-abstracted the path. Removed it and hardcoded `/usr/local/bin/aws`.

cookbooks/aws-parallelcluster-platform/files/cookbook_virtualenv/requirements.txt
hehe7318: Great catch! Done!

cookbooks/aws-parallelcluster-platform/test/controls/awscli_spec.rb
```diff
@@ -10,23 +10,32 @@
...
+  # AWS CLI v2 is not published on PyPI, so any awscli package in the virtualenv would be v1.
   describe bash("#{node['cluster']['cookbook_virtualenv_path']}/bin/pip list") do
     its('exit_status') { should eq(0) }
-    its('stdout')      { should match('awscli') }
+    its('stdout')      { should_not match('awscli') }
   end
 
-  describe file('/usr/local/bin/aws') do
-    it { should exist }
+  describe file("#{node['cluster']['cookbook_virtualenv_path']}/bin/aws") do
```
gmarciani, line 32: [minor] This test used to verify the existance of a path. now it is overkilling to test the non existance. It is enough to verify that pip list does not contain aws-cli

cookbooks/aws-parallelcluster-computefleet/recipes/config/fleet_status.rb
gmarciani: I agree with not exposing a chef attribute to customize the installation of awscli (overkilling and expands the customization surface exposed to users which can cause noise and ops load).

However, hardwiring the same path multiple times is not the best practice here. we shoudl avoid redefining paths, and instead have a constant.

hehe7318: Done!

cookbooks/aws-parallelcluster-platform/test/controls/awscli_spec.rb
hehe7318: Done!

cookbooks/aws-parallelcluster-platform/recipes/install/awscli.rb
gmarciani: You must inject the new aws cli path constant here

hehe7318: Thanks! Great catch, fixed.

gmarciani (approved):

